PR #19, opened 2016-06-18
verify direction of sort api work 

smarks:
add test data that will allow comparison. 
of course the main part is yet done.


lwhite1:
Looks mainly good. The diff on Table is fucked, tho. Makes it look like you rewrote the whole file. if you just changed the Table.create() method, we're good

src/test/java/com/github/lwhite1/tablesaw/SortTest.java
```diff
@@ -1,18 +1,71 @@
...
+    /**
+     * Make sure each row in each table match
+     *
+     * @param sortedTable the table that was sorted with tablesaw
+     * @param compareWith the table that was sorted using some external means e.g. excel. i.e known good data
+     */
+    private void compareTables(Table sortedTable, Table compareWith) {
+        assertEquals("both tables have the same number of rows", sortedTable.rowCount(), compareWith.rowCount());
+        int maxRows = sortedTable.rowCount();
+        int numberOfColumns = sortedTable.columnCount();
```
smarks: This seems somehow wrong to me. 
Should there be a comparison method on table? 
Should there be a getRow method?
What is the purpose of the table.row(int i) method? (that confused me) 
Anyway some discussion points here I think.

lwhite1: >> Should there be a comparison method on table? 
um. to compare as in your new compareTables() method? or for use in sorting() as in a Comparable?

>> Should there be a getRow method?
well, you get a cell with get(int rowNum, int colNumber);
There is no get(recordNumber) and there isn't one in SQL, but it's possible to add one. Generally, you would want to select a row using a query, because the table is not required to have a unique id per row (again like sql). 

When you use a query, the result is a bitmap that works like a compressed array of ints. These are the numbers of the rows you get back. It could be made user friendly by also supporting a regular int[].

If you tell me what you want, i can probably be more helpful

>>What is the purpose of the table.row(int i) method? (that confused me) 
That's a good one. Should open a bug.

src/test/java/com/github/lwhite1/tablesaw/TestData.java
```diff
@@ -18,12 +15,19 @@
  */
 public enum TestData {
 
-    BUSH_APPROVAL(new ColumnType[]{LOCAL_DATE, INTEGER, CATEGORY}, "data/BushApproval.csv"),
```
lwhite1: no more with the bushies?  or Tornados?

src/test/java/com/github/lwhite1/tablesaw/TableTest.java
```diff
@@ -14,35 +22,61 @@
...
+    @Test
+    public void testGetRow() throws Exception {
+
+        Table table = TestData.SIMPLE_UNSORTED_DATA.getTable();
+        Path source = TestData.SIMPLE_UNSORTED_DATA.getSource();
+        List<String[]> rawData;
+        try (CSVReader csvReader = new CSVReader(Files.newBufferedReader(source, Charset.defaultCharset()))) {
+            rawData = csvReader.readAll();
+        }
+        // remove the first row from the raw data since tables don't keep column header row as part of their data
```
smarks: It looks like a table intentionally discards the first row data. I'd like to understand this better.

src/main/java/com/github/lwhite1/tablesaw/api/Table.java
```diff
@@ -35,651 +38,679 @@
...
+     * @throws IndexOutOfBoundsException if the index is invalid
+     */
+    public String getRow(int index) {
+        StringBuilder stringBuilder = new StringBuilder();
+
+        for (Column column : columnList) {
+            stringBuilder.append(column.getString(index));
+            stringBuilder.append(",");
+        }
+
```
smarks: this is sucky (but it works). I am working on a fancy streams/ java 8 version now.

src/test/java/com/github/lwhite1/tablesaw/TableTest.java
lwhite1: The file reading API assumes the csv file has a one row header, unless you use one of the variants that allow you to manage that. could that be the problem?

src/main/java/com/github/lwhite1/tablesaw/api/Table.java
lwhite1: you really can't build a csv String without a CSV library (too many issues with escaping quotes, etc.). There is code that does this in the CsvWriter class.  
IMO, this is a utility that should be outside the table class. Maybe in a test utilities class?

src/test/java/com/github/lwhite1/tablesaw/TableTest.java
```diff
@@ -14,35 +22,61 @@
...
+        rawData.remove(0);
+        int maxRows = table.rowCount();
+        for (int rowIndex = 0; rowIndex < maxRows; rowIndex++) {
+            String row = table.getRow(rowIndex);
+            String expectedRow = getCSVfromListOfStrings(rawData.get(rowIndex));
+            assertEquals("Row " + rowIndex + " matches", expectedRow,row);
+        }
```
smarks: so there is a bug in this data. TableSaw converts date format 

org.junit.ComparisonFailure: Row 0 matches 
Expected :Spencer,52,Harvard,9/4/63
Actual   :Spencer,52,Harvard,2063-09-04

which is interesting.

lwhite1: you can't expect it to give you back the date as a String in a specific
format unless you somehow specify the format. Tablesaw uses ISO format to
output Strings from dates by default, but if you need a specific format you
can get a Java 8 localDate from the column and use standard date formatting.

to get a java 8 LocalDate object, then string format that any way you
want.  But please see my comment on why you don't want to write this method
without using OpenCSV (and don't really want it in the Table class.)

lwhite1: sorry. that should have read:

you can't expect it to give you back the date as a String in a specific
format unless you somehow specify the format. Tablesaw uses ISO format to
output Strings from dates by default, but if you need a specific format you
can get a Java 8 localDate from the column and use standard date formatting.

But please see my comment on why you don't want to write this method
without using OpenCSV (and don't really want it in the Table class.)

smarks: I added better test data

smarks: In the test I had to make sure the getRow method worked. So I had to use the direct OpenCVS reader to get the "known" good data. (I could have gotten that data any number of ways, including just hard coding it, but this way seemed OK.)

Then I compare it to the expected data which TableSaw provides. 

Does this make sense?

src/test/java/com/github/lwhite1/tablesaw/TestData.java
smarks: That delete is just the diffing tool being imperfect. 
Or do you mean I am not using this test data yet? I am sure it will be useful, but I wanted to start with something really simple since I needed to create known good results form the sort. The bushes and Tornados are still available as TestData though.

src/test/java/com/github/lwhite1/tablesaw/TableTest.java
lwhite1: I don't think you want a getRow() method in the Table class at all. It
should be a test utility method. It should use OpenCSV to construct "csv"
version of the table row, or it will fail on funky data.

src/test/java/com/github/lwhite1/tablesaw/TestData.java
lwhite1: ok.

src/test/java/com/github/lwhite1/tablesaw/TableTest.java
smarks: yes, that is what bit me.The default is strip header.  From this note, there is probably another constructor that keeps the header. That works. I just need to read the documentation better.

src/test/java/com/github/lwhite1/tablesaw/TestData.java
lwhite1: please call me if you're not getting what I'm saying about Table#getRow().
thanks.

src/test/java/com/github/lwhite1/tablesaw/SortTest.java
smarks: maybe I did the wrong thing by adding the a getRow method? I think we need to chat about the above paragraph since I am not following most of it, but I hope I am not wasting your time by adding the getRow method...

src/test/java/com/github/lwhite1/tablesaw/TestData.java
smarks: ok

src/main/java/com/github/lwhite1/tablesaw/api/Table.java
```diff
@@ -35,651 +38,677 @@
...
+     * <p>
+     * if column name starts with - then sort that column descending otherwise sort ascending
+     */
+    public Table sortOn(String... columnNames) {
+        Sort key = null;
+        Order order = null;
+        for (String columnName : columnNames) {
+              order = getOrder(columnName);
+            if (key == null) { // first time through
+                order = getOrder(columnName);
```
lwhite1: order = getOrder in the conditional block is redundant. value is set just before the if.  missed this earlier

PR #22, opened 2016-06-19
Added constructors in Table that wrap the other CsvReader variations,…

lwhite1:
… so users can specify whether the file has a header, or a different delimiter. Also renamed the existing method to create()


lwhite1:
Going to just merge this because I screwed git and want to get back to a clean state.

PR #48, opened 2016-07-01
make column names case sensitive

awclives:
This bit me on my first usage of tablesaw.

I think we also should enforce uniqueness of column names in addColumn.  I think it should throw an exception if the name already exists.  Uniqueness is assumed and required in many places so I think it should be enforced.

### DIFF
--- a/src/main/java/com/github/lwhite1/tablesaw/table/Relation.java
+++ b/src/main/java/com/github/lwhite1/tablesaw/table/Relation.java
@@ -58,7 +58,7 @@ public interface Relation {
   default int columnIndex(String columnName) {
     int columnIndex = -1;
     for (int i = 0; i < columnCount(); i++) {
-      if (columnNames().get(i).equalsIgnoreCase(columnName)) {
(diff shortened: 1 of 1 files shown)


lwhite1:
I agree that enforcing column name uniqueness is important. I will add an issue to track it until it's fixed.
But I'm not sure I agree about making the names case-sensitive. How did this bite you?

My feeling is that the user will be often working with files created by others, with arbitrary column name capitalization and that it will impose an avoidable cognitive task to remember the capitalization of each name. 

How would you feel about just ensuring case-insensitive uniqueness?

awclives:
case-insensitive uniqueness sounds good to me.

that would have defanged the issue that bit me.  i added "Title" and "title" columns just as a little test of the library, and then figured out that the resulting CSV was screwed up due to the non-unique columns.  if it had thrown an exception when I added the second one, as you are proposing, I would have been happy.

lwhite1:
I will make it so :)

lwhite1:
closing after making changes to enforce column name uniqueness

PR #63, opened 2016-07-08
15 change the sort api to support '+' and '-' in column names

smarks:
Tests are not exhaustive in that they don't test every data type, but they int and dates appear to sort correctly


lwhite1:
lgtm. Want to merge?

smarks:
Yes I am ok with merging if you are.

PR #69, opened 2016-07-24
Fix failing test in LocalDateColumnTest

Wecks:
Fixed failing testAddCell test from LocalDateColumnTest.


lwhite1:
Thank you for the PR. I found and fixed the same issue a few hours ago, by adding another date rather than reducing the count:

  @Test
  public void testAddCell() throws Exception {
    column1.addCell("2013-10-23");
    column1.addCell("12/23/1924");
    column1.addCell("12-May-2015");
    column1.addCell("12-Jan-2015");
    assertEquals(4, column1.size());
    LocalDate date = LocalDate.now();
    column1.add(date);
    assertEquals(5, column1.size());
  }

The fix is in master now.  Sorry for the inconvenience

PR #73, opened 2016-07-26
AnyOf.apply doesn't work

richiethom:
The implementation of `apply()` wasn't updating the instance of `Selection`, as the `toBitmap()` method returns a clone rather than a reference to the internal implementation - the OR therefore wasn't working. By using the same approach as `AllOf.apply()`, OR functionality works as expected.


lwhite1:
Thanks for the fix. Greatly appreciated!. Again.

PR #74, opened 2016-08-01
add LongIndex

jonsondag:
Adding LongIndex; it appears this would be part of https://github.com/lwhite1/tablesaw/issues/64


lwhite1:
Thank you very much for this contribution. Looks great.

PR #76, opened 2016-08-05
Throw an exception when an Exception occurs while adding a cell

richiethom:
I have modified `CsvReader` a little so that if an `Exception` is thrown while adding a cell to a column, it is wrapped in the new `AddCellToColumnException`, which allows the following information to be retrieved about the location of the `Exception`:

- cause (eg `NumberFormatException`)
- line number
- column name
- the contents of the row

A useful error message is produced too.

`AddCellToColumnException` is a `RuntimeException` - no need to catch it everywhere.


src/main/java/com/github/lwhite1/tablesaw/io/csv/AddCellToColumnException.java
```diff
@@ -0,0 +1,53 @@
+package com.github.lwhite1.tablesaw.io.csv;
+
+import java.io.PrintStream;
+
+/**
+ * Created by Richard on 26/07/2016.
```
lwhite1: Please remove auto-generated code from comments and replace with a brief description of the class

src/main/java/com/github/lwhite1/tablesaw/io/csv/AddCellToColumnException.java
```diff
@@ -0,0 +1,53 @@
+package com.github.lwhite1.tablesaw.io.csv;
```
lwhite1: Tablesaw uses Google-style two character indentation rather than four characters.

lwhite1:
Thanks for the pull-request.

This looks good aside from two very minor issues I commented on in-line. If you prefer, I can make these changes for you.

richiethom:
I've fixed the formatting and improved the comments. Let me know if there's anything else to do.

lwhite1:
Thanks very much for your help!

PR #77, opened 2016-08-05
NullPointerException in PackedLocalDateTime

richiethom:
The `DateTimeColumn.add(LocalDateTime)` method causes a `NullPointerException` in `PackedLocalDateTime` if the `LocalDateTime` passed as a parameter is `null`.

I have modified `DateTimeColumn` to add a null check - if the value is `null`, then `Long.MIN_VALUE` is added instead, in the same way as for the method `addCell(String)`.


lwhite1:
Good catch. Thank you. 
I implemented it slightly differently and committed the fix. This is what I did: There is a constant in DateTimeColumn called MISSING_VALUE that is defined as Long.MIN_VALUE. I used that in lieu of using Long.MIN_VALUE directly. I also extended that change to the other places in DateTimeColumn where I had been using Long.MIN_VALUE directly so they are all using the MISSING_VALUE constant.

That fix is now in 0.7.2.1, which I pushed to maven central.  Thanks again.

PR #91, opened 2016-09-27
Fix: column type inference when values are empty; added: Parser config builder, "wide" csv indexer

kogupta:
Fix: 
- csv column type inference: during column type inference, values are sampled from column; if all sampled values are empty/null - inferred type should be `CATEGORY`; it was inferred as `LOCAL_DATE_TIME`.
- added a dateTime format

Added: 
- Wide csv reader: read a wide csv file in "batches" - `batch size` and other properties defined using `com.github.lwhite1.tablesaw.store.CSVParserConfig`

Misc code cleanup: 
- converting `indexed for-loop` to jdk5+ style `foreach` loop
- anonymous class to closure/lambda
- public mutable field to public immutable field
- Separated `StorageManager` into separate reader [`SawReader`] and writer [`SawWriter`].


benmccann:
@kogupta wow, this is a huge amount of effort! Thank you so much!

Apologies for the delay in responding. The contributions are very much appreciated. I would like to suggest that the best way to move forward with this would be to split it up into multiple PRs with smaller changes. Unfortunately, I think this change is just much too big to review all in one go, which is why it's sat here. I just joined the project, so sorry you didn't get some response on this sooner

I'm going to close this PR for now because it would need to be rebased, which I think would cause you a lot of difficulty. Feel free to reopen. I'm guessing it will be easier for you to create new PRs rather than try to use this one though

Again, this is an awesome amount of work and contribution from you, so I really want to say thank you and sorry it's sat here. I hope that we can help you move forward with it in a number of smaller changes

kogupta:
Ben,
Not really sure if this will reach you.

One query: are you going to be the principal maintainer of the project? Are
we going to see regular "effort" being put into the project?

My personal opinion: this is too good a project to be an abandonware; the
column store alone is worth the effort.

benmccann:
I will help Larry maintain the project. Also, deeplearning4j has a fork and we are talking to them about unforking and joining us in helping with this project

lwhite1:
@kogupta,

Sorry I wasn't able to keep up with the maintenance. I think having Ben on
board will make a huge difference. Also, thank you for the kind words about
the project.

best,

Larry

kogupta:
Sounds great guys, hope to contribute back.

PR #109, opened 2017-05-12
add one pattern for creating table from resultset which gets timestamp column in mysql

vangogh-ken:
Add one pattern for creating table from resultset which gets timestamp column in mysql. Fixed error while parse "2015-08-22 11:52:56.0"  for datetime column


benmccann:
@vangogh-ken sorry for the delay in reviewing. Would you be able to rebase so that I can merge?

Also, it might be better to put it as number 2 for organization and it looks like we could delete 3 or 4 while we're at it as they appear to be duplicates.

benmccann:
I've made this change in https://github.com/jtablesaw/tablesaw/commit/9cba5da143afffb44f8b5802cd08c4170e6d279a

PR #170, opened 2017-09-02
Infra/coverage

mario-s:
This change set will add:

- link to Apache license as image in the readme
- code coverage plugin

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -204,6 +204,25 @@
           </configuration>
         </plugin>
         <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-javadoc-plugin</artifactId>
+          <version>2.10.4</version>
+          <inherited>false</inherited>
+          <configuration>
+            <reportOutputDirectory>docs</reportOutputDirectory>
+            <destDir>apidocs</destDir>
+          </configuration>
+          <executions>
+            <execution>
+              <id>aggregate</id>
+              <phase>site</phase>
+              <goals>
+                <goal>aggregate</goal>
+              </goals>
+            </execution>
+          </executions>
+        </plugin>
+        <plugin>
             <groupId>org.codehaus.mojo</groupId>
             <artifactId>cobertura-maven-plugin</artifactId>
             <version>2.7</version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -203,28 +203,51 @@
...
+          <configuration>
+            <reportOutputDirectory>docs</reportOutputDirectory>
+            <destDir>apidocs</destDir>
+          </configuration>
+          <executions>
+            <execution>
+              <id>aggregate</id>
+              <phase>site</phase>
+              <goals>
+                <goal>aggregate</goal>
```
benmccann: This was setup for just the parent project so that running javadoc:aggregate would generate combined docs for all projects. I don't think moving it to pluginManagement is the right thing to do

pom.xml
```diff
@@ -203,28 +203,51 @@
...
+              </goals>
+            </execution>
+          </executions>
+        </plugin>
+        <plugin>
+            <groupId>org.codehaus.mojo</groupId>
+            <artifactId>cobertura-maven-plugin</artifactId>
```
benmccann: I believe JaCoCo is generally preferred to Cobertura now days

pom.xml
mario-s: Personally I like the configuration of a plugin do be done in the plugin management part and only include the group and artifact id in the build section. It looks neat to me.

pom.xml
mario-s: No problem with JaCoCo...will change it.
https://confluence.atlassian.com/clover/comparison-of-code-coverage-tools-681706101.html

pom.xml
```diff
@@ -203,6 +203,42 @@
...
+            </execution>
+            <execution>
+              <id>report</id>
+              <phase>prepare-package</phase>
+              <goals>
+                <goal>report</goal>
+              </goals>
+            </execution>
+            <execution>
+                <id>post-unit-test</id>
```
benmccann: The spacing is a bit off here

pom.xml
benmccann: Thanks!!

pom.xml
benmccann: It gets applied to the children projects then though, right? And we want the parent and children to have different configuration here. I'm not much of a Maven expert...

lwhite1:
Mario,  Thanks very much for your contribution. Really appreciated.

mario-s:
I like the idea behind this project, so I'm also willing to contribute more Good work guys!

benmccann:
That would be great! We'd love to have more help!

And yes, thanks again for setting up code coverage!

Issue #44, opened 2016-06-30
Test table#append() thoroughly


lwhite1:

lwhite1:
closed. Thanks @ustits!

PR #173, opened 2017-09-07
#44 tests for Table#append

ustits:
Added tests for cases when appending table:
* with one column
* with multiple columns
* with different shape
* null


core/src/test/java/tech/tablesaw/TableTest.java
```diff
@@ -8,16 +8,23 @@
...
  * Tests for Table
  */
 public class TableTest {
 
+    private static final int ROWS_BOUNDARY = 1000;
+
     private Table table;
     private FloatColumn floatColumn = new FloatColumn("f1");
     private DoubleColumn doubleColumn = new DoubleColumn("d1");
+    private Random random = new Random();
```
benmccann: it seems like this should be `static final`

core/src/test/java/tech/tablesaw/TableTest.java
```diff
@@ -8,16 +8,23 @@
...
+import static org.junit.Assume.assumeTrue;
 
 /**
  * Tests for Table
  */
 public class TableTest {
 
+    private static final int ROWS_BOUNDARY = 1000;
+
     private Table table;
```
benmccann: I find it makes the code quite difficult to follow having this be a class member vs having each test create its own instance of a table. All the methods have side effects on this table and it would be much clearer they acted upon the table if they took it as a parameter

core/src/test/java/tech/tablesaw/TableTest.java
```diff
@@ -64,4 +71,99 @@ public void testRowCount() throws Exception {
...
+        Table tableToAppend = Table.create("wrong", column);
+        table.append(tableToAppend);
+    }
+
+    @Test(expected = IllegalStateException.class)
+    public void testAppendTableWithDifferentShape() throws Exception {
+        FloatColumn column = new FloatColumn("e1");
+        table.addColumn(column);
+        Table tableToAppend = Table.create("different", column);
+        assumeTrue(table.columns().size() == 2);
```
benmccann: It seems like all the places where assume is used we should use assert instead. I would expect the test to fail if this is not true rather than be ignored

core/src/test/java/tech/tablesaw/TableTest.java
```diff
@@ -64,4 +71,99 @@ public void testRowCount() throws Exception {
...
+        assumeTrue(tableToAppend.columns().size() == 1);
+        table.append(tableToAppend);
+    }
+
+    private void appendTableWithSingleColumn(final int rowsToAppend) {
```
benmccann: You'd normally say "append to" rather than "append with". E.g. appendSingleColumnToTable. Though I think a better naming convention might be something like appendRandomlyGeneratedColumn

core/src/test/java/tech/tablesaw/TableTest.java
```diff
@@ -64,4 +71,99 @@ public void testRowCount() throws Exception {
...
+        table.append(tableToAppend);
+        assertColumnSize(floatColumn, initialSize + rowsToAppend);
+    }
+
+    private void assertColumnSize(Column column, int expected) {
+        int actual = table.column(column.name()).size();
+        assertEquals(expected, actual);
+    }
+
+    private int populateColumn(final FloatColumn floatColumn) {
```
benmccann: I would remove `final` everywhere for consistency. We use it only for class members or if the variable is going to be accessed from within a closure

ustits:
Got it, I'll fix

lwhite1:
@ustits Thanks very much for working on this. Greatly appreciated.

core/src/test/java/tech/tablesaw/TableTest.java
ustits: Methods like `appendColumn` now accept table as a parameter.

ustits:
Noticed, that there is also one test in `TableTest` in `api` package. Maybe they must be merged?

lwhite1:
merging the append tests into TableTest is a good idea i think.

ustits:
Only append tests? What about `testColumn`, `testFullCopy` etc?

benmccann:
Those should be moved as well.

PR #176, opened 2017-09-08
locale for date parsing

mario-s:
All test for the DateColumn failed on a machine which has not an English Locale. So a string with "12-May-2015" was not successfully parsed on a system with a German locale. 
With this PR it is possible to pass a locale to the DateColumn. The test run with English.


.travis.yml
```diff
@@ -9,6 +9,6 @@ jdk:
 before_install:
   - sudo apt-get install jq
   - wget -O ~/codacy-coverage-reporter-assembly-latest.jar $(curl https://api.github.com/repos/codacy/codacy-coverage-reporter/releases/latest | jq -r .assets[0].browser_download_url)
-install: mvn install -DskipTests -Dgpg.skip
```
benmccann: this should not be removed. the `install` just compiles the library. then the `script` section below runs the tests

see https://docs.travis-ci.com/user/languages/java/#Projects-Using-Maven

core/src/main/java/tech/tablesaw/api/DateColumn.java
```diff
@@ -97,6 +101,10 @@ public DateColumn(String name, IntArrayList data) {
         super(name);
         this.data = data;
     }
+    
+    public void setLocale(Locale locale) {
+        this.locale = locale;
```
benmccann: I think it would be better to pass this in via the constructor rather than a setter

mario-s: Let me add this to the constructor as well.

benmccann: You can create two constructors with it being optional in one of them. Immutability is almost always better. We should avoid having any setters

benmccann:
looks good to me. I'll let @lwhite1 give a final review

@lwhite1 if this looks good to you, you can press the "Squash and merge" button and it will merge as a single commit

lwhite1:
LGTM. Thanks much!

PR #180, opened 2017-10-14
javadoc

mario-s:
I added some return and param tags for the javadoc,. 
Some fields are final, as well as a overrideable method, which was called in the constructor.

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/CategoryColumn.java
+++ b/core/src/main/java/tech/tablesaw/api/CategoryColumn.java
@@ -25,6 +25,7 @@ import tech.tablesaw.util.Selection;
 
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


core/src/main/java/tech/tablesaw/api/CategoryColumn.java
```diff
@@ -291,7 +294,7 @@ public int countUnique() {
         return bottom;
     }
 
-    public void add(String stringValue) {
+    public final void add(String stringValue) {
```
benmccann: I'm not sure we need to make this `final`

mario-s: An option would be to move the content of the public method to another with private access modifier. But calling a overideable method from the constructor is considered to be a bad practice. https://stackoverflow.com/questions/3404301/whats-wrong-with-overridable-method-calls-in-constructors

benmccann: I'd probably prefer your second option so that users can still override this method if they have a need to

Issue #185, opened 2017-10-23
Add method to filter table by row index


YuanZhencai:

lwhite1:
There should be the equivalent of a slice operation, but you have to use iteration for now. 

for (int i = 50; i < 100; i++) {
  // copy the results from your table, column by column
}

benmccann:
Implemented here: https://github.com/jtablesaw/tablesaw/commit/cc72afd6a19265194ac7d7a2cd48b8ddc0a070f5

PR #192, opened 2017-11-11
simple implementation of selectRows method

artichaud1:
Based on logic in dropRows method.
Issue #185.
(Haven't ran any tests as I'm still learning how to deal with java projects...but the code is straightforward).

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/Table.java
+++ b/core/src/main/java/tech/tablesaw/api/Table.java
@@ -961,7 +961,6 @@ public class Table extends Relation implements IntIterable {
       allRows.removeAll(rows2);
       Rows.copyRowsToTable(allRows, this, newTable);
       return newTable;
-
     }
 
     /**
@@ -983,7 +982,6 @@ public class Table extends Relation implements IntIterable {
       Table newTable = emptyCopy();
       Rows.copyRowsToTable(rows, this, newTable);
       return newTable;
-
     }
     
     /**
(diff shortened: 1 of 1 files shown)


benmccann:
@artichaud1 thanks for this! I'm curious what your use case is if you don't mind sharing how you intend to use this new method

@lwhite1 thoughts?  I'm not sure it makes sense to expose `IntArrayList` to users as `dropRows` is currently doing. It seems like it should be an internal implementation detail that we're using `fastutil` and it would be better to expose an interface taking `Collection<Integer>` instead

artichaud1:
A few examples that come to mind:
- easily split a table into chunks for a home-made and quick map/reduce type of job.
- for time series data, I might want to train on top part of the table and test on bottom part.
- I might want for some odd reason to see what the table looks like in the middle instead of the head of the tail. Or maybe I just know the exact indices of the rows I'm interested in.

lwhite1:
I think collection<Integer> is better for an API method, although it sounds
like most of your use cases would be better served by specifying a range or
RangeSet

benmccann:
Thanks for reviewing @lwhite1. I had the same thought that a range would be better as well.

Thanks for suggesting this improvement @artichaud1. I changed the existing drop methods to take a range and Collection instead. I added the corresponding selectRow methods as well

artichaud1:
Very cool !
Btw: I probably did it wrong, but it wasn't my suggestion, I took the easiest issue I could find.
Question: if I want to test the new methods in beaker, what would be the simplest way ? (I assume they are not already reflected in mvn jars ?)

benmccann:
The approach you took made a lot of sense given the existing code.

If you run `mvn install` then if will install to your local Maven repository and you should be able to then reference it from BeakerX or another project. It can be pretty tricky to get it to work though. It can often take me awhile to figure out how to get it to work and I know Java pretty well. You can give it a shot, but it might just be easier to wait awhile for the next release

PR #197, opened 2017-11-14
Remove Kotlin files

benmccann:
@lwhite1 if this is okay with you, I'd like to remove the Kotlin files for now as you suggested

In case others are curious why:
- Not finished yet and not as well-maintained
- Not documented
- Makes it hard to edit the code in Eclipse since Kotlin editing is really only supported by Intellij
- Removes Kotlin dependency from core Java library. If we want to add it back later we should consider doing so in a separate project
- Not clear that the current Kotlin code lets you do anything in Kotlin that the Java code doesn't
- Doesn't appear to be implemented in the most concise manner. I don't know Kotlin well, but it appears that using `Derived` keyword would cut out 90%+ of this code

We could possibly support Kotlin in the future, but ideally we should address the above points if/when we do so


lwhite1:
You're doing most of the work so I think it's appropriate that you set the
direction. Please make whatever changes you feel are best here and
elsewhere.

benmccann:
Thanks Larry

benmccann:
@jchildress @holgerbrandl @homasnield A big driver for me in considering whether to remove the Kotlin wrappers is that it doesn't look to me like the current wrappers actually make anything easier in Kotlin, Since I don't know Kotlin and you do, I'd like to get your thoughts on whether I'm correct about that

I'm also wondering if it's possible to move to a separate project as an alternative so that Java users don't need to pull in the Kotlin runtime

Finally, I'm also curious about what the API could be (not just what the state is today) and if there's any really compelling opportunity. Even if we remove it today because it's determined there's not currently value, it may be worth having on our roadmap as something to add back with an improved implementation in the future

lwhite1:
Yeah, you should really stop using eclipse ;)

Have at it.

PR #209, opened 2017-12-07
Fix bug in BooleanMapUtils

lina2002:


core/src/main/java/tech/tablesaw/api/BooleanColumn.java
```diff
@@ -513,4 +514,16 @@ public Boolean next() {
         }
     }
 
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        BooleanColumn that = (BooleanColumn) o;
+        return java.util.Objects.equals(data, that.data);
```
benmccann: we should import `java.util.Objects`

core/src/test/java/tech/tablesaw/mapping/BooleanMapUtilsTest.java
```diff
@@ -0,0 +1,38 @@
+package tech.tablesaw.mapping;
+
+import org.junit.Test;
+import tech.tablesaw.api.BooleanColumn;
+
+import static org.junit.Assert.assertEquals;
+
+public class BooleanMapUtilsTest {
+    @Test
+    public void testAdd() {
```
benmccann: Add -> And

core/src/test/java/tech/tablesaw/mapping/BooleanMapUtilsTest.java
```diff
@@ -0,0 +1,38 @@
...
+        BooleanColumn booleanColumn = new BooleanColumn("");
```
benmccann: Most columns have a constructor like:

    public DoubleColumn(String name, double[] arr) {

I see `BooleanColumn` does not. It'd probably be good for you to add one so that creating these columns would be simpler like:

    BooleanColumn booleanColumn = new BooleanColumn("col1", new boolean[]{ true });

PR #211, opened 2017-12-08
Fix bug in min of DateColumn

lina2002:
I realised that code I "refactored"  yesterday had a bug. Actually, I changed the logic so now it occurs more frequently. Because missing value for DateColumn is set to Integer.MIN_VALUE whenever there is such value in a column "min" method is returning it. Before it was doing so only when the first value was missing.

Now I've fixed it and added test for it.


benmccann:
Thanks for adding a test

Issue #133, opened 2017-08-19
Implement a Bubble Plot


lwhite1:
Using xChart create an interface that enables easy rendering of a bubble plot from a tablesaw table. See the implementation of scatter plots for a very similar example.  The task is basically to add a new chart type for BubblePlot, with a constructor that takes exactly three columns of numeric data (subclasses of NumericColumn)

https://github.com/jtablesaw/tablesaw/blob/master/plot/src/main/java/tech/tablesaw/api/plot/Scatter.java

mario-s:
I can offer a helping hand here.

lwhite1:
That would be great! Thanks.

mario-s:
@lwhite1 Could you assign this issue to me? Thanks.

lwhite1:
@mario-s I don't think I can 'assign' anyone unless they're a member of the project, but it's not necessary in any case. The contributors guide describes the usual approach: https://jtablesaw.github.io/tablesaw/contributing

mario-s:
One questions regarding GUI-Test: I see that the Test are written as examples with a main methods. Is this intend for interested users to get a feeling of how to use API?
I'm asking this because I would like to write the test with Jemmy.
http://wiki.netbeans.org/Jemmy

lwhite1:
Implemented by @mario-s. Closing

PR #212, opened 2017-12-16
#133: Implement a Bubble Plot

mario-s:
This pull request will add the feature for bubble plots. 
There is also a new dependency as well as a change in the CI script in order to run GUI tests.


lwhite1:
Merged. Thanks very much!

PR #232, opened 2018-01-23
Fix difference computation in LongColumn

### LINKED ISSUE #231: Implementation of difference in LongColumn always returns a single value column.
https://github.com/jtablesaw/tablesaw/blob/master/core/src/main/java/tech/tablesaw/api/LongColumn.java#L692

It seems the for loop condition is reversed.

```
        for (int current = 1; current > data.size(); current++) {
            returnValue.append(get(current) - get(current + 1));
        }
```

ashvina:
Fixes #231. Fixes the for loop and adds a test for the column

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/LongColumn.java
+++ b/core/src/main/java/tech/tablesaw/api/LongColumn.java
@@ -697,14 +697,7 @@ public class LongColumn extends AbstractColumn implements LongMapUtils, NumericC
 
         returnValue.append(MISSING_VALUE);
         for (int current = 1; current < data.size(); current++) {
-            long currentValue = get(current);
-            long previousValue = get(current - 1);
-            // check for missing values
-            if (currentValue == MISSING_VALUE || previousValue == MISSING_VALUE) {
-                returnValue.append(MISSING_VALUE);
-            } else {
-                returnValue.append(currentValue - previousValue);
-            }
+            returnValue.append(subtract(get(current), get(current - 1)));
         }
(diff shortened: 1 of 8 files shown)


core/src/main/java/tech/tablesaw/api/LongColumn.java
```diff
@@ -692,8 +692,8 @@ public int byteSize() {
     public LongColumn difference() {
         LongColumn returnValue = new LongColumn(this.name(), data.size());
         returnValue.append(LongColumn.MISSING_VALUE);
-        for (int current = 1; current > data.size(); current++) {
-            returnValue.append(get(current) - get(current + 1));
+        for (int current = 1; current < data.size(); current++) {
```
benmccann: thanks for fixing. I just looked at `difference` from the other columns and noticed there's probably another bug here which is that this doesn't check for missing values. what do you think about fixing that too while you're at it?

the other inconsistency is that this method starts with `current = 1` and the others have an `if` check instead. this one is cleaner and it'd also be good to change the others to match the way it's done here for consistency if you're up for it

ashvina: Thanks @benmccann !
Sure, I will update this PR with the changes you suggested tomorrow.

ashvina:
The code quality issue reported by Codacy is incorrect. The tests are calling common validate method. Any way to avoid code duplication to fix this issue?

benmccann:
So much cleaner! Thank you!

Issue #144, opened 2017-08-22
Implement interface for hierarchical clustering using Smile


lwhite1:
Implement an interface to perform hierarchical cluster using Smile's implementation of the algorithm.  See the Non-hierarchical cluster interface code that has been implemented.

crew102:
Greetings, I've been working on this issue and wanted to see if it's OK if I introduce a dependency on smile-math in tablesaw? There are various implementations/measures of distance provided in the `smile.math.distance` package that would be helpful when creating the proximity matrix that is required for both hierarchical clustering and MDS (e.g., #149 ).

lwhite1:
As long as it's limited to the tablesaw-smile artifact that would be fine.
Thanks very much!

PR #233, opened 2018-01-24
Add draft version of hclust (re #144)

crew102:
This is a draft version of an interface to smile's hierarchical clustering API (issue #144). Two general notes:

* I added TODO placeholders for those javadoc tags which  I think need to be completed. I would like to wait to document things until the API is finalized (i.e., after incorporating your notes/feedback). Let me know if there are any items that I've flagged for documentation that don't actually have to be documented, or if I'm missing some items that you think need docs.

* I decided to not use smile's distance package, as only two of the classes inside it appeared to be relevant. Instead I used a method (`distance`) that I found in `smile.math.Math` that calculates Euclidean distance.

### DIFF
--- a/smile/src/main/java/tech/tablesaw/api/ml/clustering/Hclust.java
+++ b/smile/src/main/java/tech/tablesaw/api/ml/clustering/Hclust.java
@@ -1,0 +1,54 @@
+package tech.tablesaw.api.ml.clustering;
+
+import smile.clustering.HierarchicalClustering;
+import smile.clustering.linkage.Linkage;
+import tech.tablesaw.api.NumericColumn;
+
+/**
+ * TODO
+ */
+public class Hclust {
+	
+    private final HierarchicalClustering hierarchicalClustering;
+
+    /**
+     * Constructor.
+     * 
+     * @param linkageType TODO
+     * @param columns TODO
+     */
+    public Hclust(LinkageFactory.type linkageType, NumericColumn... columns) {
+        Proximity proximity = new Proximity(columns);
+        Linkage linkage = new LinkageFactory().createLinkage(proximity.getMatrix(), linkageType);
+        this.hierarchicalClustering = new HierarchicalClustering(linkage);
+    }
+    
+    public int[][] getTree() {
+        return hierarchicalClustering.getTree();
+    } 
+    public double[] getHeight() {
+        return hierarchicalClustering.getHeight();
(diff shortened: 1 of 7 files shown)


smile/src/main/java/tech/tablesaw/api/ml/clustering/LinkageFactory.java
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * TODO
+ */
+public class LinkageFactory {
+	
+	
+	/**
+	 * TODO
+	 */
+	public enum type {
```
crew102: Is there a better way to represent the different types of linkage methods than the one I have used here (i.e., placing a public enum inside a class whose job is to return classes, where the enum represents the different types of classes that can be created)?

smile/src/main/java/tech/tablesaw/api/ml/clustering/Hclust.java
```diff
@@ -0,0 +1,54 @@
+package tech.tablesaw.api.ml.clustering;
+
+import smile.clustering.HierarchicalClustering;
+import smile.clustering.linkage.Linkage;
+import tech.tablesaw.api.NumericColumn;
+
+/**
+ * TODO
+ */
+public class Hclust {
```
benmccann: `Hclust` isn't very clear. I would prefer `HierarchicalClustering`

smile/src/main/java/tech/tablesaw/api/ml/clustering/Proximity.java
```diff
@@ -0,0 +1,44 @@
...
+public class Proximity {
+	
+	private final double[][] proximityMatrix;
+	
+	/**
+	 * Constructor. 
+	 * 
+	 * @param columns TODO
+	 */
+	public Proximity(NumericColumn... columns) {
```
benmccann: Please use only spaces for consistency with the existing classes

crew102:
I'll do the docs sometime later this week or early next week

crew102:
Sorry, forgot to test locally last time and had some bad javadoc tags mixed into my original documentation commit. This should pass/be ready to go.

benmccann:
Looks fine to me

@lwhite1 you've done way more with the smile stuff than I have, so I'll leave this one to you

lwhite1:
Got it

PR #236, opened 2018-01-25
Generic subtract

ashvina:
Currently a `LongColumn` can be subtracted from a `LongColumn`. The current api does not allow subtracting a `IntColumn` or a `DoubleColumn`. This limitation can result in additional steps. For e.g. To compute difference between `Salary (long)` and `Median-Salary`, one might do this:  
```
NumericSummaryTable medians = table.median("Salary").by("ZipCode", "Sex") 
```
medians is a `DoubleColumn` of median, which cannot be directly subtracted from `LongColumn (Salary)`. One has to convert `Salary` to `Double` or `Median` to `Long`.

This PR makes `LongColumn.subtract` operation generic and compatible with other numeric types. It can operate on any `NumericColumn`. 

If this change seems meaningful, I can apply it to other `NumericColumns` too as part of this PR. And to other operation in future in different PRs.

Thanks !

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/LongColumn.java
+++ b/core/src/main/java/tech/tablesaw/api/LongColumn.java
@@ -244,10 +244,10 @@ public class LongColumn extends AbstractColumn implements LongMapUtils, NumericC
         return result;
     }
 
-    public LongColumn subtract(LongColumn column2) {
+    public LongColumn subtract(NumericColumn column2) {
         LongColumn result = new LongColumn(name() + " - " + column2.name(), size());
-        for (int r = 0; r < size(); r++) {
-            result.append(subtract(get(r), column2.get(r)));
+        for (int r = 0; r < size() && r < column2.size(); r++) {
+            result.append(subtract(get(r), column2.getLong(r)));
         }
         return result;
     }
--- a/core/src/main/java/tech/tablesaw/api/LongColumn.java
+++ b/core/src/main/java/tech/tablesaw/api/LongColumn.java
@@ -412,13 +412,13 @@ public class LongColumn extends AbstractColumn implements LongMapUtils, NumericC
     @Override
     public int getInt(int index) {
         double value = data.getLong(index);
-        return value == LongColumn.MISSING_VALUE ? IntColumn.MISSING_VALUE : (int) value;
+        return value == MISSING_VALUE ? IntColumn.MISSING_VALUE : (int) value;
     }
 
     @Override
     public float getFloat(int index) {
         double value = data.getLong(index);
-        return value == LongColumn.MISSING_VALUE ? FloatColumn.MISSING_VALUE : (float) value;
+        return value == MISSING_VALUE ? FloatColumn.MISSING_VALUE : (float) value;
     }
 
     @Override
@@ -429,7 +429,7 @@ public class LongColumn extends AbstractColumn implements LongMapUtils, NumericC
     @Override
     public double getDouble(int index) {
         double value = data.getLong(index);
-        return value == LongColumn.MISSING_VALUE ? DoubleColumn.MISSING_VALUE : (float) value;
+        return value == MISSING_VALUE ? DoubleColumn.MISSING_VALUE : (float) value;
     }
 
     @Override
--- a/core/src/main/java/tech/tablesaw/api/LongColumn.java
+++ b/core/src/main/java/tech/tablesaw/api/LongColumn.java
@@ -410,8 +410,26 @@ public class LongColumn extends AbstractColumn implements LongMapUtils, NumericC
     }
 
     @Override
+    public int getInt(int index) {
+        double value = data.getLong(index);
+        return value == LongColumn.MISSING_VALUE ? IntColumn.MISSING_VALUE : (int) value;
+    }
+
+    @Override
     public float getFloat(int index) {
-        return (float) data.getLong(index);
+        double value = data.getLong(index);
+        return value == LongColumn.MISSING_VALUE ? FloatColumn.MISSING_VALUE : (float) value;
+    }
+
+    @Override
+    public long getLong(int index) {
+        return data.getLong(index);
+    }
+
+    @Override
+    public double getDouble(int index) {
+        double value = data.getLong(index);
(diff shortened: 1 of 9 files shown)


core/src/main/java/tech/tablesaw/api/LongColumn.java
```diff
@@ -244,10 +244,10 @@ public LongColumn append(LongColumn column2) {
         return result;
     }
 
-    public LongColumn subtract(LongColumn column2) {
+    public LongColumn subtract(NumericColumn column2) {
```
lwhite1: Should a longColumn subtracting a double return a long column or a double column?

core/src/main/java/tech/tablesaw/api/NumericColumn.java
```diff
@@ -23,7 +23,21 @@
...
+        throw new UnsupportedOperationException("getFloat() method not supported for all data types");
+    }
+
+    default long getLong(int index) {
+        throw new UnsupportedOperationException("getLong() method not supported for all data types");
+    }
+
+    default double getDouble(int index) {
+        throw new UnsupportedOperationException("getDouble() method not supported for all data types");
+    }
```
lwhite1: getShort()?

core/src/main/java/tech/tablesaw/api/LongColumn.java
```diff
@@ -409,9 +409,27 @@ public long get(int index) {
         return data.getLong(index);
     }
 
+    @Override
+    public int getInt(int index) {
+        double value = data.getLong(index);
+        return value == LongColumn.MISSING_VALUE ? IntColumn.MISSING_VALUE : (int) value;
```
lwhite1: As noted by the build bot, it's not necessary to fully qualify MISSING_VALUE here. Similar issues elsewhere

core/src/main/java/tech/tablesaw/api/LongColumn.java
ashvina: Interesting question.
Looking at the standards, it should be double column (for e.g. `2 - 1.0 = 1.0`). However, It seems to me it will demand significant overloading. For e.g. `int - long` will be handled different from `int - double`. Or it might impact performance if additional type checks are required for every row. I am also concerned that changing the type might interfere with operator chaining.

As of now I am inclining towards returning the same type as that of the column on which the operation is invoked. What do you think?

ashvina:
The build failed outside the change introduced by this PR. Seems like there is a race condition caused in the original test.

```
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:86)
...
	at org.junit.Assert.assertFalse(Assert.java:74)
	at tech.tablesaw.api.TableTest.testAppendToNonEmptyTable(TableTest.java:113)
```

core/src/main/java/tech/tablesaw/api/NumericColumn.java
ashvina: Ahh.... missed this. Thanks. Will add this once the idea proposed here seems reasonable and aligned with rest of the code.

lwhite1:
I would prefer, I think, to get @benmccann to weigh in. I trust his design sense. 

I definitely get that it's more work, but not sure it's not worth it. I feel like it would be better to provide column conversion functions (double to long, for example) than to drop the fractional part without warning.

One compromise might be to provide two versions: One that returns Long (for long, ints, and shorts), and the other returning double for doubles and floats?

core/src/main/java/tech/tablesaw/api/LongColumn.java
benmccann: Or what about `2 - 0.5`? I'd expect it to return `1.5`

I'd expect that if we have have two columns being combined that we use the more precise type regardless of the order (i.e. `2.5 - 1` and `2 - .5` should both return `1.5`):
long, int -> long
long, short -> long
int, short -> int
double, anything -> double
float, anything-but-double -> float

ashvina: Sounds good to me. This approach prevents data loss. I will try and push an update. Thanks !

ashvina: Actually, `long` to `float` conversion could be concerning in some cases. For e.g. if a `LongColumn` contains `System.currentTimeMillis()`. A `float` conversion of this column can result in error of a few minutes. I am thinking of not supporting auto `long` to `float` conversion. 

```
float, anything-but-double-&-long -> float
```

ashvina:
Based on the discussion so far, this PR will make the following changes. 

Add to NumericColumn: `getShort(), getInt(), getLong(), getFloat() and getDouble()`

Changes to `ShortColumn`: 
  . implement `getShort(), getInt(), getLong(), getFloat() and getDouble()`
  . `aShortColumn.subtract(aLongColumn)` will delegate to LongColumn.subtract(aNumeric, bNumeric) and return a `LongColumn`
  . `aShortColumn.subtract(aFloatColumn)` will delegate to FloatColumn.subtract(aNumeric, bNumeric) and return a `FloatColumn`

Changes to `LongColumn`:
  . implement: `getLong(), getDouble()` and throw errors for others new methods.
  . `aLongColumn.subtract(aShortColumn)` will delegate to LongColumn.subtract(aNumeric, bNumeric) and return a `LongColumn`
  . add a new package-private `subtract(aNumeric, aNumeric)` method to service calls from lower range numerics
```
  LongColumn subtract(aNumeric, bNumeric) {
    loop {
       result = aNumeric.getLong() - bNumeric.getLong()
    }
  }
```

@benmccann @lwhite1 Please let me know if this looks ok.

core/src/main/java/tech/tablesaw/api/LongColumn.java
benmccann: yeah, maybe if you have a long and float you should convert them both to doubles?

benmccann:
This sounds like it could be starting to become a big change. How about if you implement the `getShort(), getInt(), getLong(), getFloat(), getDouble(), etc.` in one PR first? And then we can figure out the other changes in a separate PR to keep it more manageable

benmccann:
I merged the other PR, so you can rebase this one now or create a new one - whatever's easier

I'd recommend doing just a partial implementation first so we can see how it looks and agree on the approach before implementing fully

Issue #237, opened 2018-01-28
Allow subtraction between any two numeric columns types


ashvina:
Currently a `LongColumn` can be subtracted from a `LongColumn`. The api does not allow subtracting a `IntColumn` from a `LongColumn`. This limitation can result in additional data copies and steps. For e.g. To compute difference between `Salary (long)` and `Median-Salary`, one might do this:
```
NumericSummaryTable medians = table.median("Salary").by("ZipCode", "Gender")
```

Since medians is a `DoubleColumn`, it cannot be directly subtracted from the Salary column which is of type `LongColumn`. The user has to convert Salary to Double or Median to Long.

As discussed in #236, This issue can be solved in two phases
 1. Implement methods in `NumericColumns` to return `up-converted` values. For e.g. `ShortColumn` will implement `getInt()`, `getDouble` etc.
 1. Implement generic version of `subtract` method to return `up-converted` column types.

lwhite1:
closed. implemented

PR #239, opened 2018-01-30
Allow subtract operation between all numeric types

ashvina:
This PR addresses part-2 of #237: a generic subtract method implementation for `LongColumn` and `DoubleColumn`.

`aLongColumn.subtract(aDoubleColumn)` should result in a `DoubleColumn`. This could be achieved by delegating the operation to `DoubleColumn`. 

See discussion in #236 for more details.

**This PR is not merge-ready**. The intent of this PR is to demo high level code flow. If the approach is approved, changes will be required in other numeric types. Moreover the identification of a `delegate` could be abstracted out and be used in all numeric types.

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/DoubleColumn.java
+++ b/core/src/main/java/tech/tablesaw/api/DoubleColumn.java
@@ -570,10 +570,18 @@ public class DoubleColumn extends AbstractColumn implements DoubleIterable, Nume
         return result;
     }
 
-    public DoubleColumn subtract(DoubleColumn column2) {
-        DoubleColumn result = new DoubleColumn(name() + " - " + column2.name(), size());
-        for (int r = 0; r < size(); r++) {
-            result.append(subtract(get(r), column2.get(r)));
+    public NumericColumn subtract(NumericColumn column2) {
+        return DoubleColumn.subtractDouble(this, column2);
+    }
+
+    public static DoubleColumn subtractDouble(NumericColumn column1, NumericColumn column2) {
+        int col1Size = column1.size();
+        int col2Size = column2.size();
+        int resultSize = col1Size < col2Size ? col1Size : col2Size;
+
+        DoubleColumn result = new DoubleColumn(column1.name() + " - " + column2.name(), resultSize);
+        for (int r = 0; r < resultSize; r++) {
+            result.append(subtract(column1.getDouble(r), column2.getDouble(r)));
         }
         return result;
     }
@@ -835,7 +843,7 @@ public class DoubleColumn extends AbstractColumn implements DoubleIterable, Nume
         return val1 + val2;
     }
 
-    private double subtract(double val1, double val2) {
+    private static double subtract(double val1, double val2) {
         if (isMissing(val1) || isMissing(val2)) {
             return MISSING_VALUE;
         }
--- a/core/src/main/java/tech/tablesaw/api/LongColumn.java
+++ b/core/src/main/java/tech/tablesaw/api/LongColumn.java
@@ -244,10 +244,23 @@ public class LongColumn extends AbstractColumn implements LongMapUtils, NumericC
         return result;
     }
 
-    public LongColumn subtract(LongColumn column2) {
-        LongColumn result = new LongColumn(name() + " - " + column2.name(), size());
-        for (int r = 0; r < size(); r++) {
-            result.append(subtract(get(r), column2.get(r)));
+    public NumericColumn subtract(NumericColumn column2) {
+        if (column2 instanceof ShortColumn || column2 instanceof IntColumn || column2 instanceof LongColumn) {
+            return subtractLong(this, column2);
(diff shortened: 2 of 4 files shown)


ashvina:
Hi @benmccann. Humble reminder, please let me know if you have any comments. Thanks !

benmccann:
It looks like we may want to add support for `add`, `divide`, `multiply` in a similar manner. Are there other methods you think we might want to add beyond those down the road?

core/src/main/java/tech/tablesaw/api/DoubleColumn.java
```diff
@@ -570,10 +570,18 @@ public DoubleColumn subtract(double value) {
...
 
-    public DoubleColumn subtract(DoubleColumn column2) {
-        DoubleColumn result = new DoubleColumn(name() + " - " + column2.name(), size());
-        for (int r = 0; r < size(); r++) {
-            result.append(subtract(get(r), column2.get(r)));
+    public NumericColumn subtract(NumericColumn column2) {
+        return DoubleColumn.subtractDouble(this, column2);
+    }
+
+    public static DoubleColumn subtractDouble(NumericColumn column1, NumericColumn column2) {
```
benmccann: I don't think this needs to be `public`

core/src/main/java/tech/tablesaw/api/DoubleColumn.java
```diff
@@ -570,10 +570,18 @@ public DoubleColumn subtract(double value) {
         return result;
     }
 
-    public DoubleColumn subtract(DoubleColumn column2) {
-        DoubleColumn result = new DoubleColumn(name() + " - " + column2.name(), size());
-        for (int r = 0; r < size(); r++) {
-            result.append(subtract(get(r), column2.get(r)));
+    public NumericColumn subtract(NumericColumn column2) {
```
benmccann: This one actually can return `DoubleColumn`

core/src/main/java/tech/tablesaw/api/DoubleColumn.java
```diff
@@ -570,10 +570,18 @@ public DoubleColumn subtract(double value) {
...
+        return DoubleColumn.subtractDouble(this, column2);
+    }
+
+    public static DoubleColumn subtractDouble(NumericColumn column1, NumericColumn column2) {
+        int col1Size = column1.size();
+        int col2Size = column2.size();
+        int resultSize = col1Size < col2Size ? col1Size : col2Size;
```
benmccann: There's a lot of different ways I can imagine this working if the columns are not the same size. E.g. you could take the larger column and fill in missing values at the end. Is there any reason you chose this behavior? I'm wondering if we should just `throw new IllegalArgumentException("Columns have unequal size " + column1.size() + " and " + column2.size())`

core/src/main/java/tech/tablesaw/api/LongColumn.java
```diff
@@ -244,10 +244,23 @@ public LongColumn append(LongColumn column2) {
         return result;
     }
 
-    public LongColumn subtract(LongColumn column2) {
-        LongColumn result = new LongColumn(name() + " - " + column2.name(), size());
-        for (int r = 0; r < size(); r++) {
-            result.append(subtract(get(r), column2.get(r)));
+    public NumericColumn subtract(NumericColumn column2) {
```
benmccann: This could be a LongColumn

benmccann:
This PR will need to be rebased

benmccann:
Something interesting I just found from the Java spec:

5.6.2 Binary Numeric Promotion

When an operator applies binary numeric promotion to a pair of operands, each of which must denote a value of a numeric type, the following rules apply, in order, using widening conversion (§5.1.2) to convert operands as necessary:

If either operand is of type double, the other is converted to double.

Otherwise, if either operand is of type float, the other is converted to float.

Otherwise, if either operand is of type long, the other is converted to long.

Otherwise, both operands are converted to type int.

core/src/main/java/tech/tablesaw/api/DoubleColumn.java
ashvina: I don't have a strong reasoning for this choice. Some tools (for e.g. pandas) do retain the size of the larger column and fill missing values. The api such tools let the user provide a optional "fill" value. It is not easy to add optional parameters in java without making the api cumbersome. So I chose to save memory and ops by retaining the smaller column size. Instead of adding optional parameters I prefer throwing error as it forces the user to think (like Optionals). Let me know if you think throwing error matches rest of the implementation.

benmccann: I think throwing an error would be fine for now. I'd rather start out supporting less and then see where the need arises before introducing additional complexity

lwhite1: for the record, i made it throw an exception

PR #243, opened 2018-02-02
Add cumulative sum function to columns

numericOverflow:
Thanks for contributing.

## Description

- Added cumSum() and alias cSum() to Double/Float/Int/LongColumns to replicate cumulative sum functionality in pandas, etc. that returns a new column with cSum() applied to original column.

- Also had to add the isMissing() to IntColumn/LongColumn to keep coding similar across all type implementations.

## Testing

Did you add a unit test?

No, Sorry.   DoubleColumn version is what I added for my use-case, should work for other numeric versions too, but haven't tested that.  Compiles OK.

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/DoubleColumn.java
+++ b/core/src/main/java/tech/tablesaw/api/DoubleColumn.java
@@ -841,4 +841,27 @@ public class DoubleColumn extends AbstractColumn implements DoubleIterable, Nume
         }
         return val1 - val2;
     }
+
+    /**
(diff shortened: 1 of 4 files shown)


benmccann:
Please add this to all columns. E.g. `ShortColumn`

I prefer not to add a `cSum` method. Not sure we need a second method just to save two characters

Please add a unit test at least for the double version

numericOverflow:
Closing this pull request and I will create a new pull with new code including short and unit tests.

PR #245, opened 2018-02-04
Add cumulative sum function to columns - cumSum()

numericOverflow:
- Added cumSum() to Double/Float/Int/Long/ShortColumns to replicate cumulative sum functionality in pandas, etc. that returns a new column with cumSum() applied to original column.

- Also had to add the isMissing() to IntColumn/LongColumn/ShortColumn to keep coding similar across all type implementations.

Thanks for contributing.

## Description

New functions added as described above

## Testing

Yes, this new PR includes unit tests for all of the data types that compare function results to an expected result.

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/DoubleColumn.java
+++ b/core/src/main/java/tech/tablesaw/api/DoubleColumn.java
@@ -841,4 +841,19 @@ public class DoubleColumn extends AbstractColumn implements DoubleIterable, Nume
         }
         return val1 - val2;
     }
+
+    /**
+    * Returns a new column with a cumulative sum calculated
+    */
+    public DoubleColumn cumSum() {
+        double cSum = 0.0;
+        DoubleColumn newColumn = new DoubleColumn(name() + "[cumSum]", size());
+        for (double value : this) {
+            if (!isMissing(value)) {
+                cSum += value;
+            }
+            newColumn.append(cSum);
+        }
+        return newColumn;
+    }
 }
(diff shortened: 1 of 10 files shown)


core/src/main/java/tech/tablesaw/api/DoubleColumn.java
```diff
@@ -841,4 +841,19 @@ private double subtract(double val1, double val2) {
         }
         return val1 - val2;
     }
+
+    /**
+    * Returns a new column with a cumulative sum calculated
```
benmccann: Can you format the javadocs throughout this PR with an extra space on the lines after the first like:
```
/**
 *
 */
```

core/src/test/java/tech/tablesaw/api/DoubleColumnTest.java
```diff
@@ -465,4 +465,23 @@ private DoubleColumn createDoubleColumn(double[] originalValues) {
         }
         return initial;
     }
+
+    private boolean computeAndValidateCumSum(double[] originalValues, double[] expectedValues) {
```
benmccann: can you remove the `computeAndValidateCumSum` functions and just put the logic directly in `testCumSum`?  We only create a separate method in tests if it will be called by multiple tests. Otherwise it just adds a bit more code and makes the flow harder to follow

benmccann:
@numericOverflow I added a `cumProd` today. I updated the way that `cumSum` handles missing values to keep the two methods working the same way because I didn't think the behavior made sense for `cumProd`. The way I changed it is also how Pandas handles it so I thought it would make more sense to users from that respect as well. https://github.com/jtablesaw/tablesaw/commit/69cc150556f30211b62fd6acf11c4f7ea57193f0

PR #246, opened 2018-02-04
Fix temp paths so testing works on windows in StorageManagerTest.java

numericOverflow:
the "/tmp/*"  file paths won't work on windows, uses System.getProperty("java.io.tmpdir") to get local temp dir and updated paths to use that prefix.

Thanks for contributing.

## Description

maven tests fail on windows because the hardcoded temp test folder is invalid on Win machines:
**fails:**  /tmp/foo
**works:** System.getProperty("java.io.tmpdir") + "/foo"

## Testing

No unit test, but changes now compile and the new code passes the "mvn clean test" for me on Win7x64.  Would be good to have someone on linux run the tests as well (I don't have a linux box setup for java work)


benmccann:
Thanks for this!

PR #260, opened 2018-04-01
Version1

lwhite1:
Thanks for contributing.

## Description

Major revision to API and implementation.

- Changed Pom to build using Java 9
- Re-added support for building shaded jar (core-only) to facilitate use in java shell
- Removed Float, Int, Short, Long column types
- Removed TableGroup, SubTable, and NumericSummaryTable
- Removed Smile integration
- Removed experimental time interval support
- Renamed DoubleColumn to NumberColumn
- Renamed CategoryColumn to StringColumn
- Extended Join to handle doubles (after rounding to ints)
- Reduced the number of index types
- Removed Saw file persistence
- Removed all deprecated methods
- Removed numerous methods of unclear utility
- Removed redundant comparator implementations
- Reorganized filtering to better unify column and table filtering
- Added support for formatted printing of tables and columns (esp., number and time columns)
- Modified all columns to use static method create() rather than public constructors
- Added support for lag and lead methods on all column
- Removed Lombok
- Extended PackedLocalTime and PackedLocalDate to be essentially compatible with java's LocalDate and LocalTime
- Significantly extended test coverage (currently 57% on core,vs. 44% on prior commit).
- Standardize column create method support for lists and arrays

## Testing

Many tests added.


ms123s:
why did you remove smile?

lwhite1:
Removing Smile integration was a tough decision.

Tablesaw is a large library in need of a large amount of work. It needs to
be a great dataframe before it can be anything else, so that's where the
focus will be.

Smile is a huge library. Although I don't use it in my work, I wrote and
maintained nearly all the Smile integration code (one other person
submitted a wrapper for a single Smile model), but it still covers a
relatively small part of Smile. Keeping it working requires time I don't
have.

When I added it, I hoped that people would see that doing ML in Java was
easy using Smile together with Tablesaw, and that Tablesaw users, or the
(larger) Smile community, might help with the integration. That hasn't
happened.

I will put the code back on Github as a separate repo sometime today and
make sure it works with the latest version.  When Tablesaw is complete,
I'll consider going back to actively maintaining it. If you would like to
maintain it in the meantime, that would be great.

larry

benmccann:
Yeah, the problem is that we would have to create a wrapper for every Smile class. You couldn't do even basic things like cross-validation with the current integration. That means that trying to actually use Smile you were always switching between the wrapper classes and underlying implementation, which was kind of cumbersome. We thought it would be better to simply make it easy to get data out of the Table in a format that Smile could use and encourage folks to use Smile directly

PR #281, opened 2018-04-18
API for filling columns.

hallvard:
Thanks for contributing.

- [x ] Tick to sign-off your agreement to the [Developer Certificate of Origin (DCO) 1.1](https://developercertificate.org)

## Suggestion for and implementation of API for filling various column types with data.
Currently covered are columns for boolean, double, LocalDateTime, LocalDate and LocalTime.
All are based on the same API pattern, a filler interface that is implemented by the column, with fillWith methods for Iterators, Iterables and Suppliers. Default Iterables for ranges of each type (not boolean) is provided. The time-related columns also have static createWith methods for creating columns filled with a range, since these cannot be created with null as initial values.

## Testing

There are tests for each column type and range Iterable.

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/BooleanColumn.java
+++ b/core/src/main/java/tech/tablesaw/api/BooleanColumn.java
@@ -682,44 +682,44 @@ public class BooleanColumn extends AbstractColumn implements BooleanMapUtils, In
             return false;
         }
     }
-    
+
     // fillWith methods
-    
-	@Override
-	public BooleanColumn fillWith(BooleanIterator iterator) {
-		for (int r = 0; r < size(); r++) {
-			if (! iterator.hasNext()) {
-				break;
-			}
-			set(r, iterator.nextBoolean());
-		}
-		return this;
-	}
 
-	@Override
-	public BooleanColumn fillWith(BooleanIterable iterable) {
-		BooleanIterator iterator = null;
-		for (int r = 0; r < size(); r++) {
-			if (iterator == null || (! iterator.hasNext())) {
-				iterator = iterable.iterator();
-				if (! iterator.hasNext()) {
-					break;
-				}
-			}
-			set(r, iterator.nextBoolean());
-		}
-		return this;
-	}
+    @Override
+    public BooleanColumn fillWith(BooleanIterator iterator) {
+        for (int r = 0; r < size(); r++) {
+            if (!iterator.hasNext()) {
+                break;
+            }
+            set(r, iterator.nextBoolean());
+        }
+        return this;
+    }
 
-	@Override
-	public BooleanColumn fillWith(Supplier<Boolean> supplier) {
-		for (int r = 0; r < size(); r++) {
-			try {
-				set(r, supplier.get());
-			} catch (Exception e) {
-				break;
-			}
-		}
-		return this;
-	}
+    @Override
+    public BooleanColumn fillWith(BooleanIterable iterable) {
+        BooleanIterator iterator = null;
+        for (int r = 0; r < size(); r++) {
+            if (iterator == null || (!iterator.hasNext())) {
+                iterator = iterable.iterator();
+                if (!iterator.hasNext()) {
+                    break;
+                }
+            }
+            set(r, iterator.nextBoolean());
+        }
+        return this;
+    }
+
+    @Override
+    public BooleanColumn fillWith(Supplier<Boolean> supplier) {
+        for (int r = 0; r < size(); r++) {
+            try {
+                set(r, supplier.get());
+            } catch (Exception e) {
+                break;
+            }
+        }
+        return this;
+    }
 }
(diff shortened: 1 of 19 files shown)


core/src/main/java/tech/tablesaw/api/BooleanColumn.java
```diff
@@ -667,15 +671,55 @@ public boolean hasNext() {
...
-        }
-    }
+            return false;
+        }
+    }
+    
+    // fillWith methods
+    
+	@Override
+	public BooleanColumn fillWith(BooleanIterator iterator) {
```
benmccann: you have tabs here and spaces above. should be consistent with the rest of tablesaw

core/src/main/java/tech/tablesaw/api/BooleanColumn.java
```diff
@@ -667,15 +671,55 @@ public boolean hasNext() {
...
+		for (int r = 0; r < size(); r++) {
+			if (! iterator.hasNext()) {
```
benmccann: please remove space after each `!`

hallvard:
The Quality Review says there are tests without assert or fail. The methods in question all call a helper method that includes assert.
The Travis build failure seems unrelated to my code.

hallvard:
The Quality Review says there are tests without assert or fail. The methods in question all call a helper method that includes assert.
The Travis build failure seems unrelated to my code.

lwhite1:
The Travis build fails randomly, afaict.

core/src/main/java/tech/tablesaw/api/BooleanColumn.java
```diff
@@ -667,15 +671,55 @@ public boolean hasNext() {
...
-        public Boolean next() {
+        public boolean nextBoolean() {
             byte b = iterator.nextByte();
             if (b == (byte) 0) {
                 return false;
             }
             if (b == (byte) 1) {
                 return true;
             }
-            return null;
```
lwhite1: This breaks missing value handling for BooleanColumns, as a missing value becomes false when you iterate.  That is why the original iterator uses Boolean instead of boolean.  My concern here is that if the iterator were used to copy data between columns, for example, what is missing in the source would become false in the destination.

lwhite1:
Sorry to be weighing in belatedly; it's been a busy week. 

I don't understand this comment: "The time-related columns also have static createWith methods for creating columns filled with a range, since these cannot be created with null as initial values" Can you explain?

From a design standpoint, I like the thoroughness of the implementation, but wonder if perhaps more of the work could be outside of the Colums, in a utility class perhaps? If the main intent is to create data for testing or illustration, I'm on the fence about whether we should add three or more methods to each column type. They're already very large.

core/src/main/java/tech/tablesaw/api/BooleanColumn.java
hallvard: I missed the point of missing values, probably because I used the set method, which does not support setting a missing value (or a byte), whereas append does. Would it be right to add corresponding set methods, i.e. set(int, byte) and set(int, Boolean)?

hallvard:
@lwhite1 In the test for LocalDateColumn (I think) I tried to create a column initialised to an array with nulls (to be filled), and got an NPE (I think when it was converted to packed format or something). Then I thought it would make sense to avoid the empty array and fill upon creation, rather than afterwards. So you can either use create(...).fillWith(...) or just createWith(...).
I agree the API is a bit bloated, I just tried to adapt to the fluent style. We could just as well have static methods in a Fillers class (in a fillers package). I don't have a personal opinion, I did this work mainly to get familiar with Tablesaw (and to a good deed before using all the good work).

lwhite1:
@hallvard I really appreciate this work. Thank you very much. I will look at the issue you mention as it sounds like a bug. 

I think I can reduce the bloat a little by replacing the existing create(List) methods with your iterator version which is better. 

Unfortunately, work and life have me pretty busy for a few weeks, but I'll find the time to think about the design issues as soon as i can. I really want this stuff in Tablesaw.

core/src/main/java/tech/tablesaw/api/BooleanColumn.java
hallvard: If Column was parameterised with the element type, we could in general support append and set methods that convert null to the missing value. Se issue
 #285

lwhite1:
Is there a way to do this without leading to a lot of auto-boxing?  I think
the usual Column<Double> approach would be bad for performance and
scalability.

hallvard:
I think the best way of handling this is providing two methods, one with T and one with the native type, when this will matter, and make it easy to use the one you want, and/or provide methods that override the default ones, that can take this into account. E.g. if you have a **set(int, Double)** and **set(int, double)** you will usually end up using the best one, but for methods that return the element type, like get, you will need distinct method names.  min and max for DoubleColumn, where you have double get(i) and can compare the double values, will need to override the default one, which will be inherited and used by StringColumn and the time-related columns.

Issue #280, opened 2018-04-18
Smile integration? You want some? Or no?


lwhite1:
The current development branch removes Smile integration. As it is nearing completion, I'm considering adding it back, since it wouldn't be too much work (probably) to bring it in line with the changes not that they're settling down.

If you have an opinion on the matter, it would be a great time to make your voice heard.

benmccann:
Some of the stuff that lives in tablesaw-smile should probably be moved to smile directly. E.g. support for confusion matrix and the ability to name variables

thomashaselwanter:
I dont have a strong opinion on this, as I am just starting to use tablesaw and therefore have not used the smile integration in the past. smile does look pretty useful though and I'd defenitely tinker with it using tablesaw. So a weak **Yes** from my side.

mark27q1:
I have the sense I'd use this if it were available. I don't do a lot of ML work in Java, having so far worked mostly in python, Octave and C++ when I need serious scale... but then I only found Tablesaw a week or two back so that might change...

lwhite1:
i think there's a good chance I will add this back when I finish wrapping up this release, but no promises. :)

benmccann:
Smile can still be used with Tablesaw even without support built-in. I think the question is which is easier. I actually found it easier to use Smile directly for anything but toy examples. When I tried to use the Tablesaw wrapper I found lots of functionality I couldn't access, so I had to keep switching back and forth between the two which made it harder than just using Smile directly

benmccann:
I updated the BeakerX example to show how to use Smile directly with Tablesaw: https://github.com/twosigma/beakerx/pull/7669

There are a could changes we could make in Tablesaw to make it a bit smoother.

Converting to double arrays and matrices is super common. E.g.:
```
moneyball.numberColumn("RD").asDoubleArray()
moneyball.copy().retainColumns("W").asMatrix()
```

We might be able to make this API a bit easier since it's a frequent operation:
```
moneyball.asDoubleArray("RD")
moneyball.asMatrix("W")
```

The only thing that I felt like we lost was named variables. Smile does everything via indices which isn't as nice. [I filed a feature request for this with Smile a couple months ago](https://github.com/haifengl/smile/issues/297) and the maintainer seemed quite open to it. I think it's probably better that the feature be added there as that would be much cleaner than having to maintain wrappers of all the Smile classes just to have variable names

benmccann:
[...]

PR #302, opened 2018-07-10
Add convenience methods to convert to double arrays for easier usage with Smile

benmccann:

## Description

This makes using Tablesaw with Smile a bit smoother as described in https://github.com/jtablesaw/tablesaw/issues/280#issuecomment-403702698.

Converting to double arrays and matrices is super common. E.g.:
```
moneyball.numberColumn("RD").asDoubleArray()
moneyball.copy().retainColumns("W").asMatrix()
```

This PR provides a much simpler API for this common operation:
```
moneyball.asDoubleArray("RD")
moneyball.asMatrix("W")
```

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/Table.java
+++ b/core/src/main/java/tech/tablesaw/api/Table.java
@@ -260,17 +260,6 @@ public class Table extends Relation implements Iterable<Row> {
     /**
      * Returns only the columns whose names are given in the input array
      */
-    public List<Column> columns(String... columnNames) {
-        List<Column> columns = new ArrayList<>();
-        for (String columnName : columnNames) {
-            columns.add(column(columnName));
-        }
-        return columns;
-    }
-
-    /**
-     * Returns only the columns whose names are given in the input array
-     */
     public List<CategoricalColumn> categoricalColumns(String... columnNames) {
         List<CategoricalColumn> columns = new ArrayList<>();
         for (String columnName : columnNames) {
--- a/core/src/main/java/tech/tablesaw/columns/Column.java
+++ b/core/src/main/java/tech/tablesaw/columns/Column.java
@@ -195,9 +195,7 @@ public interface Column {
         return "Column: " + name() + '\n';
     }
 
-    default double[] asDoubleArray() {
-        throw new UnsupportedOperationException("Method asDoubleArray() is not supported on non-numeric columns");
-    }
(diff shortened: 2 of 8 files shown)


lwhite1:
I think asDoubleArray() belongs on the column since you're only converting the column. You're not doing anything with the table other than retrieving the column, which is a common operation any tablesaw user would know. I would rather not add it to the relation/table API. 

AsMatrix is definitely better in the sense that it is a multi-column operation, if not an actual table operation like sort. The benefit is that it's easier to find than the DoubleArrays class, but it's not necessarily _easy_ to find. There are over 150 methods on Relation/Table.  

You don't have to copy and retain, you can use any of these: 

- DoubleArrays.to2DArray(moneyball.columns());
- DoubleArrays.to2DArray(moneyball.columns("W", "L"));
- DoubleArrays.to2DArray(moneyball.numberColumns());
etc.

These are properly multi-column operations. 

I'm sure the uses for this functionality expands beyond the (vanishingly small intersection of Smile/Tablesaw users) so it's worth spending some time on, but I suspect there are Smile functions that take int[] or int[][], or other similar things. I've seen some visualization libraries (JavaFX for example), that like float[] instead of double[].  

All things considered, I'm not convinced that there isn't a better way to organize this functionality. Maybe by renaming the DoubleArrays class (and methods) and getting it out of utils (api or a new converter/conversions package might not be so bad). 

tech.tablesaw.api.TableConverter? 

What do you think?

lwhite1:
An example of an additional, possibly useful, conversion is to ND4J (N-Dimensional Arrays for Java) format. https://nd4j.org/. 

They're billing it as like numpy for java. It seems likely that future ML libraries will support it.

lwhite1:
> Today I do have to copy and retain, because the columns(String...) method doesn't exist. 

The PR moves it from table to relation, which is a clear improvement and should be done. 

> Regarding ND4J, you create a row/column vector from a double[] or a 2d NDArray from a double[][], so these would help there as well.

or an int[] or float[], but sure. 

> I have another idea. What if we make the API look like:
```
moneyball.as().matrix("W")

```
I like it. Should the matrix method be typed, rather than always producing double[][]?

benmccann:
Ok, I've updated the PR to have `as().floatMatrix()`, `as().doubleMatrix()`, `as().intMatrix()`

PR #306, opened 2018-07-13
Removed filtering based on QueryHelper & ColumnReferences.

lwhite1:
Thanks for contributing.

### Filters
There were two ways to query a table. The first is using Selections, and the second Filters.  In the past, we emphasized filters. Here's an example.

```Java
tornadoes.structure().where(stringColumn("Column Type").isEqualTo("NUMBER"));
```

In this example, structure() returns a table, which is filtered by the where clause. 

### Selections

Here's the same query using Selections

```Java
Table structure = tornadoes.structure();
structure.where(structure.stringColumn("Column Type").isEqualTo("NUMBER"));
```

There are two differences. First, the Filter approach is more concise. The stringColumn() method there is actually a static method invoked with a static import. There's no need to explicitly reference the table returned by structure, which makes the code a bit easier to read. 

The second difference is that the Filter approach lets you to chain table queries. The selection approach does not. That's because in the Filter approach the call to ```stringColumn()``` doesn't directly return a column, but a ColumnReference object. The ColumnReference is only associated with an actual table when the where() clause is executed(), and only then does it return a real column. 

The stringColumn() method for selection is defined on Table. For that to work that has to be something (a variable referring to a table) that it can be invoked on. 

So the filter approach enables more concise query code, plus the ability to chain table-returning queries together. With selections there's no chaining. To call a method on a table in a where clause, you need a variable to invoke it on: two statements instead of one. 

Other than these two differences, there is no functional difference.  

### The cost

Those  are the benefits of the filter approach. The costs are 

- much more code in Tablesaw: it more than doubled the amount of code required to implement filters on a column, 
- much more testing is required
- more Tablesaw 'inside knowledge' to use it effectively. 
- It doesn't play as well with autocomplete, since you have to separately import QueryHelper for that to work.  

## Testing

Did you add a unit test?
No. No code was added.


lwhite1:
@benmccann Would you mind taking a look?  This is a big change, and is not backward compatible. The reasoning is in the description.

benmccann:
Wow! Amazing cleanup!  I like that you found a way to remove 7,500+ lines of code and also make the API nicer at the same time. I like not having to import `QueryHelper`. I think we should go ahead and merge this

benmccann:
Although here's one question I have... What about performance? I'm not sure selections allow us to do query planning in the future in the way that a filter might

benmccann:
Although here's one question I have... What about performance? I'm not sure selections allow us to do query planning in the future in the way that a filter might

E.g. let's say we have `WHERE userid = 50001 AND username = 'new user'`. With a selection we've already selected the actual rows, so it really matters which clause comes first before the "AND" clause which really adds a lot of mental overhead and chances for a newcomer to make a mistake. But if we're able to pass `WHERE userid = 50001 AND username = 'new user'` to Tablesaw and then let Tablesaw detect whether either has an index or utilize pre-computed statistics about the number of distinct values in the column then it can optimize which half of the where clause gets executed first

lwhite1:
Those are good points, but i'm not _too_ concerned. We're pretty fast already because everything's in memory and the filters all stream primitives through the CPUs (I think. Probably the string filter's don't always, but they should). That said, we're definitely not as fast as we could be in many cases. 

Besides looking at the distribution of data and applying indexes, parallelizing the selections is another likely win. 

I would rather not have to keep filtering around in its current form, though, because it's such a burden.  Ideally, we could just re-add and(), or, and not() as static table methods, and put the optimization logic there. But we'd need a kind of 'deferred selection' class to make it work. I'll give it some thought. LMK if you have any ideas.

lwhite1:
Having slept on it, I think I will go ahead and make this change, for a few reasons. 
- As you pointed out, it gets rid of 7500 lines and something like 100 classes
- It's fast enough for most apps. 
- It would be a non-trivial amount of work to create a 'deferred selection' solution and then write an optimizer. (although I wrote a query optimizer once and it was kinda fun).
- Anyone who needs it to go faster is probably a pretty solid java hacker, because either they're dealing with a lot of data (100 million rows), they're using it in a production program, or they're doing something pretty hardcore like @mark27q1, with his 19 million searches over 4+ million rows. 
- Said java hacker can do a lot of optimization programmatically. It's easy to split a query into clauses, run the clauses in their own threads and combine the resulting selections. You can also use indexes and incorporate what you know about the data that way.

mark27q1:
19 **thousand** searches over 4 million rows mate, jeez :) You make me sound like some kind of "mad data scientist"... oh, wait... :)

lwhite1:
That's pretty much what I was thinking, and part of the effort i mention above. 

I think where this might have helped is in two ways. using 

``` 
    and(doSelectA, doSelectB);
```
instead of 

```
    doSelectA.and(doSelectB);
```
gives you a simple way to separate the clauses, and either re-arrange them or run them in separate threads. 

Basically, since selections implemented Filter, and filters converted to selections on evaluation, Filter's were a kind of deferred selection. We also need a real column to be passed into deferredSelection so there's something to evaluate it against. Eval() could do that work. 

I think besides making a DeferredSelection for each kind of eval statement (mostly this differentiates between predicate types), we would also need to convert all the Selection returning methods to return a DeferredSelection instead, and make the where() implementation (maybe) promote the deferredSelection to an actual selection and return its results. In the half-assed prototype i made, the DeferredSelection class would construct and wrap a regular selection on first invocation of any method requiring results and it forwarded any subsequent requests to the wrapped object.

So I think the bulk of the work is making all the methods like

```
    StringColumn::startsWith(String s)
```
return a deferredSelection. The work is pretty rote and wouldn't take more than a day probably but it is tedious.

If you have a need for an optimized version LMK. I'm happy to work on it when there's someone with a actual need.

lwhite1:
I guess I misunderstood. I thought a run contained 975 query executions an
you wanted to make 19,000 runs.  ~19 million.

mark27q1:
Actually... put that way, you are right. So I guess I _am_ that crazy... Fair enough, carry on, sorry! :)

PR #311, opened 2018-07-16
Provide converter to Smile's AttributeDataset

benmccann:
Smile defines an `AttributeDataset`, which is essentially the same thing as a Tablesaw `Table`

A lot of Smile algorithms take `AttributeDataset` as input, so this will make it easier still to interface with Smile. The dependency is `optional` which means `Smile` should not be added to your classpath, but if you're using `Smile` already you can simply call `table.as().smileDataset(...)`

It will also let us support named variables. The `Attribute`s are essentially column metadata such as the variable name. In the next release of Smile, `OLS` will have an `AttributeDataset` constructor which solves the variable name problem with that algorithm. Once a release is cut, you can simply do `new OLS(table.as().smileDataset(...))`. Let me know if you encounter other algorithms which you think should take an `AttributeDataset` and do not today

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -121,6 +121,12 @@
             <version>0.7.14</version>
         </dependency>
         <dependency>
+            <groupId>com.github.haifengl</groupId>
+            <artifactId>smile-data</artifactId>
+            <version>1.5.1 </version>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
             <groupId>com.google.code.findbugs</groupId>
             <artifactId>jsr305</artifactId>
             <version>3.0.2</version>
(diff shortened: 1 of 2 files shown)


lwhite1:
> A lot of Smile algorithms take AttributeDataset
Can you tell me which?

This work looks like a nice implementation of a bridge between a Table and an AttributeDataSet, but once you get to Smile-land the bridge between ADS and models seems to be "out", mostly. Or am I missing something obvious (again)?

OLS was the only model I noticed that directly accepts an ADS as input. (The API is a bit awkward because the position of the response variable is critical to the execution. I did that in a few places but they're mostly gone now I think.) There are many uses of ADS in Test classes for other models, but these are all disassembled into double arrays, etc. before they're used in classifiers or whatever. 

If there are more models where it's useful then we should definitely add this code. And if there isn't, but its the only way to get named vars in regression models, then maybe we should do it anyway.

lwhite1:
ah. ok. I was looking at my current smile (as of a day or two ago) so i
missed those.

PR #315, opened 2018-07-29
Made ColumnType an interface

lwhite1:
Ben, if you could take a look when you some time, it would be greatly appreciated. Also LMK if it seems likely to break anything in client code. 

Thanks very much. 

This code makes ColumnType an interface so it will be easier to create new column types, and perhaps may be possible in the future for library users to add new column types specific to their needs.

Currently, this cannot be done because there are several key statements that switch on column type. It is a future enhancement to eliminate those switch statements by (if possible) moving the requisite logic to the column types.  This would be more object-oriented anyway

The interface subsumes most of the functionality of the enum it replaces, including implementations of methods like values() and valueOf() defined on Enum. The toString is compatible with the enum version. The enum method ordinal() is not supported. 

Thanks for contributing.

## Description

What was changed
Made ColumnType an interface instead of an enum

## Testing
All tests pass.

Did you add a unit test?
no

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/Table.java
+++ b/core/src/main/java/tech/tablesaw/api/Table.java
@@ -607,29 +607,29 @@ public class Table extends Relation implements Iterable<Row> {
     public void addRow(int rowIndex, Table sourceTable) {
         for (int i = 0; i < columnCount(); i++) {
             Column column = column(i);
-            ColumnType type = column.type();
+            final String type = column.type().name();
             switch (type) {
-                case NUMBER:
+                case "NUMBER":
                     NumberColumn numberColumn = (NumberColumn) column;
                     numberColumn.append(sourceTable.numberColumn(i).get(rowIndex));
                     break;
-                case BOOLEAN:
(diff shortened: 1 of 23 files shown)


lwhite1:
merged i can keep making progress.

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -607,29 +607,29 @@ public int compare(int k1, int k2) {
     public void addRow(int rowIndex, Table sourceTable) {
         for (int i = 0; i < columnCount(); i++) {
             Column column = column(i);
-            ColumnType type = column.type();
+            final String type = column.type().name();
```
benmccann: Converting to a String means we lose type safety

lwhite1:
I couldn't think of another way to do it without relying on enums, which
can't be extended by a library user.
It's my hope to remove the switches that rely on types anyway, for the same
reason. I will attempt to put the logic in the column type and use
polymorphism to do the work of the switch statement.

PR #317, opened 2018-07-30
Code Optimization

divyeshjagatiya:
## Description
Convert from array of object to string by (Collectors.joining)

### DIFF
--- a/jsplot/src/main/java/tech/tablesaw/plotly/Utils.java
+++ b/jsplot/src/main/java/tech/tablesaw/plotly/Utils.java
@@ -1,6 +1,7 @@
 package tech.tablesaw.plotly;
 
 import java.util.Arrays;
+import java.util.stream.Collectors;
 
 public class Utils {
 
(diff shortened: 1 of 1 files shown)


lwhite1:
How much time does it save?  Or is this a different kind of optimization?

benmccann:
I'm guessing it's optimization in terms of fewer lines of code. But I think this is going to be slower.

If we care about performance, then this:

    "'" + String.valueOf(d) + "'"

Would be better written as:

    new StringBuilder("'").append(String.valueOf(d)).append("'").toString()

divyeshjagatiya:
you r right

lwhite1:
First, thanks very much. 

This code is **much** cleaner than mine, and normally I would be all for it. In this case, though, I'd rather not add the overhead. Even if you make the change that Ben suggests, you'll be making a new StringBuilder for each element in the array, because map gets called on every element.

divyeshjagatiya:
i will try to make batter than this as per your suggestion.
Thanks,

lwhite1:
Closing unless/until we get a better, fast implementation.

PR #328, opened 2018-08-08
Remove need for packed column types

benmccann:
@lwhite1 I wanted to get your thoughts on this one. I think it simplifies the code a lot by allowing us to remove quite a bit of code, but I wanted to make sure there wasn't anything I was missing as to why we might need to keep these classes


lwhite1:
I wrote those classes to avoid the time and memory overhead of creating millions of objects for each temporal column iterated.

benmccann:
I'm not sure what objects would be created for a temporal column? The temporal columns internally store ints (long in the case of datetime), right?

lwhite1:
Sorry, I should have said "so that users could treat the data  in those columns **as objects** without " the time and memory overhead of creating millions of objects...

With the new version, you have to use the utility classes PackedLocalDate etc, to do anything with the returned value. For example (i didn't check exact syntax):

```
int packedDate = row.getPackedDate("foo");
PackedLocalDate comparisonDate = PackedLocalDate.pack(myLocalDate);
boolean after = PackedLocalDate.isAfter(packedDate, comparisonDate);  
```
Note that this is not completely clear about which date is which in the method isAfter.
Even ignoring that, this approach is less accessible to a user than the wrappers:

```
PackedDate date = row.getPackedDate("foo);
boolean after = date.isAfter(myLocalDate);
```
The second version is more direct, more "object-oriented", and has the advantage in working with code completion. 

So these classes required effort to create (and maintain), but I don't think we have much of a story about high-performance, row-oriented access without them. Unless you can demonstrate that the overhead of iterating with lang.LocalDate is immaterial. I expect (without testing) that those object need to be GCd, however.

benmccann:
You'll have to create millions of `PackedDate` though, won't you? I'm not sure that's any cheaper than creating millions of `LocalDate`

Another thing I'm wondering is we're generally creating millions of objects at some point because `DateColumn` is generally created from `List<LocalDate>`. In that case we're only doing it at creation time instead of iteration time. But another solution to the iteration issue would be for `DateColumn` to use `LocalDate` internally so that it wouldn't have to be recreated during iteration. I don't remember why we store `ints`/`longs` internally instead of `LocalDate`?

lwhite1:
There’s only one packed date for the whole volume.  It only looks like a
regular object.

Datecolumn is generally created from a string[]. I think it’s rarely
created from a list.

We use ints for dates because a LocalDate uses 4 or 5x the amount of memory
and is slower to work with because the cpu is chasing pointers every time
you hand it one.

lwhite1:
That first sentence should say ‘the whole column’. Autocorrect

lwhite1:
If you want to get rid of all the packed date stuff and just use java
objects, feel free.

benmccann:
I will close for now. This seems like a more complicated change than I originally thought

lwhite1:
I was starting to lean the other way.

The current approach has limitations. It works fine if the date is used for
comparison, or to get some value out of it like a month or day-of-week. But
there's no way to represent a date derived from a calculation on a
PackedDate. You have to fall back to LocalDate or use PackedLocalDate's
static utilities to keep working with it.

It could be extended to make that possible, but it would be work. The
work-around now is to make the derived date value a column *before*
iterating over the rows. Too many special cases are bad for usability,
though.

You can certainly argue that having an easy way (LocalDate) for the 80%
usage and a fast way (PackedLocalDate) for internal Tablesaw use and for
anyone who needs very high performance is sufficient.

FWIW, the memory difference between the int representation and the
LocalDate representation is only 3x, not 4-5x as I stated earlier.

I think I will sleep on it for a night or two, but I may just go ahead and
apply these changes.

benmccann:
I do think that simply using `LocalDate` internally would be much easier for us to maintain.

I haven't done any benchmarking around performance / memory. None of the datasets I work with are very big, so I'm never in a place where I would hit those constraints. And usually I just have one or two date columns in a table, so it might only be a 10-20% increase in memory for my particular usage. I'm not sure if other users would feel differently.

lwhite1:
I think my preference is to leave the underlying packed values.
you'd have to convert dates to ints to index them anyway, or indexes could
get as big as the (now bigger) columns.

I think financial apps are pretty finicky about performance for things like
tick data, even if they only use one timestamp per tick.
And FWIW the biggest dataset I use at work has about a dozen dateTime
columns. It tracks the progress of orders, one per row, and every step is
timestamped for analysis.

I don't actually see the packed values as being much work at this point. If
we did get rid of them we might as well get rid of doubleArrays and the
dictionary encoding for strings and use POJOs everywhere. Then we'd have my
first version of tablesaw.

PR #331, opened 2018-08-12
implemented generic typed columns

lwhite1:
Thanks for contributing.

## Description

What was changed

## Testing

Did you add a unit test?
no

### DIFF
--- a/core/src/main/java/tech/tablesaw/aggregate/CrossTab.java
+++ b/core/src/main/java/tech/tablesaw/aggregate/CrossTab.java
@@ -121,7 +121,7 @@ public final class CrossTab {
         NumberColumn pctsColumn = DoubleColumn.create("Percents");
         double sum = countsColumn.sum();
         for (int i = 0; i < countsColumn.size(); i++) {
-            pctsColumn.append(countsColumn.get(i) / sum);
+            pctsColumn.append(countsColumn.getDouble(i) / sum);
         }
         percentTable.addColumns(pctsColumn);
         return percentTable;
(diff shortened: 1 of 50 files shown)


core/src/main/java/tech/tablesaw/aggregate/CrossTab.java
```diff
@@ -144,13 +144,13 @@ private static Table rowPercents(Table xTabCounts) {
...
         for (int i = 0; i < xTabCounts.rowCount(); i++) {
-            float rowTotal = (float) xTabCounts.numberColumn(xTabCounts.columnCount() - 1).get(i);
+            double rowTotal = xTabCounts.numberColumn(xTabCounts.columnCount() - 1).getDouble(i);
 
             for (int c = 1; c < xTabCounts.columnCount(); c++) {
                 if (rowTotal == 0) {
                     pctTable.numberColumn(c).append(Float.NaN);
                 } else {
-                    pctTable.numberColumn(c).append((float) xTabCounts.numberColumn(c).get(i) / rowTotal);
+                    pctTable.numberColumn(c).append(xTabCounts.numberColumn(c).getDouble(i) / rowTotal);
```
benmccann: alternatively you could do `doubleColumn(c).get(i)` which would be a bit more concise

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -82,7 +82,7 @@ private Table(String name) {
      */
     protected Table(String name, Column... columns) {
         this(name);
-        for (Column column : columns) {
+        for (final Column<?> column : columns) {
```
benmccann: there's no need to add `final` throughout this file. I tend to avoid using it except for class members or to enable accessing via a closure in accordance with the google style guide (at least that used to be the rule when I was there)

benmccann:
looks good to me

core/src/main/java/tech/tablesaw/api/DoubleColumn.java
```diff
@@ -438,7 +438,7 @@ public IntComparator rowComparator() {
     }
 
     @Override
-    public double get(final int index) {
+    public Double get(final int index) {
```
benmccann: I think I preferred this returning `double` and having `getObject` to get the `Double`. We want to use `double` most often to avoid boxing,which is why I think it's a good default

lwhite1: That's true but the other columns worked the other way String, Date, Time, etc. all returned the Object when we call get(). It feels like overkill to have them all also implement getObject() when double is the only exception.

PR #334, opened 2018-08-14
Functional methods implemented and tested

hallvard:
Thanks for contributing.

- [v] Tick to sign-off your agreement to the [Developer Certificate of Origin (DCO) 1.1](https://developercertificate.org)

## Description

I've implemented some default methods on Column<T> for functional programming that correspond to Stream methods. In addition, I removed doForEach, since Column now extends Iterable which includes forEach.

## Testing

Corresponding test methods in ColumnTest. Also changed AbstractColumnTest, so it tests forEach (not really necessary, though).

### DIFF
--- a/core/src/main/java/tech/tablesaw/columns/Column.java
+++ b/core/src/main/java/tech/tablesaw/columns/Column.java
@@ -19,13 +19,20 @@ import static tech.tablesaw.selection.Selection.selectNRowsAtRandom;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Optional;
+import java.util.function.BiFunction;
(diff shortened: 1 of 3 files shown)


hallvard:
Add variants of methods that operate on double values.

lwhite1:
this looks good but i need to think about adding more double-specific code to every column. That they all can answer doubles was a hack to get around limitations in smile, where the api is generally looking for a number even for categorical variables. I'm not sure I want to entrench that even further.

hallvard:
I have no opinion, could move the doubles methods to DoubleColumn.

lwhite1:
lets do that. It's better to broaden later if need be than to try to narrow
the interface after people start to use it.

PR #336, opened 2018-08-15
Fix all StringParser generics warnings

benmccann:
Thanks for contributing.

## Description

Fixed all the warnings related to `StringParser`, which was about half of the remaining. 14 left that I will address in a follow up PR

I did a refactoring of cell appending in the `CSVReader` by calling another method to achieve the same thing. This meant that we no longer needed the new method `appendCell` which was recently added since that was the only place it was being called. This allowed for a reduction in the amount of code and doesn't require any generic types and removed most of the warnings

## Testing

N/A. Only refactoring and cleanup

### DIFF
--- a/core/src/main/java/tech/tablesaw/aggregate/AggregateFunction.java
+++ b/core/src/main/java/tech/tablesaw/aggregate/AggregateFunction.java
@@ -6,7 +6,7 @@ import tech.tablesaw.columns.Column;
 /**
  * A partial implementation of aggregate functions to summarize over a numeric column
  */
-public abstract class AggregateFunction<T, C extends Column> {
+public abstract class AggregateFunction<T, C extends Column<?>> {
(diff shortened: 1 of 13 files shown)


lwhite1:
This is definitely cleaner, but one difference between the appendObj() approach as I understand it and the one it's replacing is that appendObj() always appends an object. 

```
column.appendObj(parser.parse(value));
```

The older approach took the original string and handled it differently for each column type. For DoubleColumn, it called parseDouble(), rather than parse, which returned a double without auto-boxing it. 

This could be a lot of extra object creation.  Is there a way to get rid of the lint warnings without making an extra object for every numeric cell?

benmccann:
Ah, good point. I sent a new PR that fixes it without the object creation. It was easier for me to start from scratch than update this one.

Issue #339, opened 2018-08-15
reduce memory requirements


lwhite1:
This can be done in one of three ways. 

We could re-introduce numeric column types like integer and float.
We could compress the data.
We could use other primitive types inside NumericColumn 

Each has benefits. Adding an integer class would not only ensure that ints only use 4 bytes, but also provide automatic formatting for integer values (i.e. no decimal points in display.  

The difficulty is that the current code has made it difficult to add a new numeric column type. Mostly because the combination of the NumberColumn interface and the use of generic types, makes Number column difficult to subclass. The complex hierarchy of interfaces doesn't help, several like (NumberMapFunctions) have double iterators, which conflict with other numeric types.

Compressing the data is a second possibility. The easiest approach is to re-apply the dictionary encoding used in StringColumn to doubleColumn if a column contains relatively few distinct values compared to the total number of values. Since we have now extra data to contend with, this gets less compression than the other approach. It works well with String because most strings use way more than four bytes.

Alternately, we could simply store the values in a different array type inside of NumberColumn. If we detect that data is integral and can be fit in 4 bytes, we use an IntArray behind the scenes, rather than the double array.

PR #342, opened 2018-08-16
Integral columns

lwhite1:
Thanks for contributing.

## Description
I am looking for design feedback here, rather than a detailed look at the code. The code would need to be cleaned up quite a bit before merging. 

This is an experimental branch for testing a method to reduce the memory requirements of large tables of numbers. Earlier versions of Tablesaw used less memory because they supported numeric column types (short, float, int) that used less memory in addition to double and long.

As sketched out in #339, I see three ways to do this. 

- The first is to use compression (perhaps dictionary compression, as we do in StringColumn).

- The second is to bring back the other column subtypes.  I played with this but I couldn't find a clean way to do it given NumberColumn is a Column<Double>. Other subtypes would have to be defined as Integer<Double> as well, unless there's some refactoring or Generics magic that would get around it.

- The third was to, essentially embed the other column types in NumberType. In a way, this is just another form of compression. That's what I tried here.  Currently, the change is visible, the column gets a ColumnType of INTEGER, but potentially it could be hidden The work is just a hack (especially around iterators) but it could be fixed up. 

In this version, DoubleColumn is gone, and NumberColumn can hold either ints or doubles. The format is chosen dynamically based using column type detection. This might be less work than implementing separate columns.  

From a memory standpoint, it reduces the space for integer columns by half. It could be extended to floats as well, where the input was not high-precision.

In the code now, if you attempt to put a float that isn't just an int into a column that has been defined as integer, you get an error, but I think that maybe switching the column type dynamically would be preferable. 

In the longer run, we could use dictionary compression (or another type of compression) when that makes sense.

## Testing

Did you add a unit test?
Nah.


core/src/main/java/tech/tablesaw/columns/numbers/NumberIterator.java
```diff
@@ -0,0 +1,78 @@
...
+
+import it.unimi.dsi.fastutil.doubles.DoubleArrayList;
+import it.unimi.dsi.fastutil.doubles.DoubleIterator;
+import it.unimi.dsi.fastutil.floats.FloatIterator;
+import it.unimi.dsi.fastutil.ints.IntArrayList;
+import it.unimi.dsi.fastutil.ints.IntIterator;
+
+import java.util.Iterator;
+
+public class NumberIterator implements Iterable<Double> {
```
benmccann: Could we split this class into a `DoubleIterator` and `IntIterator`?

docs/userguide/tables.md
```diff
@@ -141,7 +141,7 @@ Often you want just one column, which you can get using *t.column(“columnName
 Since Tablesaw columns are typed, you often need to cast the returned column to something more specific. For example:
 
 ```java
-DoubleColumn dc = (DoubleColumn) t.column();
+DoubleColumn dc = (NumberColumn) t.column();
```
benmccann: lefthand side needs to be updated?

core/src/main/java/tech/tablesaw/api/NumberColumn.java
```diff
@@ -1,268 +1,716 @@
...
+        this.append(val.doubleValue());
+        return this;
+    }
+
+    public NumberColumn append(Integer val) {
+        this.append(val.doubleValue());
+        return this;
+    }
+
+    private NumberColumn(final String name, IntArrayList data) {
```
benmccann: usually convention is to put the constructor as the first method in the class

core/src/main/java/tech/tablesaw/api/NumberColumn.java
```diff
@@ -1,268 +1,716 @@
...
+    /**
+     * Returns the largest ("top") n values in the column
+     * TODO(lwhite): Consider whether this should exclude missing
+     *
+     * @param n The maximum number of records to return. The actual number will be smaller if n is greater than the
+     *          number of observations in the column
+     * @return A list, possibly empty, of the largest observations
+     */
+    public NumberColumn top(final int n) {
+//            final int[] values = intData.toIntArray();
```
benmccann: remove commented out lines?

lwhite1:
I was mainly asking for feedback on the design, but thanks for the
detailed look anyway.

core/src/main/java/tech/tablesaw/columns/numbers/DoubleDataWrapper.java
```diff
@@ -0,0 +1,158 @@
...
+        return data.contains(value);
+    }
+
+    @Override
+    public boolean contains(int value) {
+        return data.contains(value);
+    }
+
+    @Override
+    public DoubleList dataInternal() {
```
benmccann: maybe we should have `intsInternal()` and `doublesInternal()`? I think throwing an `UnsupportedOperationException` would be preferable to returning null because it will be easier for the user to understand what's happening

core/src/main/java/tech/tablesaw/api/NumberColumn.java
```diff
@@ -1,268 +1,716 @@
...
+        }
+        return new NumberColumn(name, new DoubleArrayList(doubles));
+    }
+
+    public static NumberColumn create(final String name, final int[] arr) {
+        final double[] doubles = new double[arr.length];
+        for (int i = 0; i < arr.length; i++) {
+            doubles[i] = arr[i];
+        }
+        return new NumberColumn(name, new DoubleArrayList(doubles));
```
benmccann: Should this use the `IntDataWrapper`?

core/src/main/java/tech/tablesaw/api/NumberColumn.java
```diff
@@ -299,22 +747,339 @@ default double kendalls(NumberColumn otherColumn) {
...
+        return count;
+    }
+
+    /**
+     * Returns a new NumberColumn with only those rows satisfying the predicate
+     *
+     * @param test the predicate
+     * @return a new NumberColumn with only those rows satisfying the predicate
+     */
+    public NumberColumn filter(DoublePredicate test) {
```
benmccann: I wonder if this is an operation we can implement on the data wrappers like `lead` and `lag`

core/src/main/java/tech/tablesaw/columns/numbers/DoubleDataWrapper.java
```diff
@@ -0,0 +1,158 @@
...
+import static tech.tablesaw.api.NumberColumn.MISSING_VALUE;
+
+public class DoubleDataWrapper implements NumericDataWrapper {
+
+    /**
+     * Compares two doubles, such that a sort based on this comparator would sort in descending order
+     */
+    private final DoubleComparator descendingComparator = (o2, o1) -> (Double.compare(o1, o2));
+
+    private DoubleArrayList data;
```
benmccann: this could be made `final`

benmccann:
This could work for data read from a CSV. As soon as you do almost any transformation on the column it will end up being converted to doubles internally. But overall this seems about the right tradeoff to me and more maintainable than the approach in the original version of Tablesaw

I'm trying to think through use cases for other numeric types. Smile takes `double` for data and `int` for labels. I see `float` used for things like deep learning which are computationally intensive and don't need double precision. In both these cases, I imagine you might do some data cleansing, filtering, slicing, and dicing in Tablesaw and then load it into one of those libraries. We don't need to support many complex mathematical operations in Tablesaw with multiple types, but would maybe would want to support some subset of operations like filtering and mapping which allow the user to maintain the data in it's original storage format for cleansing purposes before loading into one of these libraries

I think the main proposal here, which I really like, is to support multiple types, but have all mathematical operations support only `double` so that we don't have to support every combination of primitive types. I'm wondering if we may still wish to have `IntegerColumn`, etc. instead of a single `NumberColumn`. We should have any complex operation return `DoubleColumn` and convert to `double` internally for the majority of operations. But I think if you specifically loaded in `int`s or `float`s from a CSV and wanted to perform some cleansing operations keeping it in that original format that it may be easier to have specific column types.

I also think it'd be nice to have a way to use `int`s/`long`s as a categorical columns or for joining on. I feel that `doubles` don't fulfill that purpose as well because of possible small floating point differences. This is another area where it would actually be nice to have different column types.

core/src/main/java/tech/tablesaw/api/NumberColumn.java
lwhite1: i'm removing this method

core/src/main/java/tech/tablesaw/columns/numbers/DoubleDataWrapper.java
lwhite1: i'm removing this method

core/src/main/java/tech/tablesaw/columns/numbers/DoubleDataWrapper.java
lwhite1: done

core/src/main/java/tech/tablesaw/api/NumberColumn.java
lwhite1: yes. fixed.

core/src/main/java/tech/tablesaw/api/NumberColumn.java
lwhite1: done

lwhite1:
> This could work for data read from a CSV. As soon as you do almost any transformation on the column it will end up being converted to doubles internally. But overall this seems about the right tradeoff to me and more maintainable than the approach in the original version of Tablesaw

Yeah, i agree about conversion to doubles as the default behavior for summarization and mapping. 

> I'm trying to think through use cases for other numeric types. Smile takes double for data and int for labels. I see float used for things like deep learning which are computationally intensive and don't need double precision. In both these cases, I imagine you might do some data cleansing, filtering, slicing, and dicing in Tablesaw and then load it into one of those libraries. 

Yup. I should try DeepLearning4J and write a tutorial. The main benefit to floats, outside of that, is memory. I get a lot of data where either the input data has low precision (2 or 3 digits), or the precision is larger, but is unimportant.  

>We don't need to support many complex mathematical operations in Tablesaw with multiple types, but would maybe would want to support some subset of operations like filtering and mapping which allow the user to maintain the data in it's original storage format for cleansing purposes before loading into one of these libraries
 
One thing i hope to add is a set of conversion methods. For example,
```
DoubleColumn::asIntColumn();
```
Then people can manager their memory even if we don't provide special support for every option. It also suggests a way to provide support with minimal work, which is to wrap existing map functions (for example) doing the math as doubles and returning the results as ints or floats or whatever. 

> I think the main proposal here, which I really like, is to support multiple types, but have all mathematical operations support only double so that we don't have to support every combination of primitive types. 

Yup. I'm also a little concerned also that we might end up overriding Column returning methods as discussed in another thread. This avoids having to do that two or three more times.  And then there's just the fact that columns have a lot of methods generally and more files to edit makes refactoring difficult.

> I'm wondering if we may still wish to have IntegerColumn, etc. instead of a single NumberColumn. We should have any complex operation return DoubleColumn and convert to double internally for the majority of operations. But I think if you specifically loaded in ints or floats from a CSV and wanted to perform some cleansing operations keeping it in that original format that it may be easier to have specific column types.

This is really the key decision. I think i'm getting close to having the "feel" of different column types (There are already using specialized ColumnTypes). If you think it's really important to have true specialized column classes we should do it; I don't want to mess with that class hierarchy again after this. I'm just not sure I see the need right now.

> I also think it'd be nice to have a way to use ints/longs as a categorical columns or for joining on. I feel that doubles don't fulfill that purpose as well because of possible small floating point differences. 

Ah yes, I should look at that. I will try implementing a join for number column when we have ints internally.  I will also implement a float index. I'm thinking of adding an index() method to columns that returns an index. It could delegate to the data wrapper to ensure the right kind of index is created.

It's probably not too much extra work to implement specific column classes now that the data wrappers exist. They will all have the Double type parameter though, unless you know some way around that.

Issue #343, opened 2018-08-17
Copy table rows is boxing all cells


lwhite1:
In Rows.java appendObj instantiates every cell. 

```  
  public static void copyRowsToTable(IntArrayList rows, Table oldTable, Table newTable) {
        for (int i = 0; i < rows.size(); i++) {
            int rowIndex = rows.getInt(i);
            int columnCount = oldTable.columnCount();
            for (int columnIndex = 0; columnIndex < columnCount; columnIndex++) {
                newTable.column(columnIndex).appendObj(oldTable.column(columnIndex).get(rowIndex));
            }
        }
    }
```
This should be copying primitives as this is one of the most common and important operations in the system.

benmccann:
Sent a fix for review: https://github.com/jtablesaw/tablesaw/pull/345

PR #345, opened 2018-08-18
Reduce autoboxing

benmccann:
Thanks for contributing.

## Description

Fix for https://github.com/jtablesaw/tablesaw/issues/343

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/Table.java
+++ b/core/src/main/java/tech/tablesaw/api/Table.java
@@ -36,6 +36,7 @@ import tech.tablesaw.table.Relation;
 import tech.tablesaw.table.Rows;
 import tech.tablesaw.table.StandardTableSliceGroup;
 import tech.tablesaw.table.TableSliceGroup;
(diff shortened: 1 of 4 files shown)


lwhite1:
This still has a few issues. 

- ColumnUtils::append() doesn't cover dates, times, or booleans. 
- The ability to perform primitive copying should be implemented without switching on type. Similar methods were in ColumnType (and implementers) earlier so the approach would be available to new column types created in client code. Please take care not to re-introduce the use of branching logic on Column classes or ColumnType. It's my hope that in the future this will also make it easier to use int and float during copies instead of double. 
- Table::append(Table other) needs to use primitive methods also.
- We have the same issue in TableSliceGroup, and Summarizer, but there it's from using append() rather than appendObj().These cases aren't quite as important as the those invoked through table copying, appending, and filtering.

As a general rule, I don't think avoiding a warning in lint is a good trade off against memory usage. The latter affects everyone and causes crashes, but the former only affects people who compile tablesaw with the xLint flag.

PR #348, opened 2018-08-18
Leverage recent improvements for lightweight IntColumn and FloatColumn

benmccann:
Thanks for contributing.

## Description

These new `IntColumn` and `FloatColumn` classes are tiny (only 150 lines each), so should be easy to maintain. Having top-level classes for these columns has some nice properties. E.g. we can now say that an `IntColumn` is a `CategoricalColumn`, but a `DoubleColumn` and `FloatColumn` are not

I expect we can do some additional cleanup and testing after this, but this is the last major API change I was thinking about for the column types

### DIFF
--- a/core/src/main/java/tech/tablesaw/aggregate/AggregateFunctions.java
+++ b/core/src/main/java/tech/tablesaw/aggregate/AggregateFunctions.java
@@ -7,7 +7,7 @@ import org.apache.commons.math3.stat.descriptive.moment.Skewness;
 import tech.tablesaw.api.BooleanColumn;
 import tech.tablesaw.api.DateColumn;
 import tech.tablesaw.api.DateTimeColumn;
-import tech.tablesaw.api.NumberColumn;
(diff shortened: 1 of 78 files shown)


lwhite1:
I like it, but it needs quite a bit of cleanup.

benmccann:
Great! I did a bunch of cleanup and got the tests passing again. What do you think about merging this and then I'll address any additional cleanup in a follow up PR? Let me know if there was anything specific you had in mind

lwhite1:
I might be missing something, but it seems that I can no longer call:

```
        details.numberColumn("age").set(4, 100);

```
which is, to say, you can't use any method defined in column as foo(T) without a cast, even if there's a default implementation. This seems like a step backwards. Neither the old NumberColumn/DoubleColumn pairing or the NumberColumn with dataWrappers had this constraint.

lwhite1:
I was missing something, apparently. Or my compiler was. It seems fine.

PR #352, opened 2018-08-20
Remove need for NumberIterator

benmccann:
Thanks for contributing.

## Description

It seems like we might not need `NumberIterator` anymore. The only place it was really used was for `NumberFillers`. I'm not sure if we still need `NumberFillers`? It seems like `DoubleColumn.create` already provides this functionality in a more commonly used way


lwhite1:
i think i need a better understanding of how create() replaces fillers.  At a minimum, i didn't see a range iterable equivalent.

PR #353, opened 2018-08-20
More efficient setting of cell values

benmccann:
Thanks for contributing.

## Description

Fixes https://github.com/jtablesaw/tablesaw/issues/343

- `Table.append` and `Rows` functionality now happens without auto-boxing
- Made `select` utilize primitives vs converting to strings and then parsing
- Joining now happens without auto-boxing or converting to strings and then parsing
- SQL reading now utilizes objects rather than converting to strings and then parsing

## Testing

Covered by existing unit tests

### DIFF
--- a/core/src/main/java/tech/tablesaw/table/Rows.java
+++ b/core/src/main/java/tech/tablesaw/table/Rows.java
@@ -14,36 +14,40 @@
 
 package tech.tablesaw.table;
 
-import it.unimi.dsi.fastutil.ints.IntArrayList;
+import javax.annotation.concurrent.Immutable;
+
 import tech.tablesaw.api.ColumnType;
 import tech.tablesaw.api.Table;
+import tech.tablesaw.columns.Column;
+import tech.tablesaw.selection.BitmapBackedSelection;
 import tech.tablesaw.selection.Selection;
-
-import javax.annotation.concurrent.Immutable;
 
 /**
  * A static utility class for row operations
  */
 @Immutable
-public class Rows {
+public final class Rows {
 
     // Don't instantiate
     private Rows() {}
 
-    public static void copyRowsToTable(IntArrayList rows, Table oldTable, Table newTable) {
-        for (int i = 0; i < rows.size(); i++) {
-            int rowIndex = rows.getInt(i);
-            int columnCount = oldTable.columnCount();
-            for (int columnIndex = 0; columnIndex < columnCount; columnIndex++) {
-                newTable.column(columnIndex).appendObj(oldTable.column(columnIndex).get(rowIndex));
-            }
+    /**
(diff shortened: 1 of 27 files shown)


core/src/main/java/tech/tablesaw/table/Rows.java
```diff
@@ -14,36 +14,40 @@
...
+     */
+    public static void copyRowsToTable(Selection rows, Table oldTable, Table newTable) {
+        copyRowsToTable(rows.toArray(), oldTable, newTable);
+    }
+
+    @SuppressWarnings({"rawtypes","unchecked"})
+    public static void copyRowsToTable(int[] rows, Table oldTable, Table newTable) {
+        for (int columnIndex = 0; columnIndex < oldTable.columnCount(); columnIndex++) {
+            Column oldColumn = oldTable.column(columnIndex);
+            newTable.column(columnIndex).append(oldColumn.subset(rows));
```
lwhite1: This is much better, but i wonder if there's a way to do it without creating the intermediate column in oldColumn.subset(rows). Not a big deal, but the append method that takes a column and a row might be used here by iterating on the array, no? 

Also, maybe the same approach would work on the selection version of copyRowsToTable(), avoiding both the conversion to an int[] and the subsequent column creation step in subset(). IDK, it's really early here, so....

core/src/test/java/tech/tablesaw/SortTest.java
```diff
@@ -99,7 +101,7 @@ public void testAscendingWithPlusSign() {
         compareTables(expectedResults, sortedTable);
     }
 
-    @Test(expected = ComparisonFailure.class)
+    @Ignore
```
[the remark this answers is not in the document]
benmccann: This test was always failing, which is what the `expected = ComparisonFailure` indicated. It's more standard to use `@Ignore` for disabled tests though. `expected` is intended to be used when you want to test that an `Exception` is thrown. It was failing because of https://github.com/jtablesaw/tablesaw/issues/354

core/src/main/java/tech/tablesaw/table/Rows.java
benmccann: That's true. That would save memory

The append method taking a column and a row would call `Preconditions.checkArgument(column.type() == this.type());` on ever cell, which might take a bit more CPU. I'm not sure if that's an issue at all?

lwhite1:
i already made these changes.

lwhite1:
to answer your specific question, i prefer to have thing take slightly
longer than to have an increased risk of OOMemory errors caused by memory
spiking during operations, but I think in this case that concern is pretty
remote.  I think in general, using a loop is a little better overall so
that's what i did

benmccann:
Sounds good. Thanks!

PR #371, opened 2018-08-29
Fix for joining multiple tables

gregorco:
Thanks for contributing.

## Description

This fixes the implementation to join on multiple tables.  Looked hard for a more clever/recursive solution but could not see one.  At one point rejected a slightly more elegant solution that reset the existing DataFrameJoiner's Table, but since I had to remove the Table's "final" modifier I suspected that that would not be acceptable.  This solution relies on manually initializing the index used for renaming duplicate columns, but I could not see a more elegant solution.  Please let me know if I missed something.

## Testing

Added three junit tests - which all pass.

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/Table.java
+++ b/core/src/main/java/tech/tablesaw/api/Table.java
@@ -572,7 +572,7 @@ public class Table extends Relation implements Iterable<Row> {
             column(i).appendObj(sourceTable.column(i).get(rowIndex));
         }
     }
-    
+
     public void addRow(Row row) {
         for (int i = 0; i < row.columnCount(); i++) {
             column(i).appendObj(row.getObject(i));
--- a/core/src/main/java/tech/tablesaw/api/Table.java
+++ b/core/src/main/java/tech/tablesaw/api/Table.java
@@ -866,20 +866,10 @@ public class Table extends Relation implements Iterable<Row> {
 
     /**
      * Join on the given {@code columnName}  
-     * @param columnName			The column name to join on
+     * @param columnName            The column name to join on
      */
     public DataFrameJoiner join(String columnName) {
         return new DataFrameJoiner(this, columnName);
-    }
-
-    /**
-     * Join on the given {@code columnName} with {@code joinTableInitialId} as the initial
-     * index used when renaming duplicate columns. 
-     * @param columnName			The column name to join on
-     * @param joinTableInitialId	The initial index used when naming duplicate columns, e.g. "T3.Age"
-     */
-    public DataFrameJoiner join(String columnName, int joinTableInitialId) {
-    	return new DataFrameJoiner(this, columnName, joinTableInitialId);
     }
 
     public Table missingValueCounts() {
--- a/core/src/main/java/tech/tablesaw/api/Table.java
+++ b/core/src/main/java/tech/tablesaw/api/Table.java
@@ -866,7 +866,7 @@ public class Table extends Relation implements Iterable<Row> {
 
     /**
      * Join on the given {@code columnName}  
-     * @param columnName            The column name to join on
+     * @param columnName  The column name to join on
      */
     public DataFrameJoiner join(String columnName) {
         return new DataFrameJoiner(this, columnName);
--- a/core/src/main/java/tech/tablesaw/api/Table.java
+++ b/core/src/main/java/tech/tablesaw/api/Table.java
@@ -864,8 +864,22 @@ public class Table extends Relation implements Iterable<Row> {
         return groupingColumn.countByCategory();
     }
 
+    /**
+     * Join on the given {@code columnName}  
+     * @param columnName			The column name to join on
+     */
     public DataFrameJoiner join(String columnName) {
         return new DataFrameJoiner(this, columnName);
+    }
+
+    /**
+     * Join on the given {@code columnName} with {@code joinTableInitialId} as the initial
+     * index used when renaming duplicate columns. 
+     * @param columnName			The column name to join on
+     * @param joinTableInitialId	The initial index used when naming duplicate columns, e.g. "T3.Age"
+     */
+    public DataFrameJoiner join(String columnName, int joinTableInitialId) {
+    	return new DataFrameJoiner(this, columnName, joinTableInitialId);
     }
 
     public Table missingValueCounts() {
--- a/core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
+++ b/core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
@@ -64,7 +64,7 @@ public class DataFrameJoiner {
             joined = joinInternal(joined, currT, column.name(), false, allowDuplicateColumnNames);
         }
         return joined;
-    }    
+    }
 
     /**
      * Joins the joiner to the table2, using the given column for the second table and returns the resulting table
--- a/core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
+++ b/core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
@@ -59,7 +59,7 @@ public class DataFrameJoiner {
      */
     public Table inner(boolean allowDuplicateColumnNames, Table... tables) {
         Table joined = table;
-        
+
         for (Table currT : tables) {
             joined = joinInternal(joined, currT, column.name(), false, allowDuplicateColumnNames);
         }
@@ -89,7 +89,7 @@ public class DataFrameJoiner {
     public Table inner(Table table2, String col2Name, boolean allowDuplicateColumnNames) {
         return joinInternal(table2, col2Name, false, allowDuplicateColumnNames);
     }
-    
+
     private Table joinInternal(Table table2, String col2Name, boolean outer, boolean allowDuplicates) {
         return joinInternal(table, table2, col2Name, outer, allowDuplicates);
     }
@@ -101,7 +101,7 @@ public class DataFrameJoiner {
         }
 
         Table result = emptyTableFromColumns(table1, table2, col2Name);
-        
+
         // Need to use the column from table1 that is the same column originally
         // defined for this DataFrameJoiner. Column names must be unique within the
         // same table, so use the original column's name to get the corresponding
@@ -216,7 +216,7 @@ public class DataFrameJoiner {
 
     private void renameColumnsWithDuplicateNames(Table table1, Table table2, String col2Name) {
         String table2Alias = TABLE_ALIAS + joinTableId.getAndIncrement();
-        
+
         for (Column<?> table2Column : table2.columns()) {
             String columnName = table2Column.name();
             if (table1.columnNames().contains(columnName)
@@ -348,7 +348,7 @@ public class DataFrameJoiner {
                     "Joining is supported on numeric, string, and date-like columns. Column "
                             + column.name() + " is of type " + column.type());
         }
-        
+
         Table table2OnlyRows = table2.where(selection);
         CategoricalColumn<?> joinColumn = table2OnlyRows.categoricalColumn(col2Name);
         table2OnlyRows.removeColumns(joinColumn);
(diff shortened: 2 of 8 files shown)


gregorco:
I've committed fixes for the first four issues that were reported in the review of my Pull Request.  The fifth issue I left alone for now since that was how it was when I forked.  Not sure yet if I'm supposed to request that my latest changes be reviewed again from this comment - or whether procedure is for me to submit a separate Pull Request?

benmccann:
Can you replace the tabs with spaces so that the formatting is consistent with the existing codebase?

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -864,10 +864,24 @@ public Table countBy(CategoricalColumn<?> groupingColumn) {
...
         return new DataFrameJoiner(this, columnName);
     }
 
+    /**
+     * Join on the given {@code columnName} with {@code joinTableInitialId} as the initial
+     * index used when renaming duplicate columns. 
+     * @param columnName			The column name to join on
+     * @param joinTableInitialId	The initial index used when naming duplicate columns, e.g. "T3.Age"
+     */
+    public DataFrameJoiner join(String columnName, int joinTableInitialId) {
```
benmccann: I don't like the idea of making `joinTableInitialId` part of the public interface as it seems like it's exposing too much of the internals. Can we come up with another solution like checking if the column exists after incrementing the ID and then incrementing again if necessary?

gregorco:
Thanks for the review.  Wasn’t too keen on that change to expose the id either. I’ll work on those changes. 

Gregor

gregorco:
@benmccann  I was able to avoid exposing the ID as desired.  This solution follows your suggestion.  Thanks again.

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -864,6 +864,10 @@ public Table countBy(CategoricalColumn<?> groupingColumn) {
         return groupingColumn.countByCategory();
     }
 
+    /**
+     * Join on the given {@code columnName}  
+     * @param columnName            The column name to join on
```
benmccann: this is a lot of spaces to add without a reason like aligning to something else.

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -41,6 +42,20 @@ public DataFrameJoiner(Table table, String column) {
         this.column = table.categoricalColumn(column);
     }
 
+    /**
+     * Constructor.  Takes an initial value to start the joinTableId.  Useful during
+     * multiple table joins so that each new table joined can have its duplicate columns
+     * distinguished from the other tables.
+     * @param table                    The table to join on
```
benmccann: these are not aligned. it's also a lot of spaces

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -41,6 +42,20 @@ public DataFrameJoiner(Table table, String column) {
...
+     * @param columnName            The column name to join on
+     * @param joinTableInitialId    The initial index used when naming duplicate columns, e.g. "T3.Age"
+     */
+    public DataFrameJoiner(Table table, String columnName, int joinTableInitialId) {
```
benmccann: this constructor appears to be unused. can it be removed

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
benmccann: these are not aligned. it's also a lot of spaces

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -59,12 +74,20 @@ public Table inner(Table... tables) {
      */
     public Table inner(boolean allowDuplicateColumnNames, Table... tables) {
         Table joined = table;
-        for (Table table2 : tables) {
-          joined = inner(table2, column.name(), allowDuplicateColumnNames);
+        
+        for (int i=0; i<tables.length; i++) {
```
benmccann: can you add a space before and after the `=` and `<` for consistency with the rest of the code?

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -205,7 +228,20 @@ private Table joinInternal(Table table2, String col2Name, boolean outer, boolean
 
     private void renameColumnsWithDuplicateNames(Table table2, String col2Name) {
         String table2Alias = TABLE_ALIAS + joinTableId.getAndIncrement();
-
+        List<String> colNames = table.columnNames();
+        boolean foundNextIndex = false;
+        // loop on parent table's column names until the new joinTableId is not 
+        // already being used to prefix any of them 
+        for (int i=0; !foundNextIndex && i<colNames.size(); i++) {
```
benmccann: can you add a space before and after the `<`?

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -486,8 +522,8 @@ private void crossProduct(Table destination, Table table1, Table table2) {
     private void withMissingLeftJoin(Table destination, Table table1) {
         for (int c = 0; c < destination.columnCount(); c++) {
             if (c < table1.columnCount()) {
-        	Column t1Col = table1.column(c);
-        	destination.column(c).append(t1Col);
+            Column t1Col = table1.column(c);
```
benmccann: this should be indented past the `if`

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -504,16 +540,16 @@ private void withMissingRightJoin(Table destination, CategoricalColumn joinColum
         int t2StartCol = destination.columnCount() - table2.columnCount();
         for (int c = 0; c < destination.columnCount(); c++) {
             if (destination.column(c).name().equalsIgnoreCase(joinColumn.name())) {
-        	destination.column(c).append(joinColumn);
+            destination.column(c).append(joinColumn);
```
benmccann: indent

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -504,16 +540,16 @@ private void withMissingRightJoin(Table destination, CategoricalColumn joinColum
...
                 continue;
             }
             if (c < t2StartCol) {
                 for (int r2 = 0; r2 < table2.rowCount(); r2++) {
                     destination.column(c).appendMissing();
                 }
             } else {
-        	Column t2Col = table2.column(c - t2StartCol);
-        	destination.column(c).append(t2Col);
+            Column t2Col = table2.column(c - t2StartCol);
```
benmccann: indent

gregorco:
@benmccann  Thanks for the review.  I just submitted the corrections you noted.

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -40,7 +41,7 @@ public DataFrameJoiner(Table table, String column) {
         this.table = table;
         this.column = table.categoricalColumn(column);
     }
-
+
```
benmccann: nit: trailing whitespace

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -59,12 +60,20 @@ public Table inner(Table... tables) {
      */
     public Table inner(boolean allowDuplicateColumnNames, Table... tables) {
         Table joined = table;
-        for (Table table2 : tables) {
-          joined = inner(table2, column.name(), allowDuplicateColumnNames);
+        
+        for (int i = 0; i < tables.length; i++) {
+            Table currT = tables[i];
+            // if first iteration then join to initial table
+            if (joined.equals(table)) {
```
benmccann: checking `if (i == 0)` might be clearer

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -59,12 +60,20 @@ public Table inner(Table... tables) {
      */
     public Table inner(boolean allowDuplicateColumnNames, Table... tables) {
         Table joined = table;
-        for (Table table2 : tables) {
-          joined = inner(table2, column.name(), allowDuplicateColumnNames);
+        
+        for (int i = 0; i < tables.length; i++) {
```
benmccann: It's not clear to me why the previous code doesn't work here. can you explain what the issue was?

gregorco: From what I could determine the result wasn't accumulative.  The final value of the "joined" variable returned to the caller was only the join with the last table in the tables list.  There wasn't a junit test for it, and when I added one I saw that it only joined on the last table.

benmccann: Oh, I see why now. It's because `joinInternal` uses `table` as the base table to join off of. Can you add a first parameter `table1` to `joinInternal` and replaces all references to `table` in that function with references to `table1`? Then you can pass `joined` here. And you can keep the counter and probably no longer need the extra logic in `renameColumnsWithDuplicateNames`

gregorco: Knew there was probably a more elegant solution than my brute force one.  Thanks.  I'll work on it.

gregorco:
@benmccann Codacy didn't like the longer joinInternal method.  It also caught an unnecessary import.  We need the longer joinInternal method, but should I submit a change to remove the import, or do it in my upcoming PR for multi-column support?

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -59,11 +60,12 @@ public Table inner(Table... tables) {
...
         Table joined = table;
-        for (Table table2 : tables) {
-          joined = inner(table2, column.name(), allowDuplicateColumnNames);
+        
+        for (Table currT : tables) {
+            joined = joinInternal(joined, currT, column.name(), false, allowDuplicateColumnNames);
         }
         return joined;
-    }
+    }
```
benmccann: nit: trailing whitespace

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -88,21 +90,31 @@ public Table inner(Table table2, String col2Name) {
     public Table inner(Table table2, String col2Name, boolean allowDuplicateColumnNames) {
         return joinInternal(table2, col2Name, false, allowDuplicateColumnNames);
     }
-
+
```
benmccann: nit: trailing whitespace

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -203,12 +215,12 @@ private Table joinInternal(Table table2, String col2Name, boolean outer, boolean
         return result;
     }
 
-    private void renameColumnsWithDuplicateNames(Table table2, String col2Name) {
+    private void renameColumnsWithDuplicateNames(Table table1, Table table2, String col2Name) {
         String table2Alias = TABLE_ALIAS + joinTableId.getAndIncrement();
-
+
```
benmccann: nit: trailing whitespace

benmccann:
Yes, please remove the unneccesary import that this PR is adding. Don't worry about the method being long since that's not a result of this change

Thank you for this fix! This looks pretty good to me now

gregorco:
Thanks.  Will do.  

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

benmccann:
Thanks @gregorco !

gregorco:
Thank you.   I’m better prepared now for my next PR.  

Gregor

PR #377, opened 2018-09-05
Add support to join tables on multiple columns.

gregorco:
Thanks for contributing.

## Description

Enhanced API to accept multiple names of columns to join on.  Supported for all joins: inner, leftOuter, rightOuter, and fullOuter.  Added javadoc where missing for public methods.

## Testing

Added dozens of junit tests to cover all changed API.  Ran coverage tool in eclipse to ensure testing/exercising of all changes was being done.  Reached 100% coverage for all but one method.  Goal to reach 100% coverage exposed failure of LONG types to be parse from cvs input so added missing support for that.

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/Table.java
+++ b/core/src/main/java/tech/tablesaw/api/Table.java
@@ -572,7 +572,7 @@ public class Table extends Relation implements Iterable<Row> {
             column(i).appendObj(sourceTable.column(i).get(rowIndex));
         }
     }
-    
+
     public void addRow(Row row) {
         for (int i = 0; i < row.columnCount(); i++) {
             column(i).appendObj(row.getObject(i));
@@ -872,7 +872,7 @@ public class Table extends Relation implements Iterable<Row> {
     public DataFrameJoiner join(String... columnNames) {
         return new DataFrameJoiner(this, columnNames);
 	}
-    
+
     public Table missingValueCounts() {
         return summarize(columnNames(), countMissing).apply();
     }
--- a/core/src/main/java/tech/tablesaw/api/Table.java
+++ b/core/src/main/java/tech/tablesaw/api/Table.java
@@ -572,7 +572,7 @@ public class Table extends Relation implements Iterable<Row> {
             column(i).appendObj(sourceTable.column(i).get(rowIndex));
         }
     }
-    
+
     public void addRow(Row row) {
         for (int i = 0; i < row.columnCount(); i++) {
             column(i).appendObj(row.getObject(i));
--- a/core/src/main/java/tech/tablesaw/api/Table.java
+++ b/core/src/main/java/tech/tablesaw/api/Table.java
@@ -866,8 +866,8 @@ public class Table extends Relation implements Iterable<Row> {
 
     /**
      * Returns a new DataFrameJoiner initialized with multiple {@code columnNames}
-     * @param columnNames	Name of the columns to join on.
-     * @return 				The new DataFrameJoiner
+     * @param columnNames   Name of the columns to join on.
+     * @return              The new DataFrameJoiner
      */
     public DataFrameJoiner join(String... columnNames) {
         return new DataFrameJoiner(this, columnNames);
--- a/core/src/main/java/tech/tablesaw/api/Table.java
+++ b/core/src/main/java/tech/tablesaw/api/Table.java
@@ -866,20 +866,10 @@ public class Table extends Relation implements Iterable<Row> {
 
     /**
      * Join on the given {@code columnName}  
-     * @param columnName			The column name to join on
+     * @param columnName            The column name to join on
      */
     public DataFrameJoiner join(String columnName) {
         return new DataFrameJoiner(this, columnName);
-    }
-
-    /**
-     * Join on the given {@code columnName} with {@code joinTableInitialId} as the initial
-     * index used when renaming duplicate columns. 
-     * @param columnName			The column name to join on
-     * @param joinTableInitialId	The initial index used when naming duplicate columns, e.g. "T3.Age"
-     */
-    public DataFrameJoiner join(String columnName, int joinTableInitialId) {
-    	return new DataFrameJoiner(this, columnName, joinTableInitialId);
     }
 
     public Table missingValueCounts() {
--- a/core/src/main/java/tech/tablesaw/api/Table.java
+++ b/core/src/main/java/tech/tablesaw/api/Table.java
@@ -866,7 +866,7 @@ public class Table extends Relation implements Iterable<Row> {
 
     /**
      * Join on the given {@code columnName}  
-     * @param columnName            The column name to join on
+     * @param columnName  The column name to join on
      */
     public DataFrameJoiner join(String columnName) {
         return new DataFrameJoiner(this, columnName);
--- a/core/src/main/java/tech/tablesaw/api/Table.java
+++ b/core/src/main/java/tech/tablesaw/api/Table.java
@@ -572,7 +572,7 @@ public class Table extends Relation implements Iterable<Row> {
             column(i).appendObj(sourceTable.column(i).get(rowIndex));
         }
     }
-
+    
     public void addRow(Row row) {
         for (int i = 0; i < row.columnCount(); i++) {
             column(i).appendObj(row.getObject(i));
@@ -865,13 +865,14 @@ public class Table extends Relation implements Iterable<Row> {
     }
 
     /**
-     * Join on the given {@code columnName}  
-     * @param columnName  The column name to join on
+     * Returns a new DataFrameJoiner initialized with multiple {@code columnNames}
+     * @param columnNames	Name of the columns to join on.
+     * @return 				The new DataFrameJoiner
      */
-    public DataFrameJoiner join(String columnName) {
-        return new DataFrameJoiner(this, columnName);
-    }
-
+    public DataFrameJoiner join(String... columnNames) {
+        return new DataFrameJoiner(this, columnNames);
+	}
+    
     public Table missingValueCounts() {
         return summarize(columnNames(), countMissing).apply();
     }
--- a/core/src/main/java/tech/tablesaw/api/Table.java
+++ b/core/src/main/java/tech/tablesaw/api/Table.java
@@ -864,8 +864,22 @@ public class Table extends Relation implements Iterable<Row> {
         return groupingColumn.countByCategory();
     }
 
+    /**
+     * Join on the given {@code columnName}  
+     * @param columnName			The column name to join on
+     */
     public DataFrameJoiner join(String columnName) {
         return new DataFrameJoiner(this, columnName);
+    }
+
+    /**
+     * Join on the given {@code columnName} with {@code joinTableInitialId} as the initial
+     * index used when renaming duplicate columns. 
+     * @param columnName			The column name to join on
+     * @param joinTableInitialId	The initial index used when naming duplicate columns, e.g. "T3.Age"
+     */
+    public DataFrameJoiner join(String columnName, int joinTableInitialId) {
+    	return new DataFrameJoiner(this, columnName, joinTableInitialId);
     }
 
     public Table missingValueCounts() {
--- a/core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
+++ b/core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
@@ -37,8 +37,8 @@ public class DataFrameJoiner {
     private static final String TABLE_ALIAS = "T";
 
     private final Table table;
-    private CategoricalColumn<?>[] columns;
-    private String[] columnNames;
+    private final CategoricalColumn<?>[] columns;
+    private final String[] columnNames;
     private AtomicInteger joinTableId = new AtomicInteger(2);
 
     /**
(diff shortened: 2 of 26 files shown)


core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -572,7 +572,7 @@ public void addRow(int rowIndex, Table sourceTable) {
             column(i).appendObj(sourceTable.column(i).get(rowIndex));
         }
     }
-
+
```
benmccann: unnecessary whitespace

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -865,13 +865,14 @@ public Table countBy(CategoricalColumn<?> groupingColumn) {
...
+     * @return 				The new DataFrameJoiner
      */
-    public DataFrameJoiner join(String columnName) {
-        return new DataFrameJoiner(this, columnName);
-    }
-
+    public DataFrameJoiner join(String... columnNames) {
+        return new DataFrameJoiner(this, columnNames);
+	}
+
```
benmccann: unnecessary whitespace

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -434,42 +487,62 @@ public Table rightOuter(boolean allowDuplicateColumnNames, Table... tables) {
...
 
         // reverse the columns
         Table result = Table.create(leftOuter.name());
+        // loop on table that was originally first (left) and add the left-joined matching columns by name
         for (String name : table.columnNames()) {
-            result.addColumns(leftOuter.column(name));
+            try {
+                result.addColumns(leftOuter.column(name));
+            } catch (IllegalStateException e) {
+                System.out.println("NOTE: DataFrameJoiner.rightOuter(): skipping left table's column,'"
```
benmccann: We should either throw an exception here or do logging with slf4j rather than printing to std out

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -26,19 +27,32 @@
...
+     * Constructor.
+     * @param table       The table to join on
+     * @param columnNames The column names to join on
+     */
+    public DataFrameJoiner(Table table, String... columnNames) {
         this.table = table;
-        this.column = table.categoricalColumn(column);
+        columns = new CategoricalColumn<?>[columnNames.length];
+        this.columnNames = columnNames;
+        for(int i = 0; i < this.columnNames.length; i++) {
```
benmccann: add a space after `for`

benmccann: Could simplify this as `for (String colName : columnNames)`

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -85,142 +112,145 @@ public Table inner(Table table2, String col2Name) {
...
+                    rowBitMapOneCol = index.get(value);
+                } else if (type instanceof ShortColumnType) {
+                    ShortIndex index = new ShortIndex(table2.shortColumn(col2Name));
+                    ShortColumn col1 = (ShortColumn) table1Column;
+                    short value = col1.getShort(ri);
+                    rowBitMapOneCol = index.get(value);
                 } else {
-                    crossProduct(result, table1Rows, table2Rows);
+                    throw new IllegalArgumentException(
+                            "Joining is supported on numeric, string, and date-like columns. Column "
```
benmccann: I might just say categorical columns here because double columns are numeric, but not categorical, which makes this message a bit misleading

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -85,142 +112,145 @@ public Table inner(Table table2, String col2Name) {
...
-                long value = col1.getLong(i);
-                Table table1Rows = table1.where(Selection.with(i));
-                Table table2Rows = table2.where(index.get(value));
-                table2Rows.removeColumns(col2Name);
-                if (outer && table2Rows.isEmpty()) {
-                    withMissingLeftJoin(result, table1Rows);
-                } else {
-                    crossProduct(result, table1Rows, table2Rows);
+                // combine Selection's into one big AND Selection
+                if(rowBitMapOneCol != null) {
```
benmccann: add a space after `if`, `?`, and `:`

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -231,135 +261,141 @@ private String newName(String table2Alias, String columnName) {
...
-            IntColumn col2 = (IntColumn) table2.column(col2Name);
-            for (int i = 0; i < col2.size(); i++) {
-                int value = col2.getInt(i);
-                if (index.get(value).isEmpty()) {
-                    selection.add(i);
+        for (Row row : table2) {
+            int ri = row.getRowNumber();
+            Selection rowBitMapMultiCol = null;
+
+            for(int i = 0; i < columns.length; i++) {
```
benmccann: add a space after `for`

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -231,135 +261,141 @@ private String newName(String table2Alias, String columnName) {
...
-            }
-        } else if (type instanceof LongColumnType) {
-            LongIndex index = new LongIndex(result.intColumn(col2Name));
-            LongColumn col2 = (LongColumn) table2.column(col2Name);
-            for (int i = 0; i < col2.size(); i++) {
-                long value = col2.getLong(i);
-                if (index.get(value).isEmpty()) {
-                    selection.add(i);
+                // combine Selections into one big AND Selection
+                if(rowBitMapOneCol != null) {
```
benmccann: space after `if`, `?`, and `:`

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -231,135 +261,141 @@ private String newName(String table2Alias, String columnName) {
...
             }
-        } else if (type instanceof ShortColumnType) {
-            ShortIndex index = new ShortIndex(result.shortColumn(col2Name));
-            ShortColumn col2 = (ShortColumn) table2.column(col2Name);
-            for (int i = 0; i < col2.size(); i++) {
-                short value = col2.getShort(i);
-                if (index.get(value).isEmpty()) {
-                    selection.add(i);
-                }
+            if(rowBitMapMultiCol.isEmpty()) {
```
benmccann: space after `if`

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -231,135 +261,141 @@ private String newName(String table2Alias, String columnName) {
...
+                    long value = col2.getLong(ri);
+                    rowBitMapOneCol = index.get(value);
+                } else if (type instanceof ShortColumnType) {
+                    ShortIndex index = new ShortIndex(result.shortColumn(col2Name));
+                    ShortColumn col2 = (ShortColumn) table2.column(col2Name);
+                    short value = col2.getShort(ri);
+                    rowBitMapOneCol = index.get(value);
+                } else {
+                    throw new IllegalArgumentException(
+                            "Joining is supported on numeric, string, and date-like columns. Column "
```
benmccann: would change message to categorical as above

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -511,21 +584,27 @@ private void withMissingLeftJoin(Table destination, Table table1) {
...
+            for (CategoricalColumn joinColumn : joinColumns) {
+                if (destination.column(c).name().equalsIgnoreCase(joinColumn.name())) {
+                    destination.column(c).append(joinColumn);
+                    addedJoinColumns = true;
+                }
             }
-            if (c < t2StartCol) {
-                for (int r2 = 0; r2 < table2.rowCount(); r2++) {
-                    destination.column(c).appendMissing();
+            if(!addedJoinColumns) {
```
benmccann: space after if

benmccann:
Wow, this is great!

gregorco:
@benmccann  Thanks for the review.  I'll work on those changes.

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
gregorco: If message was changed to say that only categorical columns are supported then BooleanColumn should already be accepted, but trying to join on a boolean column will currently filter to this else condition; I got this condition to be exercised with a junit test that joined on a boolean column.   BooleanColumn implements CategoricalColumn though so should an if condition be added here for BooleanColumnType?

lwhite1:
Joining on Boolean columns Isnt supported.  If it was done on the Boolean
col only you would get n squared rows frequently

gregorco:
Thanks for the clarification.  Since BooleanColumn implements CategoricalColumn though, should the message in that IllegalArgumentException be changed - or left as is?

benmccann:
Maybe change "numeric" to "integral" ?

gregorco:
Sounds good.  Thanks. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
gregorco: I don't yet see any use of slf4j in this project.  Would this be the first use?  If so, is log4j preferred over logback?  I could introduce either one but I've gotten the impression that logback is gaining over log4j in popularity.

gregorco:
@benmccann My app used to be able to join on a DoubleColumn, but now as I'm integrating recent changes into it, I'm seeing a failure that the DoubleColumn I'm trying to join on can't be cast to a CategoricalColumn.   I know I had the join working in the past.  At what point was that ability changed/restricted?  Is there any workaround to allow a join on a DoubleColumn?

lwhite1:
it changed in master, i think.

(btw, I think this should be in a separate issue because it has nothing to
do with joining on multiple columns)

There was a concern that joining on double was a bit dangerous due to
rounding issues. We added joining on doubles when we didn't have any other
numeric column types. Usually people want to join on an int column.

Do you need to join on double or can your data be stored in an int column
before joining?

i'm not sure we should really prevent people from trying to join on double.
I'm also a bit uncertain that we should prevent people from joining on
booleans, because if they blow up their app it's on them. Perhaps we should
allow joining on any column type. Fewer special cases are better.

gregorco:
@benmccann Don't know if you received notice of the reply comments I added yesterday regarding your comments for the slf4j and simplified for-loop suggestions.  I could add a logging mechanism, but am wondering if logback would be preferred over log4j.  The classic for-loop seemed to be the best solution for that situation.  Let me know what you think.  Thanks.

lwhite1:
We don't use any library currently. Logging in java is a mess. In larger
projects you get multiple logging libraries all doing their own thing
because the libraries add them. I don't know that adding one is an
improvement or not.

I would prefer, I think, to write (relatively few) warnings/informational
output to standard out, and throw exceptions for real issues, but I'm open
to discussion. In the example code above, if the Illegal state causes real
issues we should let the client code deal with it.

gregorco:
In the above code the exception is anticipated and not a real concern.  The informational message isn't necessary, so I can remove it altogether.

benmccann:
80%+ of Java libraries today use slf4j-api, which is a facade that delegates to other logging libraries. The end user can then pick any logging implementation they want to use and the log messages will be redirected to that underlying implementation. I definitely would not introduce any other library like log4j, logback, etc. because there's such widespread fragmentation between logging implementations that it will cause problems for some subset of users. But as long as you use slf4j-api it will work with any implementation that the user may choose.

The thing I don't like about writing to std out is that the end user has no control over configuring the output

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
benmccann: ok, no need then. I didn't see that `i` was being used anywhere else

lwhite1:
yeah, in hindsight, i don't think we write to sout anywhere unless it's a
stray debugging statement that needs to be deleted. I'm trying to do to
many things at once to provide good feedback this morning.

Still not sure how i feel about adding slf4j as a dependency, or
philosophically how i feel about whether a library like this should be
doing logging.

gregorco:
The message that I output can simply be removed.  The exception was anticipated so not a real concern or need for action.  

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
gregorco: You're right - it's not being used anywhere else in that immediate method - but later on in the processing in another method the logic relies on the same index variable coordinating the name to the column.  This seemed like the simplest/cleanest solution.  Thanks.  I'll prepare and submit a new update with fixes for the other changes soon.

gregorco:
Thank you @benmccann .  I'll work on these changes.

gregorco:
@benmccann , @lwhite1  I submitted the changes, but the travis-ci build failed with no clear reason.  I don't see a way for me to tell it to restart the build.  Are you able to restart the build?

lwhite1:
I can do that    Travis is flakey

gregorco:
Looks like it tried again by itself already?  I saw this morning that the build had completed before your comment that you'd restart it.  Did you restart it earlier, or did it do so by itself?

lwhite1:
It ran by itself.  All good. I will look at your pr as soon as time
allows.  Thanks very much

lwhite1:
Thanks very much, @gregorco

PR #389, opened 2018-09-11
Intcol2

lwhite1:
Thanks for contributing.

## Description

Merged Short Column and Integer Column. IntegerColumn now has three internal storage types, byte, short, and int. LongColumns remain independent. 

## Testing

Some tests were added.


benmccann:
I'm on vacation, so only have my phone. This PR is too big to review from my phone. If you don't mind waiting until the 23rd I can look it over when I return.

benmccann:
I see a `DoubleDataWrapper` being added here, but not used anywhere. Are there plans to use it later?

PR #391, opened 2018-09-18
Fix #390

### LINKED ISSUE #390: java.lang.OutOfMemoryError when appending a column to itself
Appending a column to itself causes a `java.lang.OutOfMemoryError`

Code example:

```java
final DoubleColumn c1 = DoubleColumn.create("c1", new double[]{42});
final DoubleColumn c2 = DoubleColumn.create("c2", new double[]{9001});

final DoubleColumn c1Copy = c1.copy();
final DoubleColumn c2Copy = c2.copy();

c1.append(c2); // works
c2.append(c1); // works

c1.append(c1Copy); // works
c2.append(c2Copy); // works

c1.append(c1); // java.lang.OutOfMemoryError
c2.append(c2); // java.lang.OutOfMemoryError
```

Stacktrace:

```
java.lang.OutOfMemoryError: Java heap space

	at it.unimi.dsi.fastutil.doubles.DoubleArrays.forceCapacity(DoubleArrays.java:122)
	at it.unimi.dsi.fastutil.doubles.DoubleArrayList.grow(DoubleArrayList.java:244)
	at it.unimi.dsi.fastutil.doubles.DoubleArrayList.add(DoubleArrayList.java:259)
	at tech.tablesaw.api.DoubleColumn.append(DoubleColumn.java:229)
	at tech.tablesaw.api.DoubleColumn.append(DoubleColumn.java:298)
```

The problem is in `DoubleColumn.java`, line 297: 

```java
for (int i = 0; i < numberColumn.size(); i++) {
    append(numberColumn.getDouble(i));
}
```

Using `numberColumn.size()` in the condition for the for loop and then appending to the same column in the for loop will cause the column to grow until the JVM runs out of heap space.

EDIT: It seems like this problem also occurs in IntColumn, LongColumn, DateColumn, etc...

A simple fix would be: 

```java
final int size = numberColumn.size();
for (int i = 0; i < size; i++) {
    append(numberColumn.getDouble(i));
}
```

But it would have to be applied to most if not all classes implementing the `Column` interface. I feel like a more generic solution with less code duplication should be preferred, though.

jln-ho:

## Description

Determine the size of the column about to be appended outside of the for loop that actually appends the two columns. This way, appending a column to itself will not leat to a `java.lang.OutOfMemoryError`. See #390 for more details

## Testing

See `core/src/test/java/tech/tablesaw/columns/ColumnAppendTest.java`

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/BooleanColumn.java
+++ b/core/src/main/java/tech/tablesaw/api/BooleanColumn.java
@@ -590,7 +590,8 @@ public class BooleanColumn extends AbstractColumn<Boolean> implements BooleanMap
     public BooleanColumn append(Column<Boolean> column) {
         checkArgument(column.type() == this.type());
         BooleanColumn col = (BooleanColumn) column;
-        for (int i = 0; i < column.size(); i++) {
+        final int size = col.size();
+        for (int i = 0; i < size; i++) {
             append(col.getByte(i));
         }
         return this;
--- a/core/src/main/java/tech/tablesaw/api/DateColumn.java
+++ b/core/src/main/java/tech/tablesaw/api/DateColumn.java
@@ -244,7 +244,8 @@ public class DateColumn extends AbstractColumn<LocalDate> implements DateFilters
     public DateColumn append(final Column<LocalDate> column) {
         Preconditions.checkArgument(column.type() == this.type());
         DateColumn dateColumn = (DateColumn) column;
-        for (int i = 0; i < dateColumn.size(); i++) {
+        final int size = dateColumn.size();
+        for (int i = 0; i < size; i++) {
             appendInternal(dateColumn.getPackedDate(i));
         }
         return this;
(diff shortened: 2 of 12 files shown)


lwhite1:
@jln-ho Thanks very much. This looks good but it needs to cover Short, Long, and Float columns

jln-ho:
Whoops, no idea how I missed Short, Long and Float. Here they are, anyway.

There seems to be lots of code duplication in the various implementations of `Column`. My latest commit fixes a bug in `ShortColumn.asObjectArray` that was probably just a leftover copy/paste from `IntColumn`. I'm pretty sure there must be some way to reduce this duplication by making the code more generic. Would you be interested in another PR addressing this issue?

lwhite1:
We have noticed the duplication ;)
It can't be made more generic without greatly increasing the amount of
memory used, which decreases the size of dataset that can be analyzed.
If/when Java unifies generics so that primitives can be used the problem
will be addressed.

lwhite1:
@jln-ho Thanks again!

Issue #393, opened 2018-09-22
Dataframe to array?


jrvanalstine:
Is there a method to go from Table to java array? I'm thinking of something similar to pandas dataframe to numpy array.

lwhite1:
Nothing that will do it all in one step for an arbitrary data frame, but
there are some utilities that may help in a class called doublearrays in
utils. And some in io for. interfacing with smile.

I’m away from my computer or I would be more precise

benmccann:
I think you can call `.to().doubleMatrix()`

jrvanalstine:
Ah. I found it. The method is actually Table.as().doubleMatrix(). Thx you two.

benmccann:
Hmm. That should only be required for `.as().smileDataset`

janicebloc:
Is there a convenient way in tablesaw to discretize a number column (i.e. a continuous variable) so that you can chunk it as an ordinal variable?

janicebloc:
I am really struggling with the tablesaw internal implementation for decision tree.  It isn’t working for me.  Does anyone have code that can help to illustrate how it and its parameters are supposed to work?    I could give up and use smile separately, weka, or something like that, but I thought it would be nice to stay within tablesaw.

lwhite1:
The floating point column types implement roundInt(). Not sure if that
meets your needs. If not, you can always implement a floor/ceiling/other
round function by iterating over the column or table.

lwhite1:
Tablesaw doesn't implement any ML wrappers anymore. All you can do is
convert the dataframe to Smile format and use Smile.

PR #394, opened 2018-09-24
Move usage of classes from optional dependency to their own class

benmccann:
Thanks for contributing.

## Description

@jrvanalstine @lwhite1 this fixes https://github.com/jtablesaw/tablesaw/issues/393

I didn't realize when I originally implemented the `TableConverter` that it would require all classes to be present. I thought it would only require them to be present if the methods using them were called, but that's not true

I'm not sure about the API. Right now I have:
```
.as().doubleMatrix()
.smile().dataset()
```

Would something like the below be better?
```
.asMatrix().doubles()
.asSmile().dataset()
```

### DIFF
--- a/core/src/main/java/tech/tablesaw/conversion/SmileConverter.java
+++ b/core/src/main/java/tech/tablesaw/conversion/SmileConverter.java
@@ -1,0 +1,54 @@
+package tech.tablesaw.conversion;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+import smile.data.Attribute;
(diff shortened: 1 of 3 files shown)


jrvanalstine:
Awesome thanks!

jrvanalstine:
I like the first option more. It's clear that anything under .smile() will require the smile dependency.

benmccann:
@lwhite1 thoughts about this API before we merge?

lwhite1:
Sorry for the delay. I've been overly busy lately. 

thanks much.

Issue #395, opened 2018-09-26
It's impossible to parse CSV file with more than 512 columns


Ebalaitung:
I try to parse feature index file into Table (it's about 2000x36000 matrix).
I try to use:
`Table t = Table.read().csv(csv);`
but got error from com.univocity.parsers.csv.CsvParser (log is attached [512-ERROR.txt](https://github.com/jtablesaw/tablesaw/files/2419490/512-ERROR.txt)) that there is more than 512 columns.
I've checked sources and found out that you create CsvParser  with default settings, and there is no option to change some how this settings from your API.

I think there should be some way to change com.univocity.parsers.csv.CsvParserSettings from Table.read()  place, or at least change columns size limit.

benmccann:
I wonder if we should just hardcode `parserSettings.setMaxColumns(Integer.MAX_VALUE)`. I don't really understand why there would be a column limit

Ebalaitung:
> I wonder if we should just hardcode `parserSettings.setMaxColumns(Integer.MAX_VALUE)`. I don't really understand why there would be a column limit

I also thought about it.

lwhite1:
> I wonder if we should just hardcode parserSettings.setMaxColumns(Integer.MAX_VALUE). I don't really understand why there would be a column limit

Seems reasonable. Anyone want to submit a pull request?

Ebalaitung:
> > I wonder if we should just hardcode parserSettings.setMaxColumns(Integer.MAX_VALUE). I don't really understand why there would be a column limit
> 
> Seems reasonable. Anyone want to submit a pull request?

Will do...

Ebalaitung:
Need possibility to configure max amount of columns to parse, as default value is just 512

Ebalaitung:
Added corresponding PR, please review.

PR #403, opened 2018-10-30
fix for #395 It's impossible to parse CSV file with more than 512 col…

Ebalaitung:
Thanks for contributing.

- [X ] Tick to sign-off your agreement to the [Developer Certificate of Origin (DCO) 1.1](https://developercertificate.org)

## Description

Added setting for maximal columns in CVS file in CsvReaderOption

## Testing

Added positive and negative tests for CsvReader for 1000 columns csv file.


core/src/main/java/tech/tablesaw/io/csv/CsvReader.java
```diff
@@ -578,6 +578,9 @@ private ColumnType selectType(List<ColumnType> typeCandidates) {
     private CsvParser csvParser(CsvReadOptions options) {
         CsvParserSettings settings = new CsvParserSettings();
         settings.setFormat(csvFormat(options));
+        if(options.maxNumberOfColumns() != null){
```
benmccann: Can you please add a space after the `if` and before the `{`?

Ebalaitung: fixed

core/src/main/java/tech/tablesaw/io/csv/CsvReadOptions.java
```diff
@@ -264,6 +271,15 @@ public Builder columnTypes(ColumnType[] columnTypes) {
             return this;
         }
 
+        /**
+         * Defines maximal value of columns in csv file.
+         * @param maxNumberOfColumns - must be positive integer. Default is 512.         *
+         */
+        public Builder maxNumberOfColumns(Integer maxNumberOfColumns){
```
benmccann: can you add a space before the `{`

Ebalaitung: fixed

lwhite1:
thanks, Ben, for taking care of this.

benmccann:
I wonder if we should change the default value of this new setting to 10000 so that users will be unlikely to hit the limit. That will use only 40kb of memory (https://github.com/uniVocity/univocity-parsers/issues/286#issuecomment-434929433) and will make Tablesaw easier to use out of the box

Ebalaitung:
@benmccann sounds reasonable, I guess it'll cover 95% use cases. If everybody  are aligned with this approach, I can create PR.

benmccann:
I agree. Unfortunately, univocity-parser doesn't. If you want to change it here I'm supportive of that

PR #467, opened 2019-02-20
Add <meta charset="UTF-8">

yukoba:
I added ```<meta charset="UTF-8">``` to page_template.html for non ascii characters support.

### DIFF
--- a/jsplot/src/main/java/tech/tablesaw/plotly/Plot.java
+++ b/jsplot/src/main/java/tech/tablesaw/plotly/Plot.java
@@ -5,8 +5,11 @@ import tech.tablesaw.plotly.components.Page;
 import tech.tablesaw.plotly.display.Browser;
 
 import java.io.File;
-import java.io.FileWriter;
+import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.OutputStreamWriter;
+import java.io.Writer;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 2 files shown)


jsplot/src/main/java/tech/tablesaw/plotly/Plot.java
```diff
@@ -28,8 +31,8 @@ public static void show(Figure figure, String divName, File outputFile) {
         String output = page.asJavascript();
 
         try {
-            try (FileWriter fileWriter = new FileWriter(outputFile)) {
-                fileWriter.write(output);
+            try (Writer writer = new OutputStreamWriter(new FileOutputStream(outputFile), StandardCharsets.UTF_8)) {
```
benmccann: why are you changing this?

benmccann:
Adding the `meta` tag and `DOCTYPE` tag are small and related changes and would be better to do in a single PR

jsplot/src/main/java/tech/tablesaw/plotly/Plot.java
yukoba: FileWriter uses ```file.encoding``` encoding and if there is no ```file.encoding``` encoding, Java uses UTF-8.
Sometimes ```file.encoding``` is not UTF-8 and I forced to use UTF-8.

yukoba:
Rebased and merged two git branches.

jsplot/src/main/java/tech/tablesaw/plotly/Plot.java
benmccann: Won't this still use UTF-8 encoding since it's passing `StandardCharsets.UTF_8`?

benmccann: Oh, I misunderstood you comment. You want to force it to use UTF-8. That makes sense. Thanks

yukoba: Yes. I want to force it to use UTF-8.

benmccann:
Thanks!

Issue #466, opened 2019-02-20
Simplify support other file formats, e.g. Xslx


hallvard:
I'm looking into supporting Excel files (xslx) by using the Apache POI project. I would like to reuse some of the logic from the corresponding csv code, like CsvReadOptions and its Builder class, but that's pretty difficult. First, the CsvReadOptions has a private constructor (called by its Builder), so you cannot subclass it. Second, some options don't make sense for a format that's not line-oriented, e.g. separator and lineEnding. In addition, the customParser method in ColumnType takes an CsvReadOptions instance as argument, so it needs to be subclassed.

I propose making an AbstractReadOptions class that includes the things that other formats are most likely to need, and that is easy to subclass. And update the customParser method in ColumnType  to use that class as argument instead.

benmccann:
It sounds to me like you're looking at an old version of the code because we recently made some of these changes to support adding the JSON and fixed-width file readers. Can you sync your repo against the latest code from master?

hallvard:
Yes, you're right! It seems like the kind of changes I suggested are in the newest version. My plan is to create a new maven module with the XslxReader and XslxReadOptions in a tech.tablesaw.io.xslx package, so it needn't be part of the core. Then integrate it into the build and create a pull request. Sounds OK?

benmccann:
No need to create a separate issue

I'm not sure the best API for optional dependencies would be or how to best include them. We should probably follow the same pattern for the JSON, HTML, and CSV dependencies as well.

For including, we have the `Table.smile()` optional dependency that makes use of Maven's `optional` attribute. There's also the option to make a separate Maven dependency as you suggested. 

For API, most stuff is done today via `Table.read().csv(...)`, `Table.read().json(...)`, `Table.read().html(...)`, etc. If we have a separate Maven module, it'll have to be a bit different and we'll need to do something like `new XslxReader().read(...)`. I'm curious to learn more about how Spark handles this because I think their API is similar to ours, but they have the ability to register new readers as plugins. I'm not quite sure how that works, but I'd be interested to learn more before we make a decision

benmccann:
I think I'd probably lean towards including the POI dependency as an optional dependency like the Smile dependency and creating the reader in the main module alongside the json, csv, html readers

lwhite1:
I agree

PR #470, opened 2019-02-23
Initial xlsx support, as discussed in issue #466

hallvard:
Thanks for contributing.

## Description
Added **XlsxReader** and **XlsxReadOptions** for the xlsx file format, using Apache's POI libraries.
Added **xlsx** methods in DataFrameReader to support xlsx file format.

## Testing
Test for **XlsxReader** that reads a sample file with several column types.


hallvard:
First of all, all tests passes locally, both when run as junit tests and as mvn test.
I've been improving the assert messages, and this message seems to indicate there is a missing value, but it doesn't make much sense that it doesn't happen locally. The only thing I can think if is a locale issue.

core/pom.xml
```diff
@@ -160,5 +160,19 @@
             <version>1.5.2 </version>
             <scope>test</scope>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/org.apache.poi/poi -->
+		<dependency>
```
benmccann: can you change the tabs to spaces so the indentation aligns with the other dependencies

core/pom.xml
```diff
@@ -160,5 +160,19 @@
...
+		    <groupId>org.apache.poi</groupId>
+		    <artifactId>poi</artifactId>
+		    <version>4.0.1</version>
+		    <optional>true</optional>
+		</dependency>
+		<!-- https://mvnrepository.com/artifact/org.apache.poi/poi-ooxml -->
+		<dependency>
```
benmccann: can you change the tabs to spaces so the indentation aligns with the other dependencies?

core/pom.xml
```diff
@@ -160,5 +160,19 @@
...
             <scope>test</scope>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/org.apache.poi/poi -->
+		<dependency>
+		    <groupId>org.apache.poi</groupId>
+		    <artifactId>poi</artifactId>
+		    <version>4.0.1</version>
+		    <optional>true</optional>
+		</dependency>
+		<!-- https://mvnrepository.com/artifact/org.apache.poi/poi-ooxml -->
```
benmccann: this comment isn't necessary

core/pom.xml
benmccann: I'd suggest removing this dependency since the one below pulls it in. That will make it easier for users because they'll only need to add a single dependency to their project to use this functionality

core/pom.xml
benmccann: same with the other files. I notice they have a mix of tabs and spaces as well

core/src/main/java/tech/tablesaw/io/DataFrameReader.java
```diff
@@ -115,4 +117,14 @@ public Table db(ResultSet resultSet, String tableName) throws SQLException {
     public Table html(String url) throws IOException {
         return new HtmlTableReader().read(url);
     }
+    
+    // xlsx
```
benmccann: I'd remove this comment since it doesn't add anything

core/src/main/java/tech/tablesaw/io/DataFrameReader.java
```diff
@@ -115,4 +117,14 @@ public Table db(ResultSet resultSet, String tableName) throws SQLException {
...
+    
+    public Table[] xlsx(XlsxReadOptions options) throws IOException {
```
benmccann: I wonder if it might be nicer to return a `List<Table>`. I find `List` usually easier to work with than `Array`

core/src/test/java/tech/tablesaw/io/xlsx/XlsxReaderTest.java
```diff
@@ -0,0 +1,95 @@
...
+import java.time.LocalDateTime;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import tech.tablesaw.api.Table;
+import tech.tablesaw.columns.Column;
+
+/**
+ * Tests for XLSX Reading
```
benmccann: you could remove this comment since it doesn't add much

core/src/main/java/tech/tablesaw/io/xlsx/XlsxReader.java
```diff
@@ -0,0 +1,312 @@
...
+            column.appendCell(cell.getRichStringCellValue().getString());
+            return null;
+        case NUMERIC:
+            if (DateUtil.isCellDateFormatted(cell)) {
+                final Date date = cell.getDateCellValue();
+                final LocalDateTime localDate = LocalDateTime.of(date.getYear() + 1900, date.getMonth() + 1, date.getDate(), date.getHours(), date.getMinutes(), date.getSeconds());
+                column.appendCell(localDate.toString());
+                return null;
+            } else {
+                final double num = cell.getNumericCellValue();
```
benmccann: In the other readers we use `ColumnTypeDetector` to figure out the column type upfront and then don't need the logic at append time. I wonder if you could do the same here?

hallvard: I saw **ColumnTypeDetector**, but it doesn't quite fit my use case, since I have objects, not strings. I can reimplement the same strategy, i.e. scanning a number of elements in the columns and selecting the appropriate **ColumnType**. Is it OK that I extend ColumnType with a **supports(Object)** method to help selecting the appropriate type (similar to the parser's **canParse** method)?

benmccann: Might as well just leave it as is for now

I didn't review the PR too much cause it was hard to read with the mixture of tabs and spaces. I'll give it a closer look after that's updated. But it seems pretty good overall

core/src/main/java/tech/tablesaw/columns/Column.java
```diff
@@ -332,11 +332,15 @@ default boolean noneMatch(Predicate<? super T> test) {
...
         for (int i = 0; i < size(); i++) {
-            try {
-                into.set(i, fun.apply(get(i)));
-            } catch (Exception e) {
-                into.setMissing(i);
-            }
+        	if (isMissing(i)) {
+        		into.setMissing(i);
+        	} else {
+        		try {
```
benmccann: spaces and tabs

core/src/main/java/tech/tablesaw/io/xlsx/XlsxReader.java
```diff
@@ -0,0 +1,318 @@
...
+
+    private ColumnType getColumnType(final Cell cell) {
+        switch (cell.getCellType()) {
+        case STRING:
+            return ColumnType.STRING;
+        case NUMERIC:
+            return (DateUtil.isCellDateFormatted(cell) ? ColumnType.LOCAL_DATE_TIME : ColumnType.SHORT);
+        case BOOLEAN:
+            return ColumnType.BOOLEAN;
+        // case BLANK:
```
benmccann: can re remove any commented out code?

core/src/main/java/tech/tablesaw/io/xlsx/XlsxReader.java
```diff
@@ -0,0 +1,318 @@
...
+        }
+        return null;
+    }
+
+    private ColumnType getColumnType(final Cell cell) {
+        switch (cell.getCellType()) {
+        case STRING:
+            return ColumnType.STRING;
+        case NUMERIC:
+            return (DateUtil.isCellDateFormatted(cell) ? ColumnType.LOCAL_DATE_TIME : ColumnType.SHORT);
```
benmccann: parentheses are unnecessary

core/src/main/java/tech/tablesaw/io/xlsx/XlsxReader.java
```diff
@@ -0,0 +1,318 @@
...
+                    }
+                }
+                if (column != null) {
+                    while (column.size() <= rowNum - tableArea[0]) {
+                        column.appendMissing();
+                    }
+                }
+            }
+        }
+        while (columns.remove(null))
```
benmccann: might be cleaner to do `removeAll(Collections.singleton(null));`

core/src/main/java/tech/tablesaw/io/xlsx/XlsxReader.java
```diff
@@ -0,0 +1,318 @@
...
+            ;
+        table.addColumns(columns.toArray(new Column<?>[columns.size()]));
```
benmccann: It might be nicer to add an `addColumns(Collection<Column> columns)` method rather than having to convert to an array here

core/src/main/java/tech/tablesaw/io/DataFrameReader.java
```diff
@@ -115,4 +122,12 @@ public Table db(ResultSet resultSet, String tableName) throws SQLException {
     public Table html(String url) throws IOException {
         return new HtmlTableReader().read(url);
     }
+    
+    public List<Table> xlsx(XlsxReadOptions options) throws IOException {
```
benmccann: Maybe here, or perhaps in the docs, we should document `Requires that "org.apache.poi:poi-ooxml" be on your classpath`

core/src/main/java/tech/tablesaw/io/xlsx/XlsxReadOptions.java
```diff
@@ -0,0 +1,44 @@
...
+import tech.tablesaw.io.ReadOptions;
+
+public class XlsxReadOptions extends ReadOptions {
+
+    protected XlsxReadOptions(final Builder builder) {
+        super(builder);
+    }
+
+    public static Builder builder(File file) {
+        Builder builder = new Builder(file);
```
benmccann: can you write this as one line?

    return new Builder(file).tableName(file.getName());

benmccann: Actually you can just do `return new Builder(file)` because the super constructor already sets the table name based off the file name

core/src/main/java/tech/tablesaw/io/xlsx/XlsxReadOptions.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.io.File;
+import java.io.InputStream;
+import java.io.Reader;
+
+import tech.tablesaw.io.ReadOptions;
+
+public class XlsxReadOptions extends ReadOptions {
+
+    protected XlsxReadOptions(final Builder builder) {
```
benmccann: I would suggest removing `final` everywhere it appears except for when defining class members in order to maintain consistency with the existing code base.

core/src/main/java/tech/tablesaw/io/xlsx/XlsxReadOptions.java
hallvard: OK, but note that it doesn't work to return the value of the chained call to tableName, since its return type is the super-Builder-class, not this subclass. This is a weakness of the design, you need to introduce a type parameter **Builder<T extends Builder>** and use **return (T)** in the super class to allow this. That will make the superclass more complicated, but make the subclasses less.

core/src/main/java/tech/tablesaw/io/DataFrameReader.java
```diff
@@ -123,6 +123,9 @@ public Table html(String url) throws IOException {
         return new HtmlTableReader().read(url);
     }
     
+    // The xlsx methods use the optional dependency org.apache.poi:poi-ooxml.
+    // A modules that use them must itself add this dependency.
```
benmccann: "A modules that use them must itself add this dependency" sounds wrong

Maybe combine the two sentences into "Modules that call this method must add the optional dependency org.apache.poi:poi-ooxml"

Also, I'd suggest making it a Javadoc comment so that it shows up in the Javadoc and add it to both methods:
```
/**
 * Modules that call this method must add the optional dependency org.apache.poi:poi-ooxml
 */ 
```

core/src/main/java/tech/tablesaw/io/xlsx/XlsxReader.java
```diff
@@ -0,0 +1,313 @@
...
+            } else if (lastRowArea == null && rowArea == null) {
+                row1 = -1;
+            } else if (rowArea[0] < lastRowArea[0] || rowArea[1] > lastRowArea[1]) {
+                lastRowArea = null;
+                row2 = -1;
+            } else {
+                row2 = row.getRowNum();
+            }
+        }
+        return (row1 >= 0 && lastRowArea != null ? new int[] { row1, row2, lastRowArea[0], lastRowArea[1] } : null);
```
benmccann: these parens are unnecessary

core/src/main/java/tech/tablesaw/io/xlsx/XlsxReader.java
```diff
@@ -0,0 +1,313 @@
...
+                booleanColumn.append(cell.getBooleanCellValue());
+                return null;
+            }
+        default:
+            break;
+        }
+        return null;
+    }
+
+    public Column<?> createColumn(String name, Cell cell) {
```
benmccann: Can we move this method to the test since it looks like it's only used there? Or else make it `protected` instead of `public`

benmccann:
This looks good! Not sure if @lwhite1 has any thoughts, but I think this looks about ready to merge. Left a few last comments. Also, `XlsxReaderTest` is currently failing on Travis, so that'll need to be fixed first

Thanks for contributing this! And thanks for making all the changes

hallvard:
As mentioned above, the tests pass when run locally, both using Run As -> JUnit tests and mvn test. The (improved) assert message tells that a double isn't read correctly. Could it be a platform issue? I'm on a mac, and the test Excel file was created on a mac. Can one of you try it locally, preferably on Linux?

benmccann:
Ah, right. Forgot about that. I use Linux. I can try later

benmccann:
Hmm, I checked out the PR locally on my machine and the tests passed there as well

core/src/test/java/tech/tablesaw/io/xlsx/XlsxReaderTest.java
```diff
@@ -0,0 +1,93 @@
...
+        int colNum = 0;
+        for (final Column<?> column : table.columns()) {
+            Assert.assertEquals("Wrong column name", columnNames[colNum], column.name());
+            Assert.assertEquals("Wrong size for column " + columnNames[colNum], size, column.size());
+            colNum++;
+        }
+        return table;
+    }
+
+    private <T> void checkColumnValues(final Column<T> column, final T... ts) {
```
benmccann: Can you call this method `assertColumnContains` or something with the word assert in it? That will fix the Codacy warning about not having any asserts in the tests

core/src/main/java/tech/tablesaw/io/xlsx/XlsxReader.java
```diff
@@ -0,0 +1,313 @@
...
+            return DateUtil.isCellDateFormatted(cell) ? ColumnType.LOCAL_DATE_TIME : ColumnType.SHORT;
+        case BOOLEAN:
+            return ColumnType.BOOLEAN;
+        default:
+            break;
+        }
+        return null;
+    }
+
+    private int[] findTableArea(Sheet sheet) {
```
benmccann: Could you add a Javadoc describing what the return value is? Or better yet would be to use a static inner class instead of an array since that would be self-documenting

core/src/main/java/tech/tablesaw/io/xlsx/XlsxReader.java
```diff
@@ -0,0 +1,313 @@
...
+        case NUMERIC:
+            if (DateUtil.isCellDateFormatted(cell)) {
+                Date date = cell.getDateCellValue();
+                LocalDateTime localDate = LocalDateTime.of(date.getYear() + 1900, date.getMonth() + 1,
+                        date.getDate(), date.getHours(), date.getMinutes(), date.getSeconds());
+                column.appendCell(localDate.toString());
+                return null;
+            } else {
+                double num = cell.getNumericCellValue();
+                if (column.type() == ColumnType.SHORT) {
```
benmccann: The other readers won't return a `Column<Short>`. I'd probably remove that type from here for consistency

https://github.com/jtablesaw/tablesaw/blob/f477a81bdee5e6309d59b5bcca19b66834d52ea1/core/src/main/java/tech/tablesaw/io/ColumnTypeDetector.java#L58

https://github.com/jtablesaw/tablesaw/commit/f477a81bdee5e6309d59b5bcca19b66834d52ea1

benmccann:
@hallvard I managed to get the tests passing on this PR, by tweaking our Travis config. It seems that the issue was an old version of Ubuntu and using a more recent version made it work

core/src/main/java/tech/tablesaw/io/xlsx/XlsxReader.java
```diff
@@ -108,13 +107,23 @@ private ColumnType getColumnType(Cell cell) {
         return null;
     }
 
-    private int[] findTableArea(Sheet sheet) {
+    private static class TableRange {
+        int startRow, endRow, startColumn, endColumn;
```
benmccann: these could probably be made `final`

core/src/main/java/tech/tablesaw/io/xlsx/XlsxReader.java
```diff
@@ -0,0 +1,301 @@
...
+                if (cell != null) {
+                    if (column == null) {
+                        column = createColumn(headerNames.get(colNum), cell);
+                        columns.set(colNum, column);
+                        while (column.size() < rowNum - tableArea.startRow) {
+                            column.appendMissing();
+                        }
+                    }
+                    Column<?> altColumn = appendValue(column, cell);
+                    if (altColumn != null && altColumn != column) {
```
lwhite1: Should this be if (altColumn != null && ! altColumn.equals(column))?

hallvard: No. If **appendValue** return a new **Column** instance, that **equals** the old, it should still replace it.

lwhite1:
Thanks @hallvard!  Also, @benmccann for the very thorough review.

benmccann:
Yes, thanks again @hallvard !

core/src/main/java/tech/tablesaw/io/xlsx/XlsxReader.java
```diff
@@ -0,0 +1,301 @@
...
+import org.apache.poi.xssf.usermodel.XSSFWorkbook;
+
+import tech.tablesaw.api.ColumnType;
+import tech.tablesaw.api.DoubleColumn;
+import tech.tablesaw.api.LongColumn;
+import tech.tablesaw.api.Table;
+import tech.tablesaw.columns.Column;
+
+@Immutable
+public class XlsxReader {
```
jbduncan: I believe this class should be made `final` so that users can't subclass it and make instances mutable by adding mutable fields. :)

jbduncan: ... I've only just realised that this PR was already merged in. 🤦‍♂️

PR #475, opened 2019-03-04
Create optional modules

benmccann:
Thanks for contributing.

## Description

This is an alternative proposal to https://github.com/jtablesaw/tablesaw/pull/474. In https://github.com/jtablesaw/tablesaw/pull/474, the user would need to do add the following to use the JSON reader:

```
<dependency>	
     <groupId>com.fasterxml.jackson.core</groupId>	
     <artifactId>jackson-databind</artifactId>	
     <version>2.9.8</version>	
 </dependency>	
 <dependency>	
     <groupId>com.github.wnameless</groupId>	
     <artifactId>json-flattener</artifactId>	
     <version>0.6.0</version>	
 </dependency>
```

In this one they would just need to add a `tablesaw-json` module:
```
<dependency>	
     <groupId>tech.tablesaw</groupId>	
     <artifactId>tablesaw-json</artifactId>	
     <version>0,31.0</version>	
 </dependency>	
 ```

I think this makes it more intuitive and straightforward for users to include new readers.

I did not split out the `CsvReader` in this PR. However, that would be a possibility and the user could still use the library with CSV reading with just a single dependency import. They would just need to use `tablesaw-csv` instead of `tablesaw-core` if they wanted that functionality.


hallvard:
I like this a lot better. With optional dependencies you expose implementation details, since the consumer of our modules need to know our dependencies. We're less likely to change our own module names, than changing our (internal) dependencies, so this proposal is more robust.

hallvard:
As @lwhite1 suggested, I'd keep csv support in core, and create modules for the other readers and for smile.

lwhite1:
I guess this is a bit more work for us, but I agree with @hallvard that this is preferable to the other solution.

benmccann:
I've updated this PR to make an optional `tablesaw-smile` module as well

lwhite1:
Thanks, @benmccann. This is great.

hallvard:
I'm a bit confused, my understanding was that the core was supposed to be small and lean, and then the consumer can add functionality be depending additional modules. However, core now has optional dependencies on the other and the others have core as provided dependency. This means the shaded jar of core is  *huge*, rather than small and lean. I think I understand how it works, but doesn't this prevent you from using a jar with just what's needed for the core to run without its optional dependencies?

benmccann:
`optional` dependency means that the code will work with or without the dependency and it's the user's responsibility to pull it in. `provided` dependency means that the user will provide the dependency. It doesn't automatically get pulled in. This works very well with any package manager like Maven, Gradle, etc.

I didn't realize we have a shaded jar and am not exactly sure what the use case is. Shading is normally used to resolve classpath conflicts, but there shouldn't be any. Since I don't understand what the shaded jar is for, I'm also not sure what the correct functioning is. So I don't know whether we'd want to pull in the optional jars or have them excluded, but I imagine we could configure the shade plugin to do either. Or another option would also be to delete the shaded jar and encourage the use of a package manager

hallvard:
The problem with my understand/expectation was that I thought the change implied "proper" layered dependencies, so the core shouldn't/needn't depend on the optional modules. But since the core has code referring to the readers (e.g. the DataFrameReader) and to smile, this will not work. So although the core will not need the optional jars/classes during runtime, it depends on them at compile time.

Concerning the shaded jar, at least I'm using it for creating OSGi bundles for the Eclipse integration. Currently, the shaded jar pulls in everything, and I'd like to tweak the shade settings so it gives a jar without the optional stuff in the core project, and jars with whatever is needed in the specific optional module projects. If it's OK with you I'll experiment with it in another branch and issue a PR if I can make it work.

lwhite1:
@hallvard Feel free to do what you like with the shade plugin

hallvard:
It seems a bit awkward that the new modules don't include the tests. If you work on a module, you'll need to update the tests somewhere else.

benmccann:
Yeah, I was debating moving them as well. The trade-off is testing e.g. the `HtmlReader` directly vs using the `Table.read().html()` method. I'd be okay to do it either way

hallvard:
We could of course split the test in two, the main part residing in the Xyz module and using only the XyzReader API, and one in the core module testing the DataFrameReader API. The latter shouldn't evolve that much, so if we don't want to split the tests, I suggest moving them into the modules.

benmccann:
Sounds good to me

PR #478, opened 2019-03-05
Features/473 row setters

owlz84:
Thanks for contributing.

## Description

- Added setters to Row class for:
  - Ints, Shorts and Longs
  - Floats and Doubles
  - Dates, Times and DateTimes
  - Strings and Text
  - Booleans
- Omitted setters for PackedDate, PackedTime and PackedDateTime
- Omitted setter for Object columns

## Testing

Full set of (very basic) tests for getters and setters in Row. Selected appropriate source data and data types for testing each method.


core/src/test/java/tech/tablesaw/api/RowTest.java
```diff
@@ -6,8 +6,11 @@
 import java.io.File;
 import java.io.IOException;
 import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
 
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertEquals;
+import static tech.tablesaw.api.ColumnType.*;
```
lwhite1: Please remove the unused import. Codacy will thank you.

core/src/main/java/tech/tablesaw/api/Row.java
```diff
@@ -82,64 +84,110 @@ public boolean hasNext() {
...
 
-    public float getFloat(String columnName) {
-	return floatColumnMap.get(columnName).getFloat(rowNumber);
+    public Object getObject(int columnIndex) {
+        return columnMap.get(columnNames[columnIndex]).get(rowNumber);
     }
 
-    public float getFloat(int columnIndex) {
-        return getFloat(columnNames[columnIndex]);
+    public int getPackedDate(String columnName) {
```
benmccann: I would probably not include getters for the packed date / datetime as I think it's a bit of an implementation detail

lwhite1: I think i've used them for performance sensitive work, but i agree they'd almost never be needed in normal use. i'm ok with their being removed.

core/src/test/java/tech/tablesaw/api/RowTest.java
owlz84: Removed in latest commit.

core/src/main/java/tech/tablesaw/api/Row.java
lwhite1: I'll add an issue for this and do it in a subsequent change

PR #480, opened 2019-03-07
Reduce code duplication

benmccann:
Thanks for contributing.

## Description

Added additional `builder` methods to speed `Layout.Builder` creation

### DIFF
--- a/jsplot/src/main/java/tech/tablesaw/plotly/api/AreaPlot.java
+++ b/jsplot/src/main/java/tech/tablesaw/plotly/api/AreaPlot.java
@@ -1,7 +1,6 @@
 package tech.tablesaw.plotly.api;
 
 import tech.tablesaw.api.Table;
-import tech.tablesaw.plotly.components.Axis;
 import tech.tablesaw.plotly.components.Figure;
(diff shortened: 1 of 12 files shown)


jsplot/src/main/java/tech/tablesaw/plotly/api/AreaPlot.java
```diff
@@ -35,24 +30,8 @@ public static Figure create(String title, Table table, String xCol, String yCol,
...
-        return Layout.builder()
-                    .title(title)
-                    .height(HEIGHT)
-                    .width(WIDTH)
-                    .showLegend(true)
-                    .xAxis(Axis.builder()
-                            .title(xCol)
-                            .build())
-                    .yAxis(Axis.builder()
-                            .title(yCol)
```
lwhite1: is it me or do you lose the default use of the column titles as the x and y axis titles with the changes you made?

benmccann: I don't think so, do we? The new version passes `xCol` and `yCol` into the `builder` method, which sets them

PR #481, opened 2019-03-07
Updated the shade configuration, so it creates jars that can be used for

hallvard:
OSGi bundles and Eclipse integration.

Thanks for contributing.

## Description

Shade configurations were updated/created in each module's pom.xml files.

## Testing

I checked that the shaded jars were usable in the Eclipse integration.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -92,11 +92,6 @@
         </dependency>
         <dependency>
             <groupId>org.apache.commons</groupId>
-            <artifactId>commons-lang3</artifactId>
-            <version>3.8.1</version>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.commons</groupId>
             <artifactId>commons-math3</artifactId>
             <version>3.6.1</version>
         </dependency>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -69,30 +69,6 @@
                 <configuration>
                     <shadedArtifactAttached>true</shadedArtifactAttached>
                     <shadedClassifierName>shaded</shadedClassifierName>
-                    <artifactSet>
-                        <excludes>
-                            <exclude>tech.tablesaw:tablesaw-excel</exclude>
-                            <exclude>org.apache.poi:*</exclude>
-                            <exclude>commons-codec:commons-codec</exclude>
-                            <exclude>org.apache.commons:commons-collections4</exclude>
-                            <exclude>org.apache.xmlbeans:xmlbeans</exclude>
-                            <exclude>org.apache.commons:commons-compress</exclude>
-                            <exclude>com.github.virtuald:curvesapi</exclude>
-
-                            <exclude>tech.tablesaw:tablesaw-html</exclude>
-                            <exclude>org.jsoup:jsoup</exclude>
-                            
-                            <exclude>tech.tablesaw:tablesaw-json</exclude>
-                            <exclude>com.fasterxml.jackson.core:*</exclude>
-                            <exclude>com.github.wnameless:json-flattener</exclude>
-                            <exclude>com.eclipsesource.minimal-json:minimal-json</exclude>
-
-                            <exclude>tech.tablesaw:tablesaw-smile</exclude>
-                            <exclude>com.github.haifengl:smile-data</exclude>
-                            <exclude>org.apache.commons:commons-text</exclude>
-                            <exclude>com.github.haifengl:smile-math</exclude>
-                        </excludes>
-                    </artifactSet>
                 </configuration>
                 <executions>
                     <execution>
@@ -113,6 +89,11 @@
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
             <version>25.1-jre</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>3.8.1</version>
         </dependency>
         <dependency>
             <groupId>org.apache.commons</groupId>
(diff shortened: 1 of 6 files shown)


excel/pom.xml
```diff
@@ -49,6 +49,36 @@
...
+      
+       <plugin>
+           <groupId>org.apache.maven.plugins</groupId>
+           <artifactId>maven-shade-plugin</artifactId>
+           <version>3.1.0</version>
+           <configuration>
+               <shadedArtifactAttached>true</shadedArtifactAttached>
+               <shadedClassifierName>shaded</shadedClassifierName>
+               <artifactSet>
+                   <includes>
```
benmccann: Is it necessary to have this? I thought `provided` dependencies would be excluded automatically. Or is there something else being done here?

benmccann:
Listing every transitive dependency seems extremely brittle and annoying to maintain. If the shade plugin can't be more easily configured, perhaps [the bundle plugin](http://felix.apache.org/documentation/subprojects/apache-felix-maven-bundle-plugin-bnd.html) could be used instead? I think it's more directly meant for building OSGI packages and allows you to include or exclude dependencies by scope

hallvard:
I completely agree, but this is what I got working in my first attempt. I looked at the shaded jar for jsplot and it included everything, but I may have looked at an old one from before the modules were configured with provided. I'll try a bit more... I know of the bundle plugin, but I think it requires a bit more elaborate configuration, but I'll dig more into it. I really want to robust and easy to maintain setup, that doesn't interfere with the rest of the build setup for the non-OSGi case.

hallvard:
The core needs a list of dependencies to exclude, since shade doesn't seem to support omitting optional dependencies. It works with provided dependencies, so I can remove the include list for html, json and html. jsplot is different, since it just depends on core the normal way. There's a suggestion here: https://stackoverflow.com/questions/28458058/maven-shade-plugin-exclude-a-dependency-and-all-its-transitive-dependencies, but I haven't tried it, yet.

core/pom.xml
```diff
@@ -69,6 +69,30 @@
                 <configuration>
                     <shadedArtifactAttached>true</shadedArtifactAttached>
                     <shadedClassifierName>shaded</shadedClassifierName>
+                    <artifactSet>
+                        <excludes>
```
benmccann: What if we make the dependencies on the reader projects `provided` and `optional`?

I'd just include a comment explaining why we did it so that it's not accidentally removed late:

    <scope>provided</scope>

benmccann:
Why do you need separate jars for Eclipse? Won't you want to include all the different readers? Why not just make one uber jar that includes everything including core and the readers

hallvard:
Requiring the user to install all the features (in this case, readers), including the 19Mb xlsx reader, goes against the idea of modular and extensible frameworks, that is key to Eclipse's architecture. In the Eclipse integration of tablesaw the optional readers are provided by optional plug-ins (bundles) that will only be activated if the corresponding file types are used and opened. This is done using Eclipse's extension mechanism. I know classes in general aren't loaded before they're actively used, but jar size is still an issue. Tablesaw could use a similar trick with a registry of readers and a Table.read method that takes the file type as its only argument. This will also make it possible to avoid the "reversed" optional dependency.

benmccann:
Wow. I didn't realize `poi-ooxml` and dependencies are 19Mb. I would have thought it'd be much smaller

I like the idea of using a registry. I think the current API is nicer than having to pass the file extension. But I'm not sure you'd have to change the API to implement a registry.

hallvard:
The poi stuff is generic for the whole set of office file formats. There are smaller libraries that support just xls, but I'm not sure they support xlsx.

The problem with an API with specific methods for each file format is the dependency from the core to the file format module. There's lots of formats out there, e.g. Smile supports several that could be of interest to tablesaw (without using Smile), and each would need another method in the DataFrameReader and another dependency from core to be "officially" supported. With a file format registry (and some way of discovering the modules providing them) this wouldn't be a problem. Yes, you can implement a registry without changing the API, but that wouldn't give you the advantage of increased modularity.

benmccann:
I think we could do both. Have a registry which supports new formats. But have an easy fluent API for the most common readers which we support directly which calls the registry under the hood instead of calling the specific reader instance directly

core/pom.xml
```diff
@@ -114,6 +90,11 @@
             <artifactId>guava</artifactId>
             <version>25.1-jre</version>
         </dependency>
+        <dependency>
```
benmccann: Why do we need to add a new dependency for this?

core/pom.xml
```diff
@@ -90,6 +90,11 @@
...
```
benmccann: Why do we need to add a new dependency for this?

hallvard: It's the commons-lang3 module that has been added, since it's used by the cvs reader. OSGi runs independent modules (bundles) in isolation, and the core OSGi bundle doesn't depend on any other bundle. Without this dependency the core bundle missed a class (I don't remember which) from commons-lang3. Presumably it's also required by some other (optional) module, so you won't notice when run in a normal runtime.

When we make the modules more independent, each must list its own dependencies to make sure that each bundle's dependencies are satisfied (considering that all modules depend on the core, but not the other way around).

benmccann: Good catch. I just pushed a change to `master` so that we no longer depend on `commons-lang3` anymore in an effort to keep the core as small as possible, so you should be able to remove this now

smile/pom.xml
```diff
@@ -49,6 +49,37 @@
...
+               <shadedArtifactAttached>true</shadedArtifactAttached>
+               <shadedClassifierName>shaded</shadedClassifierName>
+               <artifactSet>
+                   <includes>
+                        <include>tech.tablesaw:tablesaw-smile</include>
+                        <include>com.github.haifengl:smile-data</include>
+                        <include>com.github.haifengl:smile-math</include>
+                        <include>org.apache.commons:commons-text</include>
+                        <!-- included in core 
+                        <include>org.apache.commons:commons-lang3</include>
```
benmccann: this one's not anymore

hallvard: Good catch!

hallvard:
Should actually be able to do without the **includes** section, as for the other optional modules that has a provided dependency on core.

benmccann:
That'd be great. I think this would look a lot better if we were able to do without the `includes`

smile/pom.xml
```diff
@@ -50,25 +50,13 @@
         <artifactId>maven-compiler-plugin</artifactId>
       </plugin>
       
-      <plugin>
+       <plugin>
```
benmccann: added an extra space

excel/pom.xml
```diff
@@ -49,6 +49,25 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
       </plugin>
+      
+       <plugin>
+           <groupId>org.apache.maven.plugins</groupId>
+           <artifactId>maven-shade-plugin</artifactId>
+           <version>3.1.0</version>
```
benmccann: Can we move all the common configuration under `pluginManagement` in the parent pom?

PR #482, opened 2019-03-07
Issue 479 template customization

hallvard:
property.

Thanks for contributing.

## Description

Modified the template to support setting location of plotly js library, and Page and PageBuilder accordingly.

## Testing

There are tests for both cases, the default and custom cases.

### DIFF
--- a/jsplot/src/main/java/tech/tablesaw/plotly/components/Page.java
+++ b/jsplot/src/main/java/tech/tablesaw/plotly/components/Page.java
@@ -1,13 +1,13 @@
 package tech.tablesaw.plotly.components;
 
-import com.mitchellbosecke.pebble.error.PebbleException;
-import com.mitchellbosecke.pebble.template.PebbleTemplate;
-
 import java.io.IOException;
 import java.io.StringWriter;
 import java.io.Writer;
 import java.util.HashMap;
 import java.util.Map;
+
+import com.mitchellbosecke.pebble.error.PebbleException;
+import com.mitchellbosecke.pebble.template.PebbleTemplate;
 
 /**
  * Represents an entire html page that contains a figure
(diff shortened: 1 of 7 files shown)


hallvard:
The test succeeds locally when run using mvn clean test. jsplot defines the following dependency on junit:
```
        <dependency>
            <groupId>junit</groupId>
            <artifactId>junit</artifactId>
            <scope>test</scope>
        </dependency>
```

benmccann:
Not anymore: https://github.com/jtablesaw/tablesaw/blob/master/jsplot/pom.xml#L65

We just upgraded to junit 5. Can you rebase against master?

hallvard:
Hm, for some reason I don't get the changes, I'm told my upstream/master branch pointing to https://github.com/jtablesaw/tablesaw.git is up-to-date and the last commit I see is "Use static imports for junit asserts"

hallvard:
Strange, when I turned off VPN I suddenly could fetch the latest changes.

hallvard:
I use Assertions.assertTrue, isn't that good enough?

benmccann:
Perhaps it doesn't have great support for junit 5 yet. Everywhere else we statically import `assertTrue` and I'm guessing that makes it happier. Can you try that?

jsplot/src/main/java/tech/tablesaw/plotly/components/Page.java
```diff
@@ -17,11 +17,18 @@
...
 
+    private String plotlyJsLocation = null;
+
     private Page(PageBuilder builder) {
         this.figure = builder.figure;
         this.divName = builder.divName;
+        this.plotlyJsLocation = builder.plotlyJsLocation;
     }
 
+    public void setPlotlyJsLocation(String plotlyJsLocation) {
```
benmccann: Could we remove this since it's unused and make `plotlyJsLocation` final?

hallvard: Of course, with a builder it is nonsense.

hallvard:
I thought about several designs, one involved a global (singleton) TemplateManager that configured the engine and loader and that had the setTemplateLocations method. But I couldn't come up with other reasons for exposing the engine and loader than this method, so I decided on just introducing the method as a start. We can still keep the method if we need to introduce a (replaceable) singleton later. I haven't tried the method it in Eclipse, yet, but will do it now.

hallvard:
I seemed to work as expected, I could use a different template located in the Eclipse workbench instead of the jsplot jar!

PR #487, opened 2019-03-09
Introduce HtmlReadOptions and JsonReadOptions

benmccann:
Thanks for contributing.

## Description

Introduced `HtmlReadOptions` and `JsonReadOptions` allowing for greater configuration. E.g. we can now read html from a file instead of just from the web. This will also make it easier to build a registry for different reader types

## Testing

Mostly covered by existing unit tests. Should add some tests for the html reader though. That wasn't possible before because it couldn't read from a local file, but now we can. However, I'll wait to do that in another PR because I'd also like to move the tests to live in the subprojects as suggested by hallvard

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -111,6 +111,12 @@
             <version>2.7.5</version>
         </dependency>
         <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <version>1.18.6</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
             <groupId>tech.tablesaw</groupId>
             <artifactId>tablesaw-excel</artifactId>
             <version>0.32.1-SNAPSHOT</version>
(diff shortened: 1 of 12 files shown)


lwhite1:
Sorry @benmccann, I made a mess of this by merging the simple request (Auto-detect CSV line endings) first, which then had merge conflicts with this request. 

In attempting to fix that, I belatedly realized that this commit re-introduced Lombok. I removed Lombok some time ago because the Idea Lombok plugin was flakey. It made the generated code sometimes invisible to the editor, breaking autocomplete, and causing warnings to appear wherever the generated methods were used. I'm not sure I want to go down that path again.

I have backed this commit out. Can it be resubmitted without Lombok or is that essential here for some reason?

benmccann:
Ok, I'll try again without Lombok. The Eclipse plugin is really good, so I didn't realize there were issues with the Intellij version

PR #497, opened 2019-03-20
Fixes toString exception when columns are not same size.

hallvard:
Adds method for padding with missing values.

Thanks for contributing.

## Description

Changed DataFramePrinter so it and Table.toString() does crash with varying sized columns.
Added method in Table for padding with missing values.

## Testing

Tests toString method and the padding method.


core/src/test/java/tech/tablesaw/api/TableTest.java
```diff
@@ -547,4 +550,56 @@ private DoubleColumn sum(DoubleColumn... columns) {
...
+
+    @Test
+    public void testToStringColumnsWithVaryingSizes() {
+        IntColumn col11 = IntColumn.create("col1");
+        IntColumn col12 = IntColumn.create("col2");
+        Table t1 = Table.create("t1", col11, col12);
+        col11.append(1).append(2);
+        col12.append(1);
+        try {
+            t1.toString();
```
benmccann: Can you assert that the result of `toString` isn't `null`? Use a static import to make Codacy happy

core/src/test/java/tech/tablesaw/api/TableTest.java
```diff
@@ -547,4 +550,56 @@ private DoubleColumn sum(DoubleColumn... columns) {
...
+        } catch (Exception e) {
+            Assertions.fail("toString shouldn't throw " + e);
```
benmccann: I don't think you need to catch the exception. I believe the test will fail automatically if an exception is thrown

hallvard: Yes, but by including the fail I hoped to make Codacy happy. And it makes it a bit clearer what I the test was about, and it would be reported as a failure, not an error.

core/src/main/java/tech/tablesaw/io/string/DataFramePrinter.java
```diff
@@ -189,22 +190,25 @@ private String tableName(Relation frame, int width) {
             int i;
             for (i = 0; i < Math.ceil((double) rowCount / 2); i++) {
                 for (int j = 0; j < colCount; j++) {
-                    data[i][j] = frame.getString(i, j);
+                    Column<?> col = frame.column(j);
+                    data[i][j] = (col.size() > i ? col.getString(i) : "?");
```
benmccann: I'd probably use "X" or "" rather than "?"

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -203,6 +203,35 @@ public DataFrameWriter write() {
...
+    private void padColumnWithMissing(final Column<?> col, int newSize) {
+        while (col.size() < newSize) {
+            col.appendMissing();
+        }
+    }
+
+    /**
+     * Appends missing values to columns in the table and/or the Column argument,
+     * so their sizes end up being the same.
+     * @param cols
```
benmccann: either fill out or remove

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -203,6 +203,35 @@ public DataFrameWriter write() {
...
+     * @return this Table
+     */
+    public Table padColumnsWithMissing(final Column<?>... cols) {
```
benmccann: This seems like a strange method to me. It pads the given columns and columns in the table? Why? Presumably to enable adding those columns to the table, so why not do that here?

core/src/main/java/tech/tablesaw/io/string/DataFramePrinter.java
lwhite1: Can we have it print whatever it prints when the column is the correct size but contains a missing  value?  I believe calling getUnformattedString() handles that.

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -203,6 +203,35 @@ public DataFrameWriter write() {
         return new DataFrameWriter(defaultWriterRegistry, this);
     }
 
+    private void padColumnWithMissing(final Column<?> col, int newSize) {
```
lwhite1: Should this be a public column method?

core/src/main/java/tech/tablesaw/api/Table.java
lwhite1: I think there's also an open question about what should happen if the column padding makes the columns differ in size from the other columns in the table. Should that cause an exception to be thrown?

In general, it should not be possible to create a non-rectangular table, and adding columns that are the wrong size should cause an exception to be thrown. I think that means the only way to then make a non-rectangular table is to call append() on some but not all columns. 

To avoid problems, I might remove the columns argument from this method so the result was always a legal table. Again (though i haven't explicitly tried this) I would think that the safest way to add data to a table is to pad all the columns with missing and set the values in the new row(s). 

For me using column append() and forgetting to call appendMissing() on a column where there's no value to add is a source of frequent bugs.

benmccann: I think the tricky case is what if you add all equally sized columns to a table and then append to one columns and not the others. How would we catch that? Otherwise, I agree

core/src/main/java/tech/tablesaw/io/string/DataFramePrinter.java
benmccann: I think you'd want to differentiate between empty and non-existent though. I think I'd rather focus on how to avoid getting the table into this state rather than fixing `toString` though

core/src/test/java/tech/tablesaw/api/TableTest.java
benmccann: The `assertNotNull` will do both of those things

core/src/main/java/tech/tablesaw/api/Table.java
benmccann: I think it if we wanted to make it public, we should put it on the `Column` in that case. I'm fine with it either way

core/src/main/java/tech/tablesaw/api/Table.java
lwhite1: AFAICT, there is no reasonable way to prevent that, what I'm suggesting is "Don't do that". 
Do this instead:

```
t.padColumnsWithMissing(); // no params, so all columns are still the same size
t.stringColumn("foo").set(t.rowCount()-1, "my new value"); // no append so all columns are the same size

```

core/src/main/java/tech/tablesaw/api/Table.java
lwhite1: I think a method that modifies a single column and does nothing else belongs in column.

core/src/main/java/tech/tablesaw/io/string/DataFramePrinter.java
lwhite1: > I think you'd want to differentiate between empty and non-existent though. 

I don't think there is a practical difference between empty and non-existent in this context. An empty string is just a convenient representation of "no value". It's the only datatype we suport that can even represent an empty value that's not a null value, I think consistency argues for dealing only in present or not-present (i.e. missing).

>  I think I'd rather focus on how to avoid getting the table into this state rather than fixing toString though

As @hallvard pointed out, since toString() is useful for debugging, we should try to ensure it doesn't blow up.

hallvard: I think we agree that a missing value (present, but using the missing value marker) is perfectly ok, while a too short column is a bug. That's why the output should be different, to help discover that something is wrong. It also seems we agree that toString shouldn't crash, so what remains is agreeing on what to show to indicate the missing-because-the-column-is-too-short situation. I prefer "?" since it's less likely to be a real value, but "x" is also fine. Perhaps we even can use 💣(bomb emoji unicode character)?

core/src/main/java/tech/tablesaw/api/Table.java
hallvard: @lwhite1 The method pads columns in the table too, so afterwards all the columns, both the ones already on the table and those given as arguments will be the same size. It makes sense to call this method without any arguments, to fix the table itself, and with additional arguments, to fix columns before adding them.
@benmccann Since the API is fluid, you'd do padColumnsWithMissing(cols).append(cols). We could also have both a non-args padColumnsWithMissing() to fix the table and a padWithMissingAndAppend(cols) to pad and append afterwards.

core/src/main/java/tech/tablesaw/api/Table.java
hallvard: If it's private and only used here it should be here. If it's made public I agree it should be moved to Column. We can also just remove it and copy the code into the public method.

core/src/main/java/tech/tablesaw/io/string/DataFramePrinter.java
lwhite1: Got it. Bomb would be nice, but lets go with "?"

core/src/main/java/tech/tablesaw/api/Table.java
lwhite1: > @lwhite1 The method pads columns in the table too, so afterwards all the columns, both the ones already on the table and those given as arguments will be the same size.

Nah, the table's columns are only the same size if you do it right. If you pad all the columns the result is always legal (assuming it was legal before you started).  If you specify the columns you can turn a legal table into an illegal one.

core/src/main/java/tech/tablesaw/api/Table.java
lwhite1: This method operates on one column and doesn't have table as an argument or return value. It is also potentially useful outside of a table. It needs to be on column.

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -203,6 +203,49 @@ public DataFrameWriter write() {
...
+            }
+        }
+        if (addColumns) {
+            addColumns(cols);
+        }
+        return this;
+    }
+
+    /**
+     * Appends missing values to columns in this table,
```
benmccann: don't need the comment. the two lines could probably fit on a single line

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -203,6 +203,49 @@ public DataFrameWriter write() {
...
+    }
+
+    /**
+     * Appends missing values to columns in this table and to the columns given as arguments,
+     * before adding the columns. This ensures the table is rectangular.
+     * @param cols the columns to pad and add 
+     * @return this Table
+     */
+    public Table padWithMissingAndAddColumns(final Column<?>... cols) {
+        return padColumnsWithMissing(true, cols);
```
benmccann: I think you could simplify `padColumnsWithMissing` by making it just act on the table and take no args if you add the columns first before calling like:

```
    addColumns(cols);
    return padColumnsWithMissing();
```

benmccann: Actually, I think I'd remove this method. It's not much harder to call `addColumns(cols).padColumnsWithMissing();` vs `padWithMissingAndAddColumns(cols)`

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -203,6 +203,49 @@ public DataFrameWriter write() {
...
+        }
+        return this;
+    }
+
+    /**
+     * Appends missing values to columns in this table,
+     * to ensure the table is rectangular.
+     * @return this Table
+     */
+    public Table padColumnsWithMissing() {
```
benmccann: Can we come up with a shorter name?

Maybe `rectangularize`? That's what R calls it: https://www.rdocumentation.org/packages/SciencesPo/versions/1.3.7/topics/rectangularize

core/src/main/java/tech/tablesaw/api/Table.java
hallvard: But isn't the point that you want to pad _before_ adding, since addColumns throw an Exception if the column are not properly sized? The non-args padColumnsWithMissing() is for the case where the column sizes have been changed after they've been added.

benmccann: Ah, good point. I might rename to `addColumnsAndRectangularize` so that if you're auto-completing or looking through the javadocs it comes up right next to `addColumns`. That name still seems long to me though...  Maybe `addColumns(boolean rectangularize, Column<?>... cols)`?

lwhite1: @hallvard I don't think that's really the point. 

The exception is good because adding a column of a different size is almost always a logic error. Most new columns are derived from existing ones, so you rarely intentionally add a column that aligns perfectly, except that it doesn't have a value for some contiguous block of rows at the end of the table.  In other words, appending missing is rarely the fix you want.

On the other hand, forgetting to add values to a column inside of a table _is_ a frequent programming mistake where appending a missing value is often the right solution. So that's the (second) problem we're trying to solve.

The first problem was to prevent toString() from blowing up. I think we're trying to do too many things here. Maybe we should just merge the toString fix and leave all the other stuff separate until we agree.

lwhite1: @benmccann 
> Actually, I think I'd remove this method. It's not much harder to call addColumns(cols).padColumnsWithMissing() vs padWithMissingAndAddColumns(cols) (or addColumns(cols).rectangularize() if we rename the method)

I don't really like rectangularize, because I think we're trying to prevent tables from becoming non-rectangular to start with. 

This is what I've been asking for from the beginning. 

On the very rare occasion that you have n different columns of n different lengths and want them all in a single table of one length, you can write a single while loop.

lwhite1:
Hey @hallvard, are you planning to come back to this? 
I feel like I may have inadvertently insulted you or something. If so, I'm very sorry. It was not my intent.

hallvard:
> Hey @hallvard, are you planning to come back to this?
> I feel like I may have inadvertently insulted you or something. If so, I'm very sorry. It was not my intent.

Nothing like that, it's just that I've been away this week, so haven't had time to work in this. I will probably look at it this weekend. My understanding is that you prefer a method for fixing existing columns, not new ones (about to be added). But it's not high priority and the name is still undecided. In this situation I agree it's best to focus on the point of this issue: avoiding that toString() blows up.

lwhite1:
> > Hey @hallvard, are you planning to come back to this?
> > I feel like I may have inadvertently insulted you or something. If so, I'm very sorry. It was not my intent.
> 
> Nothing like that, it's just that I've been away this week, so haven't had time to work in this. I will probably look at it this weekend. My understanding is that you prefer a method for fixing existing columns, not new ones (about to be added). But it's not high priority and the name is still undecided. In this situation I agree it's best to focus on the point of this issue: avoiding that toString() blows up.

That sounds good. There's no rush, but I'd have felt bad if you didn't think your contributions were appreciated. Besides @benmccann and I, I think you're our most regular contributor and it's been truly great having you on board.

Speaking of @benmccann, the only problem with him is that he's a much better programmer than I am and that can get kinda annoying. ;)  Overall all though I think it's kind of amazing how many people have contributed recently. I can barely keep up sometimes. So, thanks again for everything.

benmccann:
@lwhite1 that's very nice of you to say, but hard for me to believe since you've written most of the code :-) 
 
@hallvard I'll echo @lwhite1's sentiments and say it's been great having you contribute! Thank you so much!

lwhite1:
> @lwhite1 that's very nice of you to say, but hard for me to believe since you've written most of the code :-)
That's true, but the quality would suck if you weren't involved.

core/src/main/java/tech/tablesaw/io/string/DataFramePrinter.java
```diff
@@ -171,7 +174,7 @@ private String tableName(Relation frame, int width) {
         });
         return header;
     }
-
+
```
benmccann: extra whitespace

core/src/main/java/tech/tablesaw/io/string/DataFramePrinter.java
```diff
@@ -189,22 +192,25 @@ private String tableName(Relation frame, int width) {
...
+                    data[i][j] = (col.size() > frame.rowCount() - maxRows + i - 1 ? col.getString(frame.rowCount() - maxRows + i - 1) : TOO_SHORT_COLUMN_MARKER);
                 }
             }
         } else {
             data = new String[rowCount][colCount];
             for (int i = 0; i < rowCount; i++) {
                 for (int j = 0; j < colCount; j++) {
-                    String value = frame.getString(i, j);
+                    Column<?> col = frame.column(j);
+                    String value = (col.size() > i ? col.getString(i) : TOO_SHORT_COLUMN_MARKER);
```
benmccann: It might be nice to make a method that refactors out the common logic:

```
private String getString(Column<?> col, index i) {
  return col.size() > i ? col.getString(i) : TOO_SHORT_COLUMN_MARKER;
}
```

core/src/main/java/tech/tablesaw/io/string/DataFramePrinter.java
```diff
@@ -189,22 +192,25 @@ private String tableName(Relation frame, int width) {
...
                 }
             }
             for (int j = 0; j < colCount; j++) {
                 data[i][j] = "...";
             }
             for (++i; i <= rowCount; i++) {
                 for (int j = 0; j < colCount; j++) {
-                    data[i][j] = frame.getString(frame.rowCount() - maxRows + i - 1, j);
+                    Column<?> col = frame.column(j);
+                    data[i][j] = (col.size() > frame.rowCount() - maxRows + i - 1 ? col.getString(frame.rowCount() - maxRows + i - 1) : TOO_SHORT_COLUMN_MARKER);
```
benmccann: I'd probably refactor out `maxRows + i - 1` as a common variable. Though that will happen automatically if you make a common method as suggested below

core/src/main/java/tech/tablesaw/io/string/DataFramePrinter.java
```diff
@@ -29,6 +30,8 @@
  */
 public class DataFramePrinter {
 
+    private final static String TOO_SHORT_COLUMN_MARKER = "?";
```
benmccann: Maybe use the Unicode null character? https://www.fileformat.info/info/unicode/char/0000/index.htm

I'm wondering what if someone has a column containing `?`. That might help distinguish

hallvard: I suggested the bomb character, since it stands out and is an unlikely column value. The crucial question is what does the character look like in the console (view)?
BTW, we could also try/catch runtime exceptions and just toString the exception object.

PR #500, opened 2019-03-23
Add a callback for creating HTML elements

benmccann:
Thanks for contributing.

## Description

Added the ability for user to override html element creation functionality

## Testing

Moved existing unit test to `HtmlWriterTest` and created a new unit test there for the new functionality


lwhite1:
LGTM

benmccann:
Yeah, I would probably be fine to remove the splitting of column names. I don't remember why that was added initially. Maybe there's a test that demonstrates it? (might remind me why it was added). Even if we wanted to keep it, that still seems like a bug if it's splitting "DNAx" to "DN Ax", so it could probably be improved at the very least. But anyway, I don't have any objection to removing it

PR #513, opened 2019-04-30
Remove unused epoch conversion methods

benmccann:
Thanks for contributing.

## Description

I'd like to remove these methods because they are unused and also because they rely on `ZoneId.systemDefault()` which will introduce problems by returning different results on different systems

## Testing

N/A


lwhite1:
LGTM.

Issue #519, opened 2019-05-02
Reading incorrect column type


edgar131:
I'm using tablesaw to get tabular data from a postgres database.  Currently getting the following 

> error: org.postgresql.util.PSQLException: Bad value for type short : -744443
>        at org.postgresql.jdbc.PgResultSet.toShort(PgResultSet.java:2807)
>	at org.postgresql.jdbc.PgResultSet.getShort(PgResultSet.java:2047)
>	at com.zaxxer.hikari.pool.HikariProxyResultSet.getShort(HikariProxyResultSet.java)
>	at tech.tablesaw.io.jdbc.SqlResultSetReader.read(SqlResultSetReader.java:147)
>	at tech.tablesaw.io.DataFrameReader.db(DataFrameReader.java:111)

Which stems from me doing a simple Table read on a resultSet:

`return Table.read().db(preparedStatement.executeQuery());`.

For some reason it seems to be misinterpreting certain numeric columns as shorts even though they have data larger than that.

Update:
The actual issue is that by default NUMERIC types in Postgres do not have a precision or scale.  This results in them being treated as a short column since the code checks for the precision being <=4 and assigning it that way.  This then leads to an error when translating the data since it tries to retrieve the values of that column as shorts even though the data may have values that don't fit as a short type.

A workaround is to ensure that the database columns you query have an explicit precision specified ie. "NUMERIC(20)".  You can also cast those columns inline to bypass the issue ie.
```SELECT column_a::NUMERIC(20)```

The solution would probably be to have a specific case of if the precision is 0 (which is technically an invalid precision) to treat it like a DOUBLE to ensure that it is more likely the data fits properly.

lwhite1:
what's the actual PG column type?

edgar131:
> what's the actual PG column type?

Numeric

lwhite1:
Pg doesn’t support the precision operation in its drivers.  It always
returns 0. We’ll need to work around that.  I guess that means you’ll
always get a long.

There is a static method on sqlresultsetreader that lets you specify the
col type for a resultset type tho. You can try that in the meantime

edgar131:
I'll give that a try tomorrow and see if that works, thanks.

> Pg doesn’t support the precision operation in its drivers.  It always
returns 0. We’ll need to work around that.  I guess that means you’ll
always get a long.

> There is a static method on sqlresultsetreader that lets you specify the
col type for a resultset type tho. You can try that in the meantime

[...]

PR #523, opened 2019-05-03
Workaround for precision of 0 when reading jdbc resultset

benmccann:
Thanks for contributing.

## Description

Closes https://github.com/jtablesaw/tablesaw/issues/519. Thanks @edgar131 for the report, investigation, and suggested solution

### DIFF
--- a/core/src/main/java/tech/tablesaw/io/jdbc/SqlResultSetReader.java
+++ b/core/src/main/java/tech/tablesaw/io/jdbc/SqlResultSetReader.java
@@ -24,7 +24,6 @@ import tech.tablesaw.api.LongColumn;
 import tech.tablesaw.api.ShortColumn;
 import tech.tablesaw.api.Table;
 import tech.tablesaw.columns.Column;
-import tech.tablesaw.columns.numbers.ShortColumnType;
 
 import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
@@ -115,14 +114,16 @@ int, Integer:       -2147483648             2147483647          NUMBER(9)       
 long, Long:         -9223372036854775808    9223372036854775807 NUMBER(18)      999_999_999_999_999_999
 
 */
-                    // Start with SHORT (since ColumnType.BYTE isn't supported yet)
-                    // and find the smallest java integer type that fits
-                    if (p <= 4) {
-                        type = ShortColumnType.instance();
-                    } else if (p <= 9) {
-                        type = ColumnType.INTEGER;
-                    } else if (p <= 18) {
-                        type = ColumnType.LONG;
+                    if (p > 0) {
+                        if (p <= 4) {
+                            // Start with SHORT (since ColumnType.BYTE isn't supported yet)
+                            // and find the smallest java integer type that fits
+                            type = ColumnType.SHORT;
+                        } else if (p <= 9) {
+                            type = ColumnType.INTEGER;
+                        } else if (p <= 18) {
+                            type = ColumnType.LONG;
+                        }
                     }
                 } else { // s is not zero, so a decimal value is expected. First try float, then double
                     if (s <= 7) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/tech/tablesaw/io/jdbc/SqlResultSetReader.java
```diff
@@ -115,10 +114,13 @@ public static Table read(ResultSet resultSet) throws SQLException {
 long, Long:         -9223372036854775808    9223372036854775807 NUMBER(18)      999_999_999_999_999_999
 
 */
-                    // Start with SHORT (since ColumnType.BYTE isn't supported yet)
-                    // and find the smallest java integer type that fits
-                    if (p <= 4) {
-                        type = ShortColumnType.instance();
+                    if (p <= 0) {
+                	// Default to double if invalid precision
+                	type = ColumnType.DOUBLE;
```
edgar131: Since both NUMERIC and DECIMAL types are mapped to DOUBLE in SQL_TYPE_TO_TABLESAW_TYPE, it might be better to actually not change the type from what it pulls from the map.  That way if if you use SqlResultSetReader.mapJdbcTypeToColumnType you can have NUMERIC treated as whatever type you want (and by default it would be DOUBLE).

note: I'll still be happy if it is just hardcoded to DOUBLE.

benmccann: Thanks for the review! I've updated per your suggestion. Please take another look

lwhite1:
Hi Ben, 

I'm on vacation and a bit short on time.  Thanks for taking care of this. 

A couple questions:

It looks like the case for p=0, falls through to an error. Is that correct? If so, why not return a LongColumn? It seems like anyone who hits the error will likely open a new issue. 

Could we make the 0 case explicit and add a comment explaining what it's for? It may save us troubleshooting this again down the line.

Could we add a test?

edgar131:
Can't use long because "numeric" types could be decimal numbers.  Has to be double to avoid losing any data.  If you do add an explicit case it is still preferable to not explicitly set it to double so that people can override the type if they want to.

> Hi Ben,
> 
> I'm on vacation and a bit short on time. Thanks for taking care of this.
> 
> A couple questions:
> 
> It looks like the case for p=0, falls through to an error. Is that correct? If so, why not return a LongColumn? It seems like anyone who hits the error will likely open a new issue.
> 
> Could we make the 0 case explicit and add a comment explaining what it's for? It may save us troubleshooting this again down the line.
> 
> Could we add a test?

lwhite1:
> Can't use long because "numeric" types could be decimal numbers.

So the scale value returned by PG is also incorrect? Otherwise you know for a fact it's not a double.

edgar131:
I cant test right now, but pretty sure it was -1 (essentially useless). According to postgres docs...

> NUMERIC without any precision or scale creates a column in which numeric values of any precision and scale can be stored, up to the implementation limit on precision. A column of this kind will not coerce input values to any particular scale, whereas numeric columns with a declared scale will coerce input values to that scale.

Because of that you would have to implement it as at least DOUBLE to ensure that no data is lost.

benmccann:
> It looks like the case for p=0, falls through to an error. Is that correct? If so, why not return a LongColumn? It seems like anyone who hits the error will likely open a new issue.

No, I don't think that quite right. `type` is initialized to `Double` here: https://github.com/jtablesaw/tablesaw/blob/master/core/src/main/java/tech/tablesaw/io/jdbc/SqlResultSetReader.java#L99

This PR will simply leave it as `Double` in the case that `precision <= 0` instead of trying to set it to a more specific numeric type based on the `precision`

> Could we make the 0 case explicit and add a comment explaining what it's for? It may save us troubleshooting this again down the line.

That's what I initially did, but @edgar131 suggested this solution as being more configurable and I think he had a good point. See his comment in https://github.com/jtablesaw/tablesaw/pull/523#discussion_r280852011

> Could we add a test?

Done

PR #528, opened 2019-05-19
Fixed Smile Nominal Attribute creation and added stratified sampling

kiamesdavies:
…column

Thanks for contributing.

## Description
There was a mistake when creating nominal attribute for smile, it uses the length of the original column and not the unique column. The issue is noticed when you have a reasonable dataset of say 10,000 records and you  are trying to use smile.feature.OneHotEncoder, it creates close to 16,000 columns after encoding from  an original 12 columns and sometimes leading to OutOfMemoryException. 

I added stratified sampling as well

## Testing

For the nominal attribute creation, It already had a unit test  it
For the stratified sampling, I added a test **testStratifiesSampleSplit** to check the  that the proportion of foxes in the split table is equal to the proportion specified for the whole table.

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/Table.java
+++ b/core/src/main/java/tech/tablesaw/api/Table.java
@@ -457,26 +457,24 @@ public class Table extends Relation implements Iterable<Row> {
     
     /**
      * Splits the table into two stratified samples, this uses the specified column to divide the table into groups, randomly assigning records to each according to the proportion given in trainingProportion. 
-     * 
-     *
      * @param column the column to be used for the stratified sampling
      * @param table1Proportion The proportion to go in the first table
      * @return An array two tables, with the first table having the proportion specified in the method parameter,
      * and the second table having the balance of the rows
      */
-    public Table[] stratifiedSampleSplit(CategoricalColumn column, double table1Proportion){
+    public Table[] stratifiedSampleSplit(CategoricalColumn column, double table1Proportion) {
         Preconditions.checkArgument( CategoricalColumn.class.isInstance(this.column(column.name())),
                 "The categorical column must be part of the table, you can create a string column and add it to this table before sampling.");
         final Table first = this.emptyCopy();
         final Table second = this.emptyCopy();
         
         this.splitOn(column).asTableList().forEach(tab-> {
-           Table[] splits = tab.sampleSplit(table1Proportion); 
+            Table[] splits = tab.sampleSplit(table1Proportion); 
             first.append(splits[0]);
             second.append(splits[1]);
         });
         
-        return new Table[]{first, second};
+        return new Table[]{ first, second };
     }
 
     /**
--- a/core/src/main/java/tech/tablesaw/api/Table.java
+++ b/core/src/main/java/tech/tablesaw/api/Table.java
@@ -463,7 +463,7 @@ public class Table extends Relation implements Iterable<Row> {
      * and the second table having the balance of the rows
      */
     public Table[] stratifiedSampleSplit(CategoricalColumn column, double table1Proportion) {
-        Preconditions.checkArgument( CategoricalColumn.class.isInstance(this.column(column.name())),
+        Preconditions.checkArgument(containsColumn(column),
                 "The categorical column must be part of the table, you can create a string column and add it to this table before sampling.");
         final Table first = this.emptyCopy();
         final Table second = this.emptyCopy();
--- a/core/src/main/java/tech/tablesaw/api/Table.java
+++ b/core/src/main/java/tech/tablesaw/api/Table.java
@@ -468,7 +468,7 @@ public class Table extends Relation implements Iterable<Row> {
         final Table first = emptyCopy();
         final Table second = emptyCopy();
         
-        this.splitOn(column).asTableList().forEach(tab-> {
+        splitOn(column).asTableList().forEach(tab-> {
             Table[] splits = tab.sampleSplit(table1Proportion); 
             first.append(splits[0]);
             second.append(splits[1]);
--- a/core/src/main/java/tech/tablesaw/api/Table.java
+++ b/core/src/main/java/tech/tablesaw/api/Table.java
@@ -465,8 +465,8 @@ public class Table extends Relation implements Iterable<Row> {
     public Table[] stratifiedSampleSplit(CategoricalColumn column, double table1Proportion) {
         Preconditions.checkArgument(containsColumn(column),
                 "The categorical column must be part of the table, you can create a string column and add it to this table before sampling.");
-        final Table first = this.emptyCopy();
-        final Table second = this.emptyCopy();
+        final Table first = emptyCopy();
+        final Table second = emptyCopy();
         
         this.splitOn(column).asTableList().forEach(tab-> {
             Table[] splits = tab.sampleSplit(table1Proportion); 
--- a/core/src/main/java/tech/tablesaw/api/Table.java
+++ b/core/src/main/java/tech/tablesaw/api/Table.java
@@ -454,6 +454,30 @@ public class Table extends Relation implements Iterable<Row> {
         tables[1] = where(table2Selection);
         return tables;
     }
+    
+    /**
+     * Splits the table into two stratified samples, this uses the specified column to divide the table into groups, randomly assigning records to each according to the proportion given in trainingProportion. 
+     * 
+     *
+     * @param column the column to be used for the stratified sampling
+     * @param table1Proportion The proportion to go in the first table
+     * @return An array two tables, with the first table having the proportion specified in the method parameter,
+     * and the second table having the balance of the rows
+     */
+    public Table[] stratifiedSampleSplit(CategoricalColumn column, double table1Proportion){
+        Preconditions.checkArgument( CategoricalColumn.class.isInstance(this.column(column.name())),
+                "The categorical column must be part of the table, you can create a string column and add it to this table before sampling.");
+        final Table first = this.emptyCopy();
+        final Table second = this.emptyCopy();
+        
+        this.splitOn(column).asTableList().forEach(tab-> {
+           Table[] splits = tab.sampleSplit(table1Proportion); 
+            first.append(splits[0]);
+            second.append(splits[1]);
+        });
+        
+        return new Table[]{first, second};
+    }
 
     /**
      * Returns a table consisting of randomly selected records from this table. The sample size is based on the
--- a/core/src/test/java/tech/tablesaw/api/TableTest.java
+++ b/core/src/test/java/tech/tablesaw/api/TableTest.java
@@ -214,6 +214,17 @@ public class TableTest {
         Table[] results = t.sampleSplit(.75);
         assertEquals(t.rowCount(), results[0].rowCount() + results[1].rowCount());
     }
+    
+    @Test
+    public void testStratifiedSampleSplit() throws Exception {
+        Table t = Table.read().csv("../data/bush.csv");
+        Table[] results = t.stratifiedSampleSplit(t.stringColumn("who"), .75);
+        assertEquals(t.rowCount(), results[0].rowCount() + results[1].rowCount());
+        int totalFoxCount = t.where(t.stringColumn("who").equalsIgnoreCase("fox")).rowCount();
+        int stratifiedFoxCount = results[0].where(results[0].stringColumn("who").equalsIgnoreCase("fox")).rowCount();
+        
+        assertEquals(.75, (double)stratifiedFoxCount/totalFoxCount, 0.0);
+    }
 
     @Test
     public void testDoWithEachRow() throws Exception {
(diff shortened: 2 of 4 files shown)


core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -454,6 +454,30 @@ public Table emptyCopy(int rowSize) {
...
+     * Splits the table into two stratified samples, this uses the specified column to divide the table into groups, randomly assigning records to each according to the proportion given in trainingProportion. 
+     * 
+     *
+     * @param column the column to be used for the stratified sampling
+     * @param table1Proportion The proportion to go in the first table
+     * @return An array two tables, with the first table having the proportion specified in the method parameter,
+     * and the second table having the balance of the rows
+     */
+    public Table[] stratifiedSampleSplit(CategoricalColumn column, double table1Proportion){
+        Preconditions.checkArgument( CategoricalColumn.class.isInstance(this.column(column.name())),
```
benmccann: Why not do `isInstance(column)` instead of looking up the column we already have from the table?

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -454,6 +454,30 @@ public Table emptyCopy(int rowSize) {
         tables[1] = where(table2Selection);
         return tables;
     }
+    
+    /**
+     * Splits the table into two stratified samples, this uses the specified column to divide the table into groups, randomly assigning records to each according to the proportion given in trainingProportion. 
+     *
```
benmccann: no need for two blank lines here

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -454,6 +454,30 @@ public Table emptyCopy(int rowSize) {
...
+    /**
+     * Splits the table into two stratified samples, this uses the specified column to divide the table into groups, randomly assigning records to each according to the proportion given in trainingProportion. 
+     * 
+     *
+     * @param column the column to be used for the stratified sampling
+     * @param table1Proportion The proportion to go in the first table
+     * @return An array two tables, with the first table having the proportion specified in the method parameter,
+     * and the second table having the balance of the rows
+     */
+    public Table[] stratifiedSampleSplit(CategoricalColumn column, double table1Proportion){
```
benmccann: please add a space before the `{`

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -454,6 +454,30 @@ public Table emptyCopy(int rowSize) {
...
+        Preconditions.checkArgument( CategoricalColumn.class.isInstance(this.column(column.name())),
+                "The categorical column must be part of the table, you can create a string column and add it to this table before sampling.");
+        final Table first = this.emptyCopy();
+        final Table second = this.emptyCopy();
+        
+        this.splitOn(column).asTableList().forEach(tab-> {
+           Table[] splits = tab.sampleSplit(table1Proportion);
```
benmccann: the indentation is off on this line

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -454,6 +454,30 @@ public Table emptyCopy(int rowSize) {
...
+        final Table first = this.emptyCopy();
+        final Table second = this.emptyCopy();
+        
+        this.splitOn(column).asTableList().forEach(tab-> {
+           Table[] splits = tab.sampleSplit(table1Proportion); 
+            first.append(splits[0]);
+            second.append(splits[1]);
+        });
+        
+        return new Table[]{first, second};
```
benmccann: can you add a space after the `{` and before the `}`

core/src/test/java/tech/tablesaw/api/TableTest.java
```diff
@@ -214,6 +214,17 @@ public void testSampleSplit() throws Exception {
...
+    
+    @Test
+    public void testStratifiedSampleSplit() throws Exception {
+        Table t = Table.read().csv("../data/bush.csv");
+        Table[] results = t.stratifiedSampleSplit(t.stringColumn("who"), .75);
+        assertEquals(t.rowCount(), results[0].rowCount() + results[1].rowCount());
+        int totalFoxCount = t.where(t.stringColumn("who").equalsIgnoreCase("fox")).rowCount();
+        int stratifiedFoxCount = results[0].where(results[0].stringColumn("who").equalsIgnoreCase("fox")).rowCount();
+        
+        assertEquals(.75, (double)stratifiedFoxCount/totalFoxCount, 0.0);
```
benmccann: please add a space after `(double)`

core/src/main/java/tech/tablesaw/api/Table.java
benmccann: no need to check if it's a `CategoricalColumn`. we know it is from the method signature. I would do `containsColumn(column)` instead

kiamesdavies: A user doing fancy stuff like table.nCol("xxx").asStringColumn() will cause issue

core/src/main/java/tech/tablesaw/api/Table.java
kiamesdavies: fixed

core/src/main/java/tech/tablesaw/api/Table.java
kiamesdavies: fixed

core/src/main/java/tech/tablesaw/api/Table.java
kiamesdavies: fixed

core/src/main/java/tech/tablesaw/api/Table.java
kiamesdavies: fixed

core/src/test/java/tech/tablesaw/api/TableTest.java
kiamesdavies: fixed

core/src/main/java/tech/tablesaw/api/Table.java
benmccann: If you do `.asStringColumn()` it will create a new column not in the table, so `containsColumn(column)` should still work in that case

benmccann:
Thanks for this @kiamesdavies. It looks good to me. Can you add a regression test for the change to `SmileConverter`? There are some existing tests in `SmileConverterTest`

kiamesdavies:
@benmccann when will you merge this to master, planning to add a Weka Converter soon, not satisfied with Smile ML .

benmccann:
I was going to give @lwhite1 a chance to review

kiamesdavies:
Roger

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -454,6 +454,28 @@ public Table emptyCopy(int rowSize) {
...
+     * Splits the table into two stratified samples, this uses the specified column to divide the table into groups, randomly assigning records to each according to the proportion given in trainingProportion. 
+     * @param column the column to be used for the stratified sampling
+     * @param table1Proportion The proportion to go in the first table
+     * @return An array two tables, with the first table having the proportion specified in the method parameter,
+     * and the second table having the balance of the rows
+     */
+    public Table[] stratifiedSampleSplit(CategoricalColumn column, double table1Proportion) {
+        Preconditions.checkArgument(containsColumn(column),
+                "The categorical column must be part of the table, you can create a string column and add it to this table before sampling.");
+        final Table first = this.emptyCopy();
```
lwhite1: i do this all the time, but it's probably best if we don't use **this.** unless it's needed.

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -454,6 +454,28 @@ public Table emptyCopy(int rowSize) {
...
         return tables;
     }
+    
+    /**
+     * Splits the table into two stratified samples, this uses the specified column to divide the table into groups, randomly assigning records to each according to the proportion given in trainingProportion. 
+     * @param column the column to be used for the stratified sampling
+     * @param table1Proportion The proportion to go in the first table
+     * @return An array two tables, with the first table having the proportion specified in the method parameter,
+     * and the second table having the balance of the rows
+     */
```
lwhite1: IIRC the tables will be returned based on the alphabetical order of the category column. That should probably be specified in the comments because the proportion variable is associated with the first table. 

Should the method name indicate that it's only for two-way splits?  It seems like it might be useful in the future to support n-way splits where n is the cardinality of the category column.  If we ever implement the general solution, it might be helpful to name this method so it's clear that it isn't general.

benmccann: If we made it more general, we'd have to change the parameters, so I tend to be okay with the name as it is

core/src/main/java/tech/tablesaw/api/Table.java
benmccann: @kiamesdavies would you be able to address this comment so that we can merge the PR?

core/src/main/java/tech/tablesaw/api/Table.java
lwhite1: The more I look at this, the less I think it should be added as written. 

It's not really like the existing split method. which isn't dependent on the contents of a column. It just randomly divides the table into two parts based on the specified proportion.  The use case is creating a simple hold-out sample for testing a model. The order of the tables returned is clear, the first table returned is the one with the specified size or proportion. 

This one claims to be doing stratified sampling. According to wikipedia: 

> "The strata should define a partition of the population. That is, it should be collectively exhaustive and mutually exclusive: every element in the population must be assigned to one and only one stratum."

Based on that definition (which I believe to be the common one), this method is problematic. What happens if there are more than two categories in the input column? You get a result, but any data in categories 3 through n is simply lost. That violates the rule that every element be assigned to a stratum and is a bug waiting to happen. 

The second issue is that the result is dependent on the sorting of the tables returned, which I expect will be alphabetical by category name, although the split method doesn't specify that explicitly so conceivably that could change. Even if it doesn't, this is another bug waiting to happen. You could use it in an application and the categories are renamed, or another category could be added with a name that sorts first.

Let me go back to wikipedia: 

> Stratified sampling strategies
Proportionate allocation uses a sampling fraction in each of the strata that is proportional to that of the total population. For instance, if the population consists of X total individuals, m of which are male and f female (and where m + f = X), then the relative size of the two samples (x1 = m/X males, x2 = f/X females) should reflect this proportion.
Optimum allocation (or disproportionate allocation) - The sampling fraction of each stratum is proportionate to both the proportion (as above) and the standard deviation of the distribution of the variable. Larger samples are taken in the strata with the greatest variability to generate the least possible overall sampling variance.

These are the only two strategies wikipedia lists. The first one, I think, is the usual approach. That's what this method should do by default, and I would love to have a method that does that, or better still, two methods that implement both approaches. 

This method is arbitrary about what goes in each strata. If you want arbitrarily sized samples by category and only want to handle the case where there are two categories, I think that's a very specific use case that should be left to the developer to handle. 

This should be re-written to implement proportional sampling for n categories or it should be removed.

benmccann: > What happens if there are more than two categories in the input column? You get a result, but any data in categories 3 through n is simply lost.

I don't think so. Categories 3 and onward will be sampled according to the given proportion and then added to the result arrays. I do think the code is slightly confusingly written and also read it the same way you did at first, but I don't think what you described is what's actually happening

> This one claims to be doing stratified sampling.

I may have misunderstood the Wikipedia article or the code, but I think this does do proportionate allocation

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -454,6 +454,28 @@ public Table emptyCopy(int rowSize) {
...
+    public Table[] stratifiedSampleSplit(CategoricalColumn column, double table1Proportion) {
```
benmccann: Please fix the compilation warning on this line:
> CategoricalColumn is a raw type. References to generic type CategoricalColumn<T> should be parameterized

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -454,6 +454,28 @@ public Table emptyCopy(int rowSize) {
...
+        Preconditions.checkArgument(containsColumn(column),
```
benmccann: This check might better belong in `splitOn`

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -454,6 +454,28 @@ public Table emptyCopy(int rowSize) {
...
+                "The categorical column must be part of the table, you can create a string column and add it to this table before sampling.");
+        final Table first = emptyCopy();
```
benmccann: I think you could rewrite this method as:

```
        return splitOn(column).asTableList().stream().map(t -> t.sampleSplit(table1Proportion)).reduce(new Table[2], (acc, res) -> {
            acc[0].append(res[0]);
            acc[1].append(res[1]);
            return acc;
        });
```

core/src/main/java/tech/tablesaw/api/Table.java
benmccann: @kiamesdavies could you take a look at my last round of comments?

core/src/main/java/tech/tablesaw/api/Table.java
kiamesdavies: I will take a look at it this week, sorry for the delay

PR #530, opened 2019-06-01
Add set via predicate method

benmccann:
Thanks for contributing.

## Description

I had some code using Tablesaw 0.24.9 that used `fillMissing` in a chainable fastion:
```
table.addColumns(table.numberColumn("A").pctChange().fillMissing(table.numberColumn("B").pctChange()).fillMissing(table.numberColumn("C").pctChange()).fillMissing(table.numberColumn("D").pctChange()))
```

[`fillMissing` was removed since it's too specialized](https://github.com/jtablesaw/tablesaw/commit/ceea9e525ed1a1470745cb13c671f6ee7f2ffde5#diff-fe7c0c804c9d68f3e7080cc5210e784d). However, the only way to accomplish `fillMissing` today is via `col1.set(col1.isMissing(), col2)` which is not chainable. That makes writing the code above very difficult.

With this change I can do `col1.set(isMissing(), col2)`, which is much easier to write complex expressions with.

## Testing

Added two unit tests


lwhite1:
i think this does more or less the same thing: 

```
    @Override
    public IntColumn set(Selection condition, Column<Integer> other) {
        return (IntColumn) super.set(condition, other);
    }

    @Override
    public IntColumn set(Selection rowSelection, Integer newValue) {
        return (IntColumn) super.set(rowSelection, newValue);
    }

```

benmccann:
The difference is that this is chainable so I can use it in one line:
```
table.numberColumn("A").pctChange().set(isMissing, table.numberColumn("B").pctChange());
```

Otherwise I can't chain it and need to use multiple lines:
```
DoubleColumn col = table.numberColumn("A").pctChange();
col.set(col.isMissing, table.numberColumn("B").pctChange());
```

lwhite1:
As usual, I'm missing something.  I just tested:

`yCol.pctChange().set(yCol.isMissing(), xCol.pctChange());`

where yCol and xCol are DoubleColumns. How is this not chainable?

if the difference is in the return type, why not change the existing
method's return type? If what you really want is to use predicates rather
than selections, I think that should be done universally (assuming it can
be) rather than introducing more inconsistency in the API.

benmccann:
Maybe chainable wasn't the best way to describe it. The issue is that I need to create an intermediate variable for the column, which means multiple lines vs just one. [My example](https://github.com/jtablesaw/tablesaw/pull/530#issuecomment-498055303) demonstrates writing the same thing with 1 line vs multiple

What inconsistency are you referring to? My goal was for it to be possible to call these new `set` methods on any numeric column type. I just added additional methods to the other column types to cast the return value and make that clearer.

lwhite1:
> Maybe chainable wasn't the best way to describe it. The issue is that I need to create an intermediate variable for the column, which means multiple lines vs just one. [My example](https://github.com/jtablesaw/tablesaw/pull/530#issuecomment-498055303) demonstrates writing the same thing with 1 line vs multiple

ok. I see the difference now.

> 
> What inconsistency are you referring to? My goal was for it to be possible to call these new `set` methods on any numeric column type. I just added additional methods to the other column types to cast the return value and make that clearer.

I just meant having set() methods that use predicates in addition to set methods that use selections. Originally methods were written using selections and new methods have been added in various places using predicates, but there's not necessarily a predicate-based alternative for every selection-based method. Here's kind of a big example: Where clauses on table use selections only.

(also, i do get that predicates are more consistent with general Java use than selections. Selections just worked well for efficient querying so they're ubiquitous in our code.) 

FWIW, from a usability standpoint, the inability to do what you want to do here is a much bigger issue in the table query context. We lost the ability to refer to a result column in a sub-clause a while back and it's a significant handicap relative to what you can do in SQL. I don't know if this approach could work in where clauses as well, but it would be cool if it did. 

Go ahead and merge this. It's a nice addition.

benmccann:
Thanks for the review!

PR #531, opened 2019-06-01
Rename join to joinOn to make it work in BeakerX

benmccann:
Thanks for contributing.

## Description

Renamed `join` to `joinOn` so that it can be easily called from Groovy. Because `Table` implements `Iterable`, Groovy's `Iterable.join` method clobbers our own. Renaming makes it much easier to use in BeakerX.

See my [post on StackOverflow](https://stackoverflow.com/questions/56401713/prevent-groovydefaultmethods-join-from-shadowing-join-method-of-java-class) for full details of the issue.

## Testing

Updated existing unit tests


lwhite1:
seems ok to me.

PR #533, opened 2019-06-05
Remove duplicate map function

benmccann:
Thanks for contributing.

## Description

I couldn't call `map` because it would result in a compilation error:

>The method map is ambiguous for the type DoubleColumn

It'd be nice to have a `map` method taking `DoubleUnaryOperator` to avoid auto-boxing, but then we'd have to remove the generic `map` on `Column`. The Java collections avoid auto-boxing and method collisions by having the `map` method on the `stream()`. We should probably do the same after https://github.com/vigna/fastutil/pull/120 is merged

## Testing

Added a unit test


benmccann:
@lwhite1 any thoughts on the most recent handful of PRs?

lwhite1:
They look good. Go ahead and merge

benmccann:
Thanks!

PR #539, opened 2019-06-18
Add fillWith(double)

benmccann:
Thanks for contributing.

## Description

Now you can do `fillWith(1.0)` instead of `fillWith(() -> 1.0)`

## Testing

Added a unit test


lwhite1:
i see the need for fillWith(double), but is the supplier version not also sometimes useful? IDK, like if you wanted to fill with random numbers, or a sequence of some kind.

benmccann:
Great point! I hadn't been able to think of a usecase, but that's a good example. I've added it back

lwhite1:
LGTM

PR #544, opened 2019-07-01
Correct issue #541 : unique values in StringColumns

### LINKED ISSUE #541: Column.countUnique() returns a wrong value after using Column.set()
Hi,

After assigning all values to a previously empty Column, the empty value is still counted by countUnique().  Below is a simple test case demonstrating the issue. The count is correct when the column is created with no size and append() is used (2nd test case).

```
import static org.junit.Assert.*;
import org.junit.Test;
import tech.tablesaw.api.StringColumn;

public class TestTablesawUniqueValues {

	@Test
	public void testUniqueValuesAfterSet() {
		StringColumn col = StringColumn.create("col1", 2);
		col.set(0, "A");
		col.set(1, "B");
		assertEquals("Error in unique() after set()", 2, col.countUnique());
	}

	@Test
	public void testUniqueValuesAfterAppend() {
		StringColumn col = StringColumn.create("col1");
		col.append("A");
		col.append("B");
		assertEquals("Error in unique() after append()", 2, col.countUnique());
	}
}
```
Test result:
java.lang.AssertionError: Error in unique() after set() expected:<2> but was:<3>

Tested on version 0.34.1

Thanks for tablesaw !

ccleva:
Thanks for contributing.

## Description

Implemented instance counting in all DictionaryMap implementations.

key/values are removed from all maps when value count reaches 0, so all operations based on these maps are always correct.
The only modified methods are set(), append() and clear(). There is no impact on other methods.
All count-based methods now uses the count map for performance.

Fixes #541 

## Testing

Added unit tests showing previous issue with StringColumn unique values counting

## Notes

* The new map member references the implementation used instead of the interface so we can use the addTo() method
* Now unused private contains() method was removed

### DIFF
--- a/core/src/main/java/tech/tablesaw/columns/strings/ByteDictionaryMap.java
+++ b/core/src/main/java/tech/tablesaw/columns/strings/ByteDictionaryMap.java
@@ -90,16 +90,6 @@ public class ByteDictionaryMap implements DictionaryMap {
         return values.getByte(rowIndex);
     }
 
-    /**
-     * Returns true if we have seen this stringValue before, and it hasn't been removed.
-     * <p>
-     * NOTE: An answer of true does not imply that the column still contains the value, only that
-     * it is in the dictionary map
-     */
-    private boolean contains(String stringValue) {
-        return valueToKey.containsKey(stringValue);
-    }
-
     private Set<String> categories() {
         return valueToKey.keySet();
     }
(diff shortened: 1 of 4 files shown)


lwhite1:
@ccleva 
Thanks very much for doing this.

It looks good, and I appreciate the tests. I would feel a little better if there were tests that exercised the short and int versions of the dictionary map. Adding stringified ints to the column might be easiest. 

What do you think? Is that doable?

ccleva:
Good point, I will write additional tests for those cases.

ccleva:
This should be better now: all 3 dictionary implementations are tested and there are more assertions in the unit tests. Let me know if you think of other improvements. Thanks

PR #545, opened 2019-07-06
Add mapToType Columns

### LINKED ISSUE #538: Column-wise DataFrame-like operations
Hi,
I am new to Tablesaw. I am exploring options for recreating some Pyhton DataFrame operations in Tablesaw. For example, I have a DataFrame object called data1 and I use existing columns of this data frame to create new ones (and update existing ones). Here is a couple of lines of Pyhton codes:

data1['days'] = data1['buyDate'].apply(lambda x: (today - x).days)
...
data1['CAGR'] = ((data1['curValue'] / data1['bookValue']) ** (1.0 / data1['nYears']) - 1.0) * 100

Is there a way to implement something similar using Tablesaw? 

Thanks a lot in advance.

ryancerf:

## Description

Helps with #538

Added mapToType columns making it easier to map a column to a different type. Such as int -> LocalDate. The mapToType methods are a then wrapper onto the existing mapInto method.

## Testing

Added unit tests for a single type. Can add unit tests for every type if needed.

### DIFF
--- a/core/src/main/java/tech/tablesaw/columns/AbstractColumn.java
+++ b/core/src/main/java/tech/tablesaw/columns/AbstractColumn.java
@@ -14,8 +14,24 @@
 
 package tech.tablesaw.columns;
 
+import java.time.Instant;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.util.function.Function;
+import tech.tablesaw.api.BooleanColumn;
 import tech.tablesaw.api.ColumnType;
+import tech.tablesaw.api.DateColumn;
+import tech.tablesaw.api.DateTimeColumn;
+import tech.tablesaw.api.DoubleColumn;
+import tech.tablesaw.api.FloatColumn;
+import tech.tablesaw.api.InstantColumn;
+import tech.tablesaw.api.IntColumn;
+import tech.tablesaw.api.LongColumn;
+import tech.tablesaw.api.ShortColumn;
 import tech.tablesaw.api.StringColumn;
+import tech.tablesaw.api.TextColumn;
+import tech.tablesaw.api.TimeColumn;
 
 /**
  * Partial implementation of the {@link Column} interface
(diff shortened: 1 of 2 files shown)


lwhite1:
@ryancerf 
Thanks very much. I think this will be helpful for many people and should be added to the library. 

I'm a little concerned that these methods would autobox (i think) every value in the column, whereas the non-standard approach used in, say, NumberMapFunctions would not. 

If they _do_ autobox, perhaps the method comments could include a note about the memory implications and suggest the reader, if they would like to avoid that, look at NumberMapFunctions as an example of an alternate approach.

ryancerf:
Thanks for the comments Larry. I agree that autoboxing is a concern. I'll
try benchmarking this approach and the standard number map functions
approach using jmh to see how they compare.

Unfortunately I hurt my shoulder and will have to pick this up in a week or
two.

lwhite1:
@ryancerf Sorry to hear about your shoulder. Hope it's nothing to serious.

ryancerf:
The TLDR is that I thought the lambda was autoboxing, but the jmh tests appear to show that it is not. It is possible the compiler is working its magic to eliminate the object creation.

**Details on the JMH Performance test.**
I generated a DoubleColumn with 10 million random doubles and tested the performance of squaring the double column using using the MapToType implementation proposed in this pull request and using power from NumberMapFunctions. I did not find any performance difference between the two. The ops/sec and memory numbers such as GC allocation rate  and GC time are almost fairly close.

For more information on what I think is going on see the escape analysis section of this blog post from Brian Goetz. https://www.ibm.com/developerworks/java/library/j-jtp09275/index.html


**Here are the full results from the jmh test** 

|Benchmark|Mode                         |Cnt   |Score                                        |Error      |Units |
|---------|-----------------------------|------|---------------------------------------------|-----------|------|
|mapWithMapToType|avgt                         |10    |0.473                                        |0.03       |s/op  |
|mapWithMapToType:·gc.alloc.rate|avgt                         |10    |1069.211                                     |62.322     |MB/sec|
|mapWithMapToType:·gc.alloc.rate.norm|avgt                         |10    |555972352.2                                  |12.908     |B/op  |
|mapWithMapToType:·gc.churn.PS_Eden_Space|avgt                         |10    |1074.102                                     |52.706     |MB/sec|
|mapWithMapToType:·gc.churn.PS_Eden_Space.norm|avgt                         |10    |558754902.8                                  |18725768.09|B/op  |
|mapWithMapToType:·gc.churn.PS_Old_Gen|avgt                         |10    |180.841                                      |44.608     |MB/sec|
|mapWithMapToType:·gc.churn.PS_Old_Gen.norm|avgt                         |10    |94462542.29                                  |26817241.13|B/op  |
|mapWithMapToType:·gc.churn.PS_Survivor_Space|avgt                         |10    |1.402                                        |0.28       |MB/sec|
|mapWithMapToType:·gc.churn.PS_Survivor_Space.norm|avgt                         |10    |731463.387                                   |167266.292 |B/op  |
|mapWithMapToType:·gc.count|avgt                         |10    |207                                          |           |counts|
|mapWithMapToType:·gc.time|avgt                         |10    |5806                                         |           |ms    |
|mapWithNumberFunctions|avgt                         |10    |0.456                                        |0.024      |s/op  |
|mapWithNumberFunctions:·gc.alloc.rate|avgt                         |10    |1110.355                                     |55.739     |MB/sec|
|mapWithNumberFunctions:·gc.alloc.rate.norm|avgt                         |10    |556133458.1                                  |543646.042 |B/op  |
|mapWithNumberFunctions:·gc.churn.PS_Eden_Space|avgt                         |10    |1107.155                                     |85.54      |MB/sec|
|mapWithNumberFunctions:·gc.churn.PS_Eden_Space.norm|avgt                         |10    |554314583.8                                  |21769623.98|B/op  |
|mapWithNumberFunctions:·gc.churn.PS_Old_Gen|avgt                         |10    |176.171                                      |57.088     |MB/sec|
|mapWithNumberFunctions:·gc.churn.PS_Old_Gen.norm|avgt                         |10    |88300722.93                                  |28616572.39|B/op  |
|mapWithNumberFunctions:·gc.churn.PS_Survivor_Space|avgt                         |10    |5.538                                        |6.897      |MB/sec|
|mapWithNumberFunctions:·gc.churn.PS_Survivor_Space.norm|avgt                         |10    |2757019.554                                  |3379577.949|B/op  |
|mapWithNumberFunctions:·gc.count|avgt                         |10    |215                                          |           |counts|
|mapWithNumberFunctions:·gc.time|avgt                         |10    |4616                                         |           |ms    |


**Here is the code.**

Note I included some base cases of iterating over DoubleStream and Stream<Double> to confirm auto boxing was slowing things down where I expected it too.

```java
@Fork(2)
@Warmup(iterations = 5)
@Measurement(iterations = 5)
@State(Scope.Benchmark)
@BenchmarkMode(Mode.AverageTime)
public class BoxingBenchmark {

  @State(Scope.Thread)
  public static class IterationEnvironment {

    DoubleStream getDoubleStream() {
      Random random = new Random(2000L);
      return random.doubles().limit(10_000_000);
    }

    Stream<Double> getStream() {
      return getDoubleStream().boxed();
    }

    DoubleColumn getDoubleColumn() {
      return DoubleColumn.create("column", getDoubleStream().toArray());
    }
  }

  @Benchmark
  // No Autoboxing. Show that jmh is working as expected
  public void baseCaseMap(
      final IterationEnvironment iterationEnvironment, final Blackhole blackhole) {
    iterationEnvironment.getDoubleStream()
        .forEach(value -> {
          int toAdd = 40;
          value  += toAdd;
          blackhole.consume(value);
        });
  }

  @Benchmark
  // Autoboxing. Show that autoboxing is slower and has GC overhead.
  public void baseCaseMapWithAutobox(
      final IterationEnvironment iterationEnvironment, final Blackhole blackhole) {
    iterationEnvironment.getStream()
        .forEach(value -> {
          Integer toAdd = 40;
          value  += toAdd;
          blackhole.consume(value);
        });
  }

  @Benchmark
  public void mapWithMapToType(
      final IterationEnvironment iterationEnvironment, final Blackhole blackhole) {
    DoubleColumn result =
        mapToDoubleColumn(
            iterationEnvironment.getDoubleColumn(), "column", value -> Math.pow(value, 2));
    blackhole.consume(result);
  }

  @Benchmark
  public void mapWithNumberFunctions(
      final IterationEnvironment iterationEnvironment, final Blackhole blackhole) {
    DoubleColumn result = iterationEnvironment.getDoubleColumn().power(2);
    blackhole.consume(result);
  }

  // Same as proposed implementation in the pull request.
  DoubleColumn mapToDoubleColumn(
      DoubleColumn doubleColumn, String name, Function<? super Double, ? extends Double> func) {
    DoubleColumn newColumn = DoubleColumn.create(name, doubleColumn.size());
    doubleColumn.mapInto(func, newColumn);
    return newColumn;
  }
}
```

PR #546, opened 2019-07-08
initial improvements for test outcomes on Windows - improves #441 (tests on …

### LINKED ISSUE #441: Some unit tests fail on Windows
We need someone with a Windows machine who can fix and test this issue. Help would be greatly appreciated.

paulk-asert:
…Windows)

Thanks for contributing.

- [ X] Tick to sign-off your agreement to the [Developer Certificate of Origin (DCO) 1.1](https://developercertificate.org)

## Description

What was changed? BOM handling is now catered for.

## Testing

Did you add a unit test? The current unit tests now pass on Windows. I haven't included adding:

os:
  - windows
  - linux
  - osx

to travis CI config in this PR but let me know if you want me to add it.

### DIFF
--- a/core/src/main/java/tech/tablesaw/io/Source.java
+++ b/core/src/main/java/tech/tablesaw/io/Source.java
@@ -15,16 +15,14 @@ import java.util.Scanner;
 
 public class Source {
 
+    // we always have one of these (file, reader, or inputStream)
     protected final File file;
     protected final Reader reader;
     protected final InputStream inputStream;
     protected final Charset charset;
 
     public Source(File file) {
-        this.file = file;
-        this.reader = null;
-        this.inputStream = null;
-        this.charset = Charset.defaultCharset();
+        this(file, Charset.defaultCharset());
     }
 
     public Source(File file, Charset charset) {
@@ -42,10 +40,7 @@ public class Source {
     }
 
     public Source(InputStream inputStream) {
-        this.file = null;
-        this.reader = null;
-        this.inputStream = inputStream;
-        this.charset = Charset.defaultCharset();
+        this(inputStream, Charset.defaultCharset());
     }
 
     public Source(InputStream inputStream, Charset charset) {
--- a/core/src/main/java/tech/tablesaw/io/fixed/FixedWidthReadOptions.java
+++ b/core/src/main/java/tech/tablesaw/io/fixed/FixedWidthReadOptions.java
@@ -29,7 +29,6 @@ import java.util.Locale;
 
 public class FixedWidthReadOptions extends ReadOptions {
 
-    // we always have one of these (file, reader, or inputStream)
     private final ColumnType[] columnTypes;
     private final FixedWidthFields columnSpecs;
     private final String lineEnding;
@@ -172,6 +171,10 @@ public class FixedWidthReadOptions extends ReadOptions {
         public Builder lineEnding(String lineEnding) {
             this.lineEnding = lineEnding;
             return this;
+        }
+
+        public Builder systemLineEnding() {
+            return lineEnding(System.lineSeparator());
         }
 
         public Builder padding(char padding) {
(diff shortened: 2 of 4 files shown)


paulk-asert:
Note: this is a minimal solution - it assumes you are happy with specifying the charset for the troublesome cases (which is certainly a valid stance to take) and it currently only handles the UTF-8 case. If you want to not have to specify the charset and have it work for e.g. UTF-16 cases, you need something more like what Groovy does in CharsetToolkit:

https://github.com/apache/groovy/blob/master/src/main/java/groovy/util/CharsetToolkit.java#L387

My thinking was that you could start with the minimal solution and always add the more complicated version if there was demand.

core/src/main/java/tech/tablesaw/io/Source.java
```diff
@@ -19,6 +20,11 @@
     protected final Reader reader;
     protected final InputStream inputStream;
     protected final Charset charset;
+    private final Charset UTF8 = Charset.forName("UTF-8");
+
+    public Source(String filename) {
```
benmccann: I'd rather not add so many new constructors to `Source`. If we want to make it easier to call I'd probably do it by adding new methods to `CsvReadOptions`. E.g. where we have `CsvReadOptions.builderFromFile` maybe add a new one with a charset parameter

core/src/main/java/tech/tablesaw/io/Source.java
```diff
@@ -84,12 +106,37 @@ public Reader createReader(byte[] cachedBytes) throws IOException {
...
+    /**
+     *  Returns a new Reader on the underlying source object.
+     */
+    private Reader getReader(InputStream raw) throws IOException {
+        // we want to remove the BOM windows adds from a file if the encoding is UTF-8
+        // in other cases we depend on the charsets
+        InputStream in = new BufferedInputStream(raw);
+        if (UTF8.name().equalsIgnoreCase(charset.name())) {
+            in.mark(3);
+            boolean hasBOM = true;
```
benmccann: We use univocity-parsers, which is supposed to do BOM handling for us. It's not clear to me why this would be required. Can we either change our test or file a bug with univocity so that we don't have to do any special logic here?

paulk-asert: Okay, I didn't know about BOM handling within univocity-parsers. It seems like you need to explicitly make use of the BomInput class for univocity-parsers to do what is needed. I'll see if I can rework the PR based around that.

paulk-asert:
I moved the less controversial changes into this PR and will create a separate PR for the BOM changes. So this now fixes two of the tests plus does some minor refactoring (moved what looked like a stray misplaced comment and removed some minor duplication in a couple of constructors).

core/src/main/java/tech/tablesaw/io/fixed/FixedWidthReadOptions.java
```diff
@@ -174,6 +173,10 @@ public Builder lineEnding(String lineEnding) {
             return this;
         }
 
+        public Builder platformLineEnding() {
+            return lineEnding(System.getProperty("line.separator"));
```
benmccann: I'd suggest `System.lineSeparator()` over `System.getProperty("line.separator")`

core/src/main/java/tech/tablesaw/io/fixed/FixedWidthReadOptions.java
```diff
@@ -174,6 +173,10 @@ public Builder lineEnding(String lineEnding) {
             return this;
         }
 
+        public Builder platformLineEnding() {
```
benmccann: How about calling this `systemLineEnding` for greater consistency with the Java API?

paulk-asert: Done

PR #562, opened 2019-08-03
Speed up Joins

### LINKED ISSUE #383: Joins are too slow
Hi guys!
I'm trying to migrate from `python`+`pandas` to `kotlin`+`tablesaw`.
Some parts of my code are already working fast (like csv parsing, x2 times faster than in pandas)
But also i've noticed that `inner join` operation is pretty slow (~ 2 times slower than in pandas)

Then i tried to optimize my code and use `isIn` Selection instead of simple join. Unfortunately it uses `strings.toArray(new String[0])` under the hood for input parameter collection. It would be more sense to use HashSet to quicker lookup. So i wrote my own predicate:
```kotlin
val customersSet = customers.toHashSet()    // for faster lookup
val idColumn = transactions.stringColumn("customer_id")
idColumn.eval { it in customersSet }
```
Which is x15 times faster than original inner join. At least on my huge dataset. Of course this is much simpler than `join` operation since i haven't appended columns etc. But still the difference is huge.
I didn't investigate joins code yet, but i hope there is space for improvements there.
My key point is: kotlin+jvm should be at least not slower than python+pandas
What do you guys think?

ps: do you use hash indexing on table columns?

ryancerf:
Thanks for contributing.

## Description 

Fixes  #383

* Speeds up joins. Especially in the worst cases.
  * In one case a full outer join with a large number of columns went from over 5 mins to a few seconds.
* The performance is less sensitive to which table is on the left hand side of the join, but the join is still noticeably faster when the small table is on the left. This is a possible future improvement.
* Avoids using where and copying columns twice.
* Most of the join time is now spent calculating the cross product (as it should be).
* Makes no API changes.

I fixed a somewhat related issue. If the right hand side table has columns with duplicate names the duplicate columns will be renamed with a prefix. The old implementation mutated the columns on the original table. This PR changes that so the original columns are not changed. See the one added test.

There was good test coverage (Thanks Ben) and all tests are passing.

# Performance

Given a customers table and an orders table with a customerId uniformly distributed among the customerIds in the customers table

> Note: These performance tests do not use a mico benchmarking framework and have a ton of variance.

|num Order rows|num Customer rows|num Orders Cols|num Customers Columns|left hand table in join|new code (seconds)|old code (seconds)|
|---------------|-----------------|---------------|---------------------|-----------------------|------------------|------------------|
|1000000        |1000             |1              |1                    |customers              |0.143             |0.208             |
|1000000        |1000             |11             |1                    |customers              |0.901             |1.015             |
|1000000        |1000             |1              |11                   |customers              |0.364             |0.265             |
|1000000        |1000             |1              |1                    |customers              |0.083             |0.105             |
|1000000        |1000             |1              |6                    |customers              |0.244             |0.179             |
|100000         |1000             |1              |1                    |customers              |0.1               |0.009             |
|100000         |1000             |11             |1                    |customers              |0.405             |0.09              |
|100000         |1000             |1              |11                   |customers              |0.037             |0.039             |
|100000         |1000             |1              |1                    |customers              |0.019             |0.01              |
|100000         |1000             |1              |6                    |customers              |0.048             |0.022             |
|1000000        |1000             |1              |1                    |orders                 |0.884             |5.136             |
|1000000        |1000             |11             |1                    |orders                 |1.44              |18                |
|1000000        |1000             |1              |1                    |orders                 |0.21              |2.185             |
|1000000        |1000             |1              |6                    |orders                 |0.433             |10.182            |
|100000         |1000             |1              |1                    |orders                 |0.525             |0.227             |
|100000         |1000             |11             |1                    |orders                 |0.297             |1.709             |
|100000         |1000             |1              |1                    |orders                 |0.023             |0.211             |
|100000         |1000             |1              |6                    |orders                 |0.047             |0.882             |

## Testing

Yes. Added a unit test to show joining does not change the column names on the original table.

### DIFF
--- a/core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
+++ b/core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
@@ -63,7 +63,6 @@ public class DataFrameJoiner {
 
     /**
      * Constructor.
-     *
      * @param table The table to join on.
      * @param joinColumnNames The join column names to join on.
      */
@@ -75,9 +74,7 @@ public class DataFrameJoiner {
 
     /**
      * Finds the index of the columns corresponding to the columnNames.
-     *
      * E.G. The column named "ID" is located at index 5 in table.
-     *
      * @param table the table that contains the columns.
      * @param columnNames the column names to find indexes of.
      * @return a list of column indexes within the table.
@@ -194,12 +191,11 @@ public class DataFrameJoiner {
      * @param table1 the table on the left side of the join.
      * @param table2 the table on the right side of the join.
      * @param joinType the type of join.
-     * @param allowDuplicates if {@code false} the join will fail if any columns other than the join column * have the
-     * same name if {@code true} the join will succeed and duplicate columns are renamed*
+     * @param allowDuplicates if {@code false} the join will fail if any columns other than the join column have the
+     * same name if {@code true} the join will succeed and duplicate columns are renamed
      * @param table2JoinColumnNames The names of the columns in table2 to join on.
      */
-    private Table joinInternal(Table table1, Table table2, JoinType joinType, boolean allowDuplicates,
-        String... table2JoinColumnNames) {
+    private Table joinInternal(Table table1, Table table2, JoinType joinType, boolean allowDuplicates, String... table2JoinColumnNames) {
         List<Integer> table2JoinColumnIndexes = getJoinIndexes(table2, table2JoinColumnNames);
 
         Table result = Table.create(table1.name());
@@ -207,11 +203,8 @@ public class DataFrameJoiner {
         Set<Integer> resultIgnoreColIndexes = emptyTableFromColumns(result, table1, table2, joinType, allowDuplicates,
             table2JoinColumnIndexes);
 
-        // Build a reverse index for every join column in both tables.
-        List<Index> table1Indexes = joinColumnIndexes.stream().map(index -> indexFor(table1, index))
-            .collect(Collectors.toList());
-        List<Index> table2Indexes = table2JoinColumnIndexes.stream().map(index -> indexFor(table2, index))
-            .collect(Collectors.toList());
+        List<Index> table1Indexes = buildIndexesForJoinColumns(joinColumnIndexes, table1);
+        List<Index> table2Indexes = buildIndexesForJoinColumns(table2JoinColumnIndexes, table2);
 
         Selection table1DoneSelection = Selection.with();
         Selection table2DoneSelection = Selection.with();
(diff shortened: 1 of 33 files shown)


lwhite1:
@ryancerf That was impressively fast :). The performance numbers look great.  I'm not sure the changes to the javadoc formatting improve readability, but since the core java library code uses both approaches (indenting/not-indenting on multiline comments) I think it's ok.  

@benmccann Would you like to take a look?

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -353,144 +389,34 @@ public Table fullOuter(Table... tables) {
...
-        }
-        Table table2OnlyRows = table2.where(selection);
-        List<Column<?>> joinColumns = table2OnlyRows.columns(col2Names);
-        table2OnlyRows.removeColumns(col2Names);
-        withMissingRightJoin(result, joinColumns, table2OnlyRows);
-        return result;
+        return joinInternal(table, table2, JoinType.FULL_OUTER, false, col2Name);
     }
 
+    /*
```
benmccann: extraneous

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -668,44 +636,57 @@ private void crossProduct(Table destination, Table table1, Table table2, String[
...
+                for (int r1 = 0; r1 < table1Rows.size(); r1++) {
                     destination.column(c).appendMissing();
                 }
             }
         }
     }
 
     /**
-     * Adds rows to destination for each row in the joinColumn and table2
+     * Adds rows to destination for each row in table2, with the columns from table1 added as missing values.
```
benmccann: I don't think you need the comma in this sentance

ryancerf:
@benmccann  Thanks for the comments.

@lwhite1 Thanks. I wish I could always be that fast. I have an injury at the moment and do not have much else I can do.

Regarding the javadoc formatting. My personal preference is to use Google's Java formatter and live with whatever it does. https://github.com/google/google-java-format. I highly recommend it!

benmccann:
Nice. I didn't know that was open source.

@lwhite1 what are your thoughts about using Google's Java formatter? It might be nice to get a consistent style without us having to worry about enforcing it in code reviews. I can send a PR showing how it would format files if you want to see what it would do

lwhite1:
i just installed and uninstalled it. i don't think i want to go back to 2
space indentation. Other than that I'm sure it would be pretty good.

benmccann:
It looks like if you use `--aosp` it will format in Android style which is 4 spaces. I'm not sure if that would result in other changes we don't want. I personally prefer the 2 space style, but don't care too much one way or the other

ryancerf:
The [java format intellj plugin](https://plugins.jetbrains.com/plugin/8527-google-java-format) will respect an .editorconfig file.

Add a filled named .editorconfig at the root of the project and add the following lines to it.
```
[*.java]
indent_style = space
indent_size = 4
```

PR #563, opened 2019-08-03
added null support for SqlResultSetReader

albalkum:
Thanks for contributing.

- [x ] Tick to sign-off your agreement to the [Developer Certificate of Origin (DCO) 1.1](https://developercertificate.org)

## Description

During my use of Tablesaw, I found that when reading from a database, if a nullable double/integer column contained a null, the default value would instead be read into the Table.  After inspecting SqlResultSetReader, I saw that it did not utilize resultSet.wasNull() on the primitive type columns.  I added a separate function to check .wasNull() and utilize column.appendMissing() if it returned true.

Thanks for the great library!  It has helped my work tremendously and I enjoy utilizing it.

## Testing

I expanded SqlResultSetReaderTest.testSqlResultSetReader to include a test on another table ("NullValues" that I created in TestDb.java) that contained sporadic nulls on primitive typed columns.  I then asserted that the appropriate non-missing values existed in the columns.

### DIFF
--- a/core/src/main/java/tech/tablesaw/io/jdbc/SqlResultSetReader.java
+++ b/core/src/main/java/tech/tablesaw/io/jdbc/SqlResultSetReader.java
@@ -16,6 +16,8 @@ package tech.tablesaw.io.jdbc;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
+
+import tech.tablesaw.api.BooleanColumn;
 import tech.tablesaw.api.ColumnType;
 import tech.tablesaw.api.DoubleColumn;
 import tech.tablesaw.api.FloatColumn;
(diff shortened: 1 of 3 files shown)


lwhite1:
@albalkum Thank you for this. The changes look good and it's an important fix. Should it include BooleanColumn as well? I suspect that might default to false given the way the numerics are handled.

albalkum:
Sqlresultsetreader didn't have an else if for a booleancolumn. It's handled by resultset.getObject() which properly return nulls. Part of my addition to the test covers Boolean values and nulls are correctly processed.

albalkum:
Your comment made me realize that resultset.getObject() might be able to be used for all column cases. Wouldn't change anything but remove the if complexity. I might take a look when i get back to my computer.

albalkum:
I tried just utilizing resultSet.getObject() but there were issues when appending BigDecimal objects.  I kept the existing if else blocks and added a BooleanColumn check for clarity's sake, per your suggestion.

benmccann:
I'm not sure I'd add a case for `BooleanColumn` if it's not necessary. It's a bit confusing to me that it'd be singled out and not the other types such as `Date`, etc. I think adding a test to make sure it works properly is enough.

lwhite1:
@benmccann, It's only the primitives that have the issue because they're setting nulls in the db to , for example, 0 for int columns. They should be handled as missing. 

Booleans are apparently handled by getObject(), but if they weren't, I imagined they would default to false by analogy with the other primitives, and so have the same problem.

I believe this is probably good the way it is now.

Issue #564, opened 2019-08-04
Make it easier to work with missing data


ryancerf:
Useful for working with missing data. Similar to pandas fillna, but for a single column at a time.

Usage:
```java
myDoubleColumn.fillMissingWith(1.0);
```

I am not sure what the easiest way to do this is right now.

Will send a PR if we think this is worthwhile.

benmccann:
You should be able to do that today with `col.set(NumberPredicates.isMissing, 1.0)` . This is just off the top of my head, so the syntax might be slightly off

ryancerf:
Another method that might be useful for working with missing data is a getOrDefault method on Row.

```java
// Proposed
row.getLongOrDefault("col1", 1L);
```

It could simply iterating over a table with missing data.

```java
    // Current
    for (Row row : table) {
        long value1 = row.getLong("col1");
        int value2 = row.getInt("col2");
        value1 = LongColumn.valueIsMissing(value1) ? 1L : value1;
        value2 = IntColumn.valueIsMissing(value2) ? 1 : value2;

        // do something with value1 and value 2;
    }

    // Proposed
    for (Row row : table) {
        long value1 = row.getLongOrDefault("col1", 1L);
        int value2 = row.getIntOrDefault("col2", 1);
        // do something with value1 and value2
    }
```

lwhite1:
I have mixed feelings on whether it's worth having a new method.
As @benmccann points out, it's not too hard. 

`col.set(col.isMissing(), 1.0);`
will do it.

On the other hand, there's a reason why they made it easier in pandas. I feel like it's customary in data analysis tools and libraries to provide very good support for missing values. I would not be averse to:

`Column<?>::setMissing(? value)...
`
I wouldn't call it "fill" though, I think fill in Tablesaw is more likely to refer to a series of appends than a series of updates. Update methods generally start with set.

ryancerf:
Using set is fine with me, but I would like to point out we do have a fillWith method on DoubleColumn (I am not sure I like this method. map is a good alternative).

```java
// This method was added recently.
 @Override
    public DoubleColumn fillWith(double d) {
        for (int r = 0; r < size(); r++) {
            set(r, d);
        }
        return this;
    }
```

lwhite1:
@ryancerf You're right about fillWith. I forgot how it was implemented. 

I think the intent there was to create columns where every value is set in this method, and that it is intended for populating new columns. Although it uses set() instead of append, I think the intent is the same. 

[...]

PR #569, opened 2019-08-06
Fix for Make it easier to work with missing data #564

lwhite1:
plus test, plus fix for a bug in a DoubleColumn.create()

Thanks for contributing.

## Description

What was changed

## Testing

Did you add a unit test?

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/DoubleColumn.java
+++ b/core/src/main/java/tech/tablesaw/api/DoubleColumn.java
@@ -109,11 +109,15 @@ public class DoubleColumn extends NumberColumn<Double> implements NumberFillers<
   }
 
   public static DoubleColumn create(final String name, final List<Number> numberList) {
-    final double[] doubles = new double[numberList.size()];
-    for (int i = 0; i < numberList.size(); i++) {
-      doubles[i] = numberList.get(i).doubleValue();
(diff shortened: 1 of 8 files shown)


lwhite1:
@ryancerf LMK if this works for your missing value issue #564

core/src/main/java/tech/tablesaw/api/DoubleColumn.java
```diff
@@ -109,19 +109,27 @@ public static DoubleColumn create(final String name, final long[] arr) {
...
   public static DoubleColumn create(final String name, final List<Number> numberList) {
-    final double[] doubles = new double[numberList.size()];
-    for (int i = 0; i < numberList.size(); i++) {
-      doubles[i] = numberList.get(i).doubleValue();
+    DoubleColumn newColumn = new DoubleColumn(name, new DoubleArrayList(0));
+    for (Number number : numberList) {
+      if (number == null) {
+        newColumn.appendMissing();
+      } else {
+        newColumn.append(number.doubleValue());
```
benmccann: It might be easier to make `append` deal with `null`s. Then you wouldn't have to update multiple `create` methods. And it'd also do the right thing if the user calls `append(null)`

lwhite1: the change was put in because the original version copied the objects to primitive arrays so it would NPE, but that's a good point. Append should handle nulls

PR #579, opened 2019-08-08
Add mapToDoubleColumn on Table

ryancerf:
Thanks for contributing.

## Description

Change TableSliceGroup to iterate over Table rows.

Add MapTo[x] method to Abstract Relation class so they can be used by TableSlice and Table. 

Helps with #538 

If the implementation looks good I can update this PR to include other column types or send another PR.

## Testing

Yes

### DIFF
--- a/core/src/main/java/tech/tablesaw/table/Relation.java
+++ b/core/src/main/java/tech/tablesaw/table/Relation.java
@@ -18,7 +18,9 @@ import it.unimi.dsi.fastutil.ints.IntArrays;
 import java.io.ByteArrayOutputStream;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Iterator;
 import java.util.List;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import tech.tablesaw.api.BooleanColumn;
 import tech.tablesaw.api.CategoricalColumn;
@@ -530,4 +532,54 @@ public abstract class Relation implements Iterable<Row> {
   public boolean containsColumn(Column<?> column) {
     return columns().contains(column);
   }
+
+  /**
+   * Maps the supplied functions across all rows in the table appending the results to a new DoubleColumn.
+   *
+   * If you do not want the supplied function applied to rows with missing values you can supply a
+   * list of columns to check for missing values. For each row, it will check for a missing value in any of the
+   * listed columns. If there is a missing value a missing value will be appended to the results, otherwise
+   * the result of the function will be appended.
+   *
+   *
+   * @param name The name of the new column.
+   * @param function to map a table row to a double.
+   * @param skipIfMissingColNames the column names to check for missing values. If any of the
+   * columns contain a missing value (for a particular row) a missing value will be appended.
+   * @return DoubleColumn with the results appended.
+   */
+  public DoubleColumn mapToDoubleColumn(String name, Function<? super Row, ? extends Double> function, String... skipIfMissingColNames) {
+    return (DoubleColumn) mapToTypeInternal(DoubleColumn.create(name, rowCount()), function, skipIfMissingColNames);
+  }
+
+  private <T> Column<T> mapToTypeInternal(Column<T> destination,
+    Function<? super Row, ? extends T> function, String[] skipIfMissingColNames) {
+    List<Column<?>> checkMissingCols = Arrays.stream(skipIfMissingColNames).map(this::column).collect(Collectors.toList());
+    Iterator<Row> rowIterator = iterator();
+    int i = 0;
+    while(rowIterator.hasNext()) {
+      Row row = rowIterator.next();
+      if(isMissing(checkMissingCols, i)) {
+        destination.setMissing(i);
+      } else {
+        destination.set(i, function.apply(row));
+      }
+      i++;
+    }
+    return destination;
+  }
+
+  /**
+   * Checks every cell for the supplied list of columns and row index.
+   * If any cell contains missing data then return true;
+   */
+  private boolean isMissing(List<Column<?>> cols, int rowIndex) {
+    for (Column<?> col : cols) {
+      if (col.isMissing(rowIndex)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
 }
--- a/core/src/main/java/tech/tablesaw/table/Relation.java
+++ b/core/src/main/java/tech/tablesaw/table/Relation.java
@@ -32,6 +32,7 @@ import tech.tablesaw.api.IntColumn;
 import tech.tablesaw.api.LongColumn;
 import tech.tablesaw.api.NumberColumn;
 import tech.tablesaw.api.NumericColumn;
+import tech.tablesaw.api.Row;
 import tech.tablesaw.api.ShortColumn;
 import tech.tablesaw.api.StringColumn;
 import tech.tablesaw.api.Table;
@@ -47,7 +48,7 @@ import tech.tablesaw.sorting.comparators.DescendingIntComparator;
  * A tabular data structure like a table in a relational database, but not formally implementing the
  * relational algebra
  */
-public abstract class Relation {
+public abstract class Relation implements Iterable<Row> {
 
   public abstract Relation addColumns(Column<?>... cols);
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/tech/tablesaw/table/TableSlice.java
```diff
@@ -165,30 +167,23 @@ public double reduce(String numberColumnName, NumericAggregateFunction function)
...
 
   /**
-   * Returns a 0 based int iterator for use with, for example, get(). When it returns 0 for the
-   * first row, get will transform that to the 0th row in the selection, which may not be the 0th
-   * row in the underlying table.
+   * Iterate of a copy of the table.
    */
   @Override
-  public IntIterator iterator() {
+  public Iterator<Row> iterator() {
```
benmccann: Is it possible to move this to `Relation` so that it doesn't have to be duplicated between here and `Table`?

core/src/main/java/tech/tablesaw/table/Relation.java
```diff
@@ -529,4 +532,54 @@ public String getString(int r, int columnIndex) {
...
+   *
+   *
+   * @param name The name of the new column.
+   * @param function to map a table row to a double.
+   * @param skipIfMissingColNames the column names to check for missing values. If any of the
+   * columns contain a missing value (for a particular row) a missing value will be appended.
+   * @return DoubleColumn with the results appended.
+   */
+  public DoubleColumn mapToDoubleColumn(String name, Function<? super Row, ? extends Double> function, String... skipIfMissingColNames) {
+    return (DoubleColumn) mapToTypeInternal(DoubleColumn.create(name, rowCount()), function, skipIfMissingColNames);
```
benmccann: I wouldn't think this cast should be needed. If we're passing in a `DoubleColumn` then I think it might be able to return one

Perhaps if you do:

    private <C extends Column<T>, T> C mapToTypeInternal(C destination

core/src/main/java/tech/tablesaw/table/Relation.java
```diff
@@ -529,4 +532,54 @@ public String getString(int r, int columnIndex) {
...
+  }
+
+  private <T> Column<T> mapToTypeInternal(Column<T> destination,
+    Function<? super Row, ? extends T> function, String[] skipIfMissingColNames) {
+    List<Column<?>> checkMissingCols = Arrays.stream(skipIfMissingColNames).map(this::column).collect(Collectors.toList());
+    Iterator<Row> rowIterator = iterator();
+    int i = 0;
+    while(rowIterator.hasNext()) {
```
benmccann: did you rebase against `master`? I would think that the Google code formatter would add a space after `while`

core/src/main/java/tech/tablesaw/table/Relation.java
```diff
@@ -529,4 +532,54 @@ public String getString(int r, int columnIndex) {
...
+   * the result of the function will be appended.
+   *
+   *
+   * @param name The name of the new column.
+   * @param function to map a table row to a double.
+   * @param skipIfMissingColNames the column names to check for missing values. If any of the
+   * columns contain a missing value (for a particular row) a missing value will be appended.
+   * @return DoubleColumn with the results appended.
+   */
+  public DoubleColumn mapToDoubleColumn(String name, Function<? super Row, ? extends Double> function, String... skipIfMissingColNames) {
```
benmccann: I think it might be nice to have a method like `addDoubleColumn` that takes the same parameters, calls this method, and adds the result to the table

lwhite1:
This looks pretty good, but I have some reservations.

The main issue is that, if we were to add another column type, this is yet another place where we have to find column-specific functionality and add update it so things don't break. 

Plus, we're adding 12 new methods to table. 

FWIW, I think i do this kind of thing once every week or two of full time tablesaw use. 

In the thread, I mentioned the (perhaps poorly named) 

```rollWithRows(Consumer<Row[]>, int rowCount)```

which for our purposes could be simplified to:

```apply(Consumer<Row>)```

which i tend to use by declaring a column first, then updating it in the consumer.  I handle the missing values inside the consumer. (This could be improved as you suggested elsewhere by better missing value support in Row.).

If we want to make a proper map function, could we not have one or two methods like:
```
mapInto(Function<Row, T>, Column<T>)
mapInto(Function<Row, T>, Column<T>, String ... ColumnNamesToCheck)
```
and let the user cast them to the appropriate precise column they want?

File: core/src/main/java/tech/tablesaw/table/Relation.java (hunk 4 above)
Comment (lwhite1) [reply core/src/main/java/tech/tablesaw/table/Relation.java]: is that so much better than  
```
t.addColumns(t.mapToDoubleColumn());
```

Comment (benmccann) [reply core/src/main/java/tech/tablesaw/table/Relation.java]: Not tremendously, but I find most of the time I create a new column I'm adding it to the table. Let's hold off for now though. We can consider it separately later if it's much of an issue

File: core/src/main/java/tech/tablesaw/table/TableSlice.java (hunk 1 above)
Comment (ryancerf) [reply core/src/main/java/tech/tablesaw/table/TableSlice.java]: It is possible, but TableSice will still have to override it, which is fine. I will make the change.

**Comment by ryancerf** [comment]
Sounds good. I do not like the column type specific stuff either. When I need to do complicated mappings involving multiple columns I tend to user a regular old for loop.

```java
DoubleColumn newColumn = DoubleColumn.create("col1", table.size());
for(Row row: table) {
....
}
table.addColumns(newColumn);
```

Although the `mapInto(Function<Row, T>, Column<T>, String ... ColumnNamesToCheck)` could be useful, they feel like a bit of a compromise. I would like to see us either add methods that we are super happy about or nothing at all.

@lwhite1 you should send a PR to update the naming to `apply(Consumer<Row>)` I think that will help.

I will take the mapToType stuff out of this PR and resend it. I want to keep the other stuff. (Make relation implement Iterable<Row> etc)

A good discussion thanks for all the feedback.

lwhite1:
> I will take the mapToType stuff out of this PR and resend it. I want to keep the other stuff. (Make relation implement Iterable etc)

The stuff you've done to strengthen Relation and make TableSlice more like Table is awesome. Thanks.

>Although the mapInto(Function<Row, T>, Column<T>, String ... ColumnNamesToCheck) could be useful, they feel like a bit of a compromise. I would like to see us either add methods that we are super happy about or nothing at all.

Ok. Lets not lose the idea, though. For now, I can't think of a better way to do this.

>@lwhite1 you should send a PR to update the naming to apply(Consumer<Row>) I think that will help.

I will do that today.

Thanks again @ryancerf.

benmccann:
If we don't like the column type specific stuff on table, do we still need `mapToType` on a column (https://github.com/jtablesaw/tablesaw/pull/545) then? If we think it's too complex to add to `Table` then I'd think the same argument applies to `Column` and we should probably be consistent

benmccann:
I'm a bit confused what the plan is here. @ryancerf said he wanted to remove `mapToDouble`, but then we merged this PR. @lwhite1 was it something you wanted to keep?

ryancerf:
@benmccann 

Same here, maybe Larry wanted to make sure we included the changes to Relation and TableSlice. Ether way #583 removes everything.

lwhite1:
sorry. I thought @ryancerf <https://github.com/ryancerf> had removed
mapToDouble and left the other stuff to be merged.

core/src/test/java/tech/tablesaw/filters/TimeDependentFilteringTest.java
```diff
@@ -113,11 +114,11 @@ public static void main(String[] args) throws Exception {
       List<LocalDate> eventDates = new ArrayList<>();
 
       // iterate an individual table and find the rows where concept matches the target concept
-      for (int row : patientTable) {
+      for (Row row : patientTable) {
         StringColumn concepts = patientTable.stringColumn("concept");
         DateColumn dates = patientTable.dateColumn("date");
-        if (concepts.get(row).equals(conceptZ)) {
-          eventDates.add(dates.get(row));
+        if(row.getString("concept").equals(conceptZ)) {
```
benmccann: `concepts` and `dates` are now unused. Should they be removed?

ryancerf: I am not sure why this line was touched in this PR. I think it was my first PR. If something related to concepts and dates needs to be cleaned up can you create a separate issue?

PR #581, opened 2019-08-09
Add rolling stream method

benmccann:
Thanks for contributing.

## Description

This adds a rolling `stream(n)` method which allows us to do many different operations such as `forEach`, `map`, `filter`, `anyMatch`, `allMatch`, etc. on rolling groups of rows. `rollWithRows(Consumer<Row[]> rows, int n)` is the same as `stream(n).forEach(consumer)`, so I've deprecated it since it's the same thing, but less powerful

@ryancerf depending on which PR is merged first (i.e. this one or https://github.com/jtablesaw/tablesaw/pull/579) either you or I might need to rebase to put this method on `Relation`

## Testing

Existing unit tests cover this functionality. They're effective too because they caught a bug in my initial implementation.

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/Table.java
+++ b/core/src/main/java/tech/tablesaw/api/Table.java
@@ -1069,8 +1069,84 @@ public class Table extends Relation implements Iterable<Row> {
     };
   }
 
+  /**
+   * Iterates over rolling sets of rows. I.e. 0 to n-1, 1 to n, 2 to n+1, etc.
+   *
+   * @param n the number of rows to return for each iteration
+   */
+  public Iterator<Row[]> rollingIterator(int n) {
+
+    return new Iterator<Row[]>() {
+
+      private int currRow = 0;
+
+      @Override
+      public Row[] next() {
+        Row[] rows = new Row[n];
+        for (int i = 0; i < n; i++) {
+          rows[i] = new Row(Table.this, currRow + i);
+        }
+        currRow++;
+        return rows;
+      }
+
+      @Override
+      public boolean hasNext() {
+        return currRow + n <= rowCount();
+      }
+    };
+  }
+
+  /**
+   * Streams over stepped sets of rows. I.e. 0 to n-1, n to 2n-1, 2n to 3n-1, etc. Only returns full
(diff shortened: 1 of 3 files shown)


benmccann:
What do you think about the name? Would `rollingStream(n)` be better than `stream(n)`?

lwhite1:
> This adds a rolling stream(n) method which allows us to do many different operations such as forEach, map, filter, anyMatch, allMatch, etc. on rolling groups of rows. rollWithRows(Consumer<Row[]> rows, int n) is the same as stream(n).forEach(consumer), so I've deprecated it since it's the same thing, but less powerful

There is so much awesome stuff getting done now. It's really exciting.

> What do you think about the name? Would rollingStream(n) be better than stream(n)?

So long as you're rolling (i.e. moving forward one row at a time as in a rolling average), I think rollingStream is preferable. I often, also, find it useful to step forward by groups of n rows. We may want to support that option. Or is that built in here as well? If you can do either with this implementation I prefer stream().

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -1069,15 +1069,48 @@ public boolean hasNext() {
     };
   }
 
+  /**
+   * Iterates over rolling consecutive rows
+   *
```
lwhite1: I'm not entirely clear, but it looks like this is stepping forward (e.g. if n = 2, the first call to next() gets rows 0 and 1, and the second gets rows 2 and 3). Is that it, or is it rolling, such that the first call to next() gets rows 0 and 1, and the second gets rows 1 and 2? 

Either way could we add that kind of example to the comment so it's clear without reading the code?

ryancerf:
This looks good to me, but I would like to discuss the possibility of supporting analytic functions. See #582.  Analytic functions would provide similar rolling stream support.

core/src/main/java/tech/tablesaw/api/Table.java
benmccann: I've renamed the methods and added comments

benmccann:
I renamed to `rollingStream` and added a `steppingStream` as well

PR #583, opened 2019-08-09
Remove mapToType on Column and Table

### LINKED ISSUE #538: Column-wise DataFrame-like operations

ryancerf:
Thanks for contributing.

## Description

Remove mapToType methods as discussed. Closes  #538

## Testing

No. Removed associated unit tests.


lwhite1:
If no one objects, I will bring the column-version back. I don't see how we can do better than that, and I don't see how we could do without it. I hadn't intended for this to be removed, but i wasn't communicating clearly, so I thought I would think on more before undoing the undo. :)

Sorry for all the confusion. 

On the table version, i haven't though of anything better, but wanted to give some thought to how much might be achieved by composing column functions. It may end up coming back as well. I would like to avoid requiring the missing column list as that seems likely to be error prone. It will probably end-up being necessary tho.

ryancerf:
Overall I am indifferent to adding back the map methods on column, but I do think there is room for improvement in this area.

If we can come up with a better way to handle the multiple column case then maybe we can live out the map methods on column?

### Ideas for multi column API

This is probably not the right place to discuss this, but what do you think about an API like the one below? We could use a similar builder/fluent pattern to the one used in the analytic queries.

```java
table.columnBuilder()
  .addDateColumn(row -> row.getDate("date").plusDays(row.getNumber("daysToAdd", 1))) 
  .as("adjustedDate")
  .addDoubleColumn(row -> Math.pow(row.getNumber("curValue") / row.getNumber("bookValue"),
        1.0 / row.getNumber("nYears") - 1) * 100)
  .as("CAGR")
  .executeInPlace();
```
I am not sure I like `columnBuilder()` as the entry point, but I think we could come up with a decent name. I do like that the addColumn methods are not directly located on Table.

Lastly I am experimenting with a way for these multi-column map methods to handle missing values the same way pandas sql does. I can share more details later.

benmccann:
Eventhough I had suggested adding `mapToType` methods originally, I'd prefer to hold off on adding them back at least until we nail down what if any changes we want to make to the `Table` API. The main issue is that adding it would be inconsistent with the decision we made not to add [`mapToType` on `Table`](https://github.com/jtablesaw/tablesaw/pull/579). There were a few issues raised there that would apply to the column methods as well:
* Adds column-specific code. Makes adding new column types harder
* Adds numerous methods to the interface
* Doesn't work with custom column types

What's the benefit of a multi-column API? Is it that you only iterate over the rows once for performance? One drawback with the proposed API is that it'd have two of the three above issues

I just realized another possible improvement we could make and sent https://github.com/jtablesaw/tablesaw/pull/634. I'm reasonably happy with computing new columns as it stands. I wouldn't be against other improvements as well, but am leaning towards agreeing with [Larry's suggestion that we not add more column-specific code](https://github.com/jtablesaw/tablesaw/pull/579#issuecomment-519653706)

Issue #582, opened 2019-08-09
Support Analytic/Window functions


ryancerf:
This requires some research. This is just some initial thoughts. I would like to work on this and am happy to do the heavy lifting.

## Analytic Functions

Analytic functions are one of the most powerful features of SQL. [BigQuery has some good docs.](https://cloud.google.com/bigquery/docs/reference/standard-sql/analytic-function-concepts) The short version is that they allow you to split the data set into partitions, order each partition, define a window (relative to the current row) and then operate over that window.

## Example

```
          Sales By Region          
    date     |  region  |  sales  |
-----------------------------------
 2018-01-01  |      US  |     10  |
 2018-02-01  |      US  |     40  |
 2018-03-01  |      US  |     20  |
 2018-01-01  |  Europe  |     30  |
 2018-02-01  |  Europe  |     50  |
 2018-03-01  |  Europe  |     40  |
```

### Cumulative Sales By Region
```sql
SELECT date, region,
SUM(sales) over(PARTITION BY region ORDER By DATE ASC ROWS BETWEEN 3 PRECEDING AND CURRENT ROW) AS ytdSales
```

```
| date       | region | ytdSales |
|------------|--------|----------|
| 2019-01-01 | Europe | 30       |
| 2019-02-01 | Europe | 80       |
| 2019-03-01 | Europe | 120      |
| 2019-01-01 | US     | 10       |
| 2019-02-01 | US     | 50       |
| 2019-03-01 | US     | 70       |

```

Operate over trailing two runs. (There is a lot of flexibility in choosing this syntax)

```sql
// Sum trailing 2 rows. No partition.
SELECT date, region,
SUM(sales) OVER (ROWS BETWEEN 2 PRECEDING AND 1 PRECEDING) AS precedingWindow
```

```
| date       | region | precedingWindow |
|------------|--------|-----------------|
| 2019-01-01 | US     | null            |
| 2019-02-01 | US     | 10              |
| 2019-03-01 | US     | 50              |
| 2019-01-01 | Europe | 60              |
| 2019-02-01 | Europe | 50              |
| 2019-03-01 | Europe | 80              |
```

## Thoughts on Syntax (Almost exactly like SQL)

We create a similar functionality, but with an `accept(Consumer<Row[]> consumer)` method that adds flexibility.

#### Calculates ytd Sales
```java
 table.over()
      .partitionBy( "region")
      .orderBy("date", ASC)
[...]

PR #589, opened 2019-08-14
Improve TableSlice. Make TableSlice sortable.

ryancerf:

## Description

Made improvements to TableSlice in preparation for #582 .

* TableSlice can be sorted independently of its source table.
* Make it possible to iterate over the rows of a table slice without copying the underlying table.
* Create a new int iterator on TableSlice that makes it possible to iterate over the row numbers (indexes) in the source table. This allows users to iterate over cell's of a column without making a copy.
* Drive Row logic using TableSlice. This makes it easier to iterate over a selection of a table.

## Testing

Yes. 100% coverage of TableSlice.


core/src/main/java/tech/tablesaw/api/Row.java
```diff
@@ -9,10 +9,11 @@
 import java.util.Map;
 import java.util.TreeMap;
 import tech.tablesaw.columns.Column;
+import tech.tablesaw.table.TableSlice;
 
 public class Row implements Iterator<Row> {
 
-  private final Table table;
+  private final TableSlice tableSlice;
```
benmccann: Maybe we should use `Relation` instead?

ryancerf: We could use relation, but the obvious implementation would require adding yet another method to Table. It would be nice to avoid that if we can.

To iterate over a Selection of Rows we need to map the row number in the view to the row number on the underlying table. This PR creates a method on TableSlice to do this called `mappedRowNumber`. If we were to use Relation (and a similar implementation) it would  involve adding another method to Relation and overriding it in TableSlice.

ryancerf:
@lwhite1 Can you please review? This PR touches some critical functionality such as iterating over the rows of a table. I tried to be thorough with the tests, but let me know if you see a way I can improve them.

This PR is a prerequisite to #582, which I have started fiddling with.

lwhite1:
> Doesn't Table have a method that returns a row that would need updating for this change?
Never mind. I see this is handled. 

@ryancerf How would you feel about joining the project?

core/src/main/java/tech/tablesaw/table/TableSlice.java
```diff
@@ -147,8 +187,20 @@ public Table asTable() {
...
-  private IntIterator intIterator() {
-    return selection.iterator();
+  /**
+   * IntIterator of source table row numbers that are present in this view. This can be used to
+   * in combination with the source table to iterate over the cells of a column
+   * in a sorted order without copying the column.
+   *
+   * @return an int iterator of row numbers in the source table that are present in this view.
+   */
+  public PrimitiveIterator.OfInt sourceRowNumberIterator() {
```
benmccann: Should this be `public`? E.g. it seems like it could be `protected`

core/src/main/java/tech/tablesaw/table/TableSlice.java
```diff
@@ -184,4 +242,53 @@ public boolean hasNext() {
...
+    } else if (hasSelection()) {
+      return selection.get(rowNumber);
+    }
+    return rowNumber;
+  }
+
+  /**
+   * Sort the TableSlice independently from the underlying table.
+   */
+  public void sortOn(Sort key) {
```
benmccann: It looks like this duplicates a method in `Table`. If this method is moved to `Relation` then we don't need it in `Table` or `TableSlice`

ryancerf: I agree this is not well documented.
`Table#sortOn` returns a sorted copy of the Table. `TableSlice#sortOn` returns void and sorts the slice/view in place. I will send a PR improving the javadoc.

core/src/main/java/tech/tablesaw/table/TableSlice.java
ryancerf: Now that you mention it I think it can be private. The method it replace was private. I will send a PR.

PR #593, opened 2019-08-14
Add Count Unique Tests. 

### LINKED ISSUE #590: Unique after asStringColumn() returns one more empty row
Hi everyone,
I have noticed that using `.unique()` after `asStringColumn()` returns one additional empty row.

Version: `0.32.2`

example:
```
StringColumn col = StringColumn.create("col");
col.append("this");
col.append("this");
col.append("this");
col.append("this");
col.append("is");
col.append("is");
col.append("example");
System.out.println(col.unique().size());
System.out.println(col.unique().asStringColumn().size());
System.out.println(col.asStringColumn().unique().size());
```
output:
```
3
3
4 // should be 3
```

ryancerf:
Thanks for contributing.

## Description
Highlight difference between how StringColumn and NumberColumn handle counting missingValues
* StringColumn includes missingValues in countUnique.
* NumberColumn does not include missingValues in countUnique.

Closes #590. Contains a test showing the issue does not exist in master.
## Testing

Yes. Only tests.

### DIFF
--- a/core/src/test/java/tech/tablesaw/api/NumberColumnTest.java
+++ b/core/src/test/java/tech/tablesaw/api/NumberColumnTest.java
@@ -469,6 +469,19 @@ public class NumberColumnTest {
   }
 
   @Test
+  public void testCountUniqueWithMissing() {
+    IntColumn ints = IntColumn.create("ints");
+    ints.append(1);
+    ints.append(2);
+    ints.append(3);
+    ints.appendMissing();
+
+    assertEquals(3, ints.countUnique());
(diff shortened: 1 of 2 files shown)


benmccann:
It looks to me like Pandas includes missing values, so that might be more expected for users who are familiar with how that library works

Also, if we include missing values, then it's easy enough to drop missing rows before or after doing the `unique` that users could then get either behavior depending on what they want. If we filter it out automatically then it's less flexible for users who have a use case where they want missing values

PR #599, opened 2019-08-15
Implementation change of AbstractColumn.asStringColumn

ccleva:
Thanks for contributing.

## Description

- asStringColumn now uses delegation to the source column getUnformattedString in order to manage missing values
- Added javadoc

## Testing

- Added a couple of tests for DoubleColumn, IntColumn and DateTimeColumn with missing values

### DIFF
--- a/core/src/main/java/tech/tablesaw/columns/AbstractColumn.java
+++ b/core/src/main/java/tech/tablesaw/columns/AbstractColumn.java
@@ -58,7 +58,7 @@ public abstract class AbstractColumn<T> implements Column<T> {
   @Override
   public StringColumn asStringColumn() {
     StringColumn sc = StringColumn.create(name() + " strings");
-    for(int i = 0, size = size(); i < size; i++) {
+    for(int i = 0; i < size(); i++) {
     	sc.append(getUnformattedString(i));
     }
     return sc;
--- a/core/src/main/java/tech/tablesaw/columns/AbstractColumn.java
+++ b/core/src/main/java/tech/tablesaw/columns/AbstractColumn.java
@@ -58,8 +58,8 @@ public abstract class AbstractColumn<T> implements Column<T> {
   @Override
   public StringColumn asStringColumn() {
     StringColumn sc = StringColumn.create(name() + " strings");
-    for (T value : this) {
-      sc.append(String.valueOf(value));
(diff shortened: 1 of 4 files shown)


lwhite1:
sorry, @ccleva. The tools are there, they're just not obvious. I had to go looking myself to remember the methods. This area needs a bit of reflection at some point so it can be made more obvious.   You can do what you need by implementing the conversion method as : 

```
    StringColumn sc = StringColumn.create(name() + " strings");
    for (int i = 0; i < size(); i++) {
      if (isMissing(i)) {
        sc.appendMissing();
      } else {
        sc.append(getString(i));
      }
    }
    return sc;
  }

```
and actually, getString() alone would probably work without the missing value check because it should return an empty string for missing values, which StringColumn treats as missing.  

If you wouldn't mind revising your commit along these lines it would be appreciated.

ccleva:
Sorry @lwhite1 I don't understand what you mean: do you want me to replace getUnformattedString by getString ? This would change the behavior of asStringColumn because all getString methods use the column formatter while the current implementation use String.valueOf. getUnformattedString is doing just that but checking for missing value in addition.

Thanks for taking the time to help me understand. I will fix the Codacy check and the merge conflict in the meantime.

core/src/main/java/tech/tablesaw/columns/AbstractColumn.java
```diff
@@ -58,8 +58,8 @@ public String toString() {
   @Override
   public StringColumn asStringColumn() {
     StringColumn sc = StringColumn.create(name() + " strings");
-    for (T value : this) {
-      sc.append(String.valueOf(value));
+    for(int i = 0, size = size(); i < size; i++) {
```
lwhite1: I don't think the size = size() clause here is necessary.

ccleva:
No need to apologize @lwhite1, I hope you could finish your project on time.

I made a couple of performance tests: while there is a slight advantage in caching the size on the first call, once the JIT comes into play there is no measurable difference. I modified the PR accordingly.

Thanks

lwhite1:
Thanks @ccleva

ccleva:
@lwhite1 Thanks to you guys.

@benmccann nice catch, I will take a look

PR #602, opened 2019-08-15
Stop TableSliceGroup from making a copy of the source table while sorting.

ryancerf:

## Description
The current implementation of splitOn (which is used by summarize) makes a copy of the source table while sorting. Making unnecessary copies of tables is usually slow.

This PR.
* Implements a partitioning algorithm that does not rely on sort.
* Does not make an extra copy of the source table.
* Make the TableSlice's returned by StandardTableSliceGroup views of the original source table, not views of the sorted copy created during the implementation of TableSliceGroup. This is important for #582 

Please excuse the formatting changes. Makes it look like there are more diffs here than there really are.

## Testing

Yes. Added tests. Fixed tests that relied on the order of TableSlices in TableSliceGroup.

### DIFF
--- a/core/src/main/java/tech/tablesaw/table/StandardTableSliceGroup.java
+++ b/core/src/main/java/tech/tablesaw/table/StandardTableSliceGroup.java
@@ -16,19 +16,25 @@ package tech.tablesaw.table;
 
 import java.nio.ByteBuffer;
 import java.util.Arrays;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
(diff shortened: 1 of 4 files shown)


lwhite1:
LGTM. @benmccann would you like to take a look in case i missed something?

ryancerf:
The PR creates a map of byte array to selection using LinkedHashMap. Would something from fastutils be more appropriate? If users split on a high cardinality column the map could contain a lot of entries.

PR #604, opened 2019-08-17
Move the Sort prefix logic from Table class to Sort class.

ryancerf:

## Description

* Move Sort prefix logic (+, -) from Table to Sort, so it can be used elsewhere.
* Improve  error messages when the prefix is invalid or the column does not exist on the table.

## Testing

Yes. Improved testing of error messages.

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/Table.java
+++ b/core/src/main/java/tech/tablesaw/api/Table.java
@@ -530,7 +530,7 @@ public class Table extends Relation implements Iterable<Row> {
    * <p>if column name starts with - then sort that column descending otherwise sort ascending
    */
   public Table sortOn(String... columnNames) {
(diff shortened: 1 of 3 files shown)


ryancerf:
The new method on Sort is a static constructor. It creates a sort object, but does not sort the table.

The method is very very poorly named on my part. How about renaming it to `create(table, "col1", "col2")`

ryancerf:
I renamed to the method to create and added javadoc. I am going to merge this so I can more easily continue working #582 @Ben let me know if you want me to make any other changes. I am happy to revisit this if needed.

Since I am new to this let me know if I am not following proper etiquette. I would obviously wait to get approval from you and Larry before I merge anything major.

benmccann:
Ok. Thanks for clearing that up. I was reviewing quickly from mobile earlier

PR #610, opened 2019-08-18
Formatting

lwhite1:
Thanks for contributing.

## Description

Submitting this as a PR just to get feedback on how to avoid this in the future. My IDE is reformatting code automatically to match Google standard, but I don't understand why everything I fetch is not already in Google standard format. 

## Testing

Did you add a unit test? No. Formatting only.


benmccann:
I think the issue is that @ryancerf is only using an IDE, which is not configured with Google's code formatting. If you run `mvn compile` on the command line it will reformat, but if you're only using an IDE I suppose that can be skipped

Perhaps the better way to do it is via git pre-commit. There's another Maven plugin that's setup to do that. I have a branch to test it out, but [it's not working on my local machine](https://github.com/Cosium/maven-git-code-format/issues/22)

ryancerf:
I agree with Ben, I think this is my fault. I thought the git pre-commit hook was in place and I was using a machine that did not have java-format installed. I will start running mvn compile before I send PRs.

PR #613, opened 2019-08-18
Use HashSet for Textcolumn `isIn`

### LINKED ISSUE #611: Use a faster data structure for isIn membership check on TextColumn
`isIn` is implemented using an array list for TextColumn. Use a set or something else with O(1) lookup.

ryancerf:
Thanks for contributing.

## Description

Convert the collection of Strings input into `TextColumn.isIn` to a HashSet. This should get better performance when the input list is large. Fixes #611 

## Testing

No change in behavior.

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/TextColumn.java
+++ b/core/src/main/java/tech/tablesaw/api/TextColumn.java
@@ -15,7 +15,6 @@
 package tech.tablesaw.api;
 
 import com.google.common.base.Preconditions;
-import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import it.unimi.dsi.fastutil.ints.IntComparator;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


lwhite1:
is the same issue present in StringColumn? Why wouldn't we do the same thing there?

lwhite1:
i thought he was talking about using the map for the list of variables to
match against.

benmccann:
Here's from `StringColumn`:
```
  @Override
  public Selection isIn(String... strings) {
    return lookupTable.selectIsIn(strings);
  }

  @Override
  public Selection isIn(Collection<String> strings) {
    return lookupTable.selectIsIn(strings);
  }
```

It looks like the lookup table implementation is efficient

ryancerf:
Yes, `StringColumn` looks ok to me. You only need one of the two data structures to have O(1) lookup.

For `TextColumn` we loop through the rows in the table looking to see if there is a match in the HashSet of values to match against.

For `StringColumn` we loop through the values to match against and look in the DictionaryMap to see if there is a matching row.

PR #621, opened 2019-08-21
Clean up Analytic Functions

ryancerf:
Thanks for contributing.

## Description

* Added first pass of JavaDoc.
* Made as many of the classes/methods package private as possible.
* Added comments explaining implementation.
* Implemented executeInPlace method.
* cleaned up sliding vs append windows. Do not allow append windows to override removeLeftMost.
* Use consistent naming for referring to left and right side of a window.
* Implemented remaining basic analytic functions min, mean and count

Continues work on #582 

## Testing

Yes added and improved testing.

### DIFF
--- a/core/src/main/java/tech/tablesaw/analytic/AggregateFunction.java
+++ b/core/src/main/java/tech/tablesaw/analytic/AggregateFunction.java
@@ -20,14 +20,14 @@ interface AggregateFunction<T, R> {
   void removeLeftMost();
 
   /**
-   * Slides the right side (aka end) of the window to the right (aka end) by appending a new missing
-   * element. E.G. [(1, 2, 3), NULL, 5] -> [(1, 2, 3, NULL), 5)]
+   * Slides the right side (aka end) of the window to the right (aka end) by appending a new
+   * non-missing element. E.G. [(1, 2, 3), 4, 5] -> [(1, 2, 3, 4), 5)]
    */
   void addRightMost(T value);
 
   /**
-   * Slides the right side (aka end) of the window to the right (aka end) by appending a new
-   * non-missing element. E.G. [(1, 2, 3), 4, 5] -> [(1, 2, 3, 4), 5)]
+   * Slides the right side (aka end) of the window to the right (aka end) by appending a new missing
+   * element. E.G. [(1, 2, 3), NULL, 5] -> [(1, 2, 3, NULL), 5)]
    */
   void addRightMostMissing();
 
--- a/core/src/main/java/tech/tablesaw/analytic/AggregateFunction.java
+++ b/core/src/main/java/tech/tablesaw/analytic/AggregateFunction.java
@@ -5,8 +5,8 @@ package tech.tablesaw.analytic;
  *
  * <p>See {@link AggregateFunctions} for more background.
  *
- * <p>This class allows callers to independently slide a the left side, and right side of the
- * window. Implementers must keep track of the state of the window as it slides.
+ * <p>This class allows callers to independently slide the left side, and right side of the window.
+ * Implementers must keep track of the state of the window as it slides.
  *
  * @param <T> the type of the values in the input column (E.G. Integer).
  * @param <R> the type of values in the return column (E.G. Double).
--- a/core/src/main/java/tech/tablesaw/analytic/AggregateFunction.java
+++ b/core/src/main/java/tech/tablesaw/analytic/AggregateFunction.java
@@ -1,16 +1,45 @@
 package tech.tablesaw.analytic;
 
-public interface AggregateFunction<T, R> {
+/**
+ * Analytic Aggregate function.
+ *
+ * <p>See {@link AggregateFunctions} for more background.
+ *
+ * <p>This class allows callers to independently slide a the left side, and right side of the
+ * window. Implementers must keep track of the state of the window as it slides.
+ *
+ * @param <T> the type of the values in the input column (E.G. Integer).
+ * @param <R> the type of values in the return column (E.G. Double).
+ */
+interface AggregateFunction<T, R> {
 
+  /**
+   * Slides the left side (aka start) of the window window to the right (aka end) by removing the
+   * first element in the window. E.G. [(1, 2, 3, 4, 5)] -> [1, (2, 3, 4, 5)]
+   */
   void removeLeftMost();
 
+  /**
+   * Slides the right side (aka end) of the window to the right (aka end) by appending a new missing
+   * element. E.G. [(1, 2, 3), NULL, 5] -> [(1, 2, 3, NULL), 5)]
+   */
   void addRightMost(T value);
 
+  /**
+   * Slides the right side (aka end) of the window to the right (aka end) by appending a new
+   * non-missing element. E.G. [(1, 2, 3), 4, 5] -> [(1, 2, 3, 4), 5)]
+   */
   void addRightMostMissing();
 
+  /** Shortcut to appending multiple values to the right (aka end) of a window in one pass. */
   default void addAllRightMost(Iterable<T> newValues) {
     newValues.forEach(this::addRightMost);
   }
 
+  /**
+   * Get the aggregate value for the current window.
+   *
+   * @return the aggregate value for the current window.
+   */
   R getValue();
 }
--- a/core/src/main/java/tech/tablesaw/analytic/AggregateFunctions.java
+++ b/core/src/main/java/tech/tablesaw/analytic/AggregateFunctions.java
@@ -70,9 +70,9 @@ enum AggregateFunctions implements FunctionMetaData {
   private abstract static class WindowDependentAggregateFunction<T extends Number> {
 
     /** Sub classes of append windows should never call removeLeft. */
-    private abstract static class AppendAggregateFunction<T, R> implements AggregateFunction<T, R> {
+    abstract static class AppendAggregateFunction<T, R> implements AggregateFunction<T, R> {
       @Override
-      public void removeLeftMost() {
+      public final void removeLeftMost() {
         throw new UnsupportedOperationException(
             "Implementers of append aggregate functions should never call removeLeftMost");
       }
--- a/core/src/main/java/tech/tablesaw/analytic/AggregateFunctions.java
+++ b/core/src/main/java/tech/tablesaw/analytic/AggregateFunctions.java
@@ -221,15 +221,61 @@ enum AggregateFunctions implements FunctionMetaData {
   }
 
   static class Min<T extends Number> extends WindowDependentAggregateFunction<T> {
-
     @Override
     AppendAggregateFunction<T, Double> functionForAppendWindows() {
-      throw new UnsupportedOperationException("Analytic Function Min Is Not implemented");
+      return new AppendAggregateFunction<T, Double>() {
+        private Double min = Double.NaN;
+
+        @Override
+        public void addRightMost(T newValue) {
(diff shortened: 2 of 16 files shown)


core/src/main/java/tech/tablesaw/analytic/AggregateFunction.java
```diff
@@ -1,16 +1,45 @@
 package tech.tablesaw.analytic;
 
-public interface AggregateFunction<T, R> {
+/**
+ * Analytic Aggregate function.
+ *
+ * <p>See {@link AggregateFunctions} for more background.
+ *
+ * <p>This class allows callers to independently slide a the left side, and right side of the
```
lwhite1: typo

core/src/main/java/tech/tablesaw/analytic/AggregateFunction.java
```diff
@@ -1,16 +1,45 @@
...
+  /**
+   * Slides the left side (aka start) of the window window to the right (aka end) by removing the
+   * first element in the window. E.G. [(1, 2, 3, 4, 5)] -> [1, (2, 3, 4, 5)]
+   */
   void removeLeftMost();
 
+  /**
+   * Slides the right side (aka end) of the window to the right (aka end) by appending a new missing
+   * element. E.G. [(1, 2, 3), NULL, 5] -> [(1, 2, 3, NULL), 5)]
+   */
```
lwhite1: this comment and the one below seem to be reversed.

core/src/main/java/tech/tablesaw/analytic/AggregateFunctions.java
```diff
@@ -213,41 +233,188 @@ public Double getValue() {
...
-    AggregateFunction<T, Double> fixedFunction() {
-      throw new UnsupportedOperationException("Analytic Function Min Is Not implemented");
+    AppendAggregateFunction<T, Double> functionForAppendWindows() {
+      return new AppendAggregateFunction<T, Double>() {
+        private Double min = Double.NaN;
+
+        @Override
+        public void addRightMost(T newValue) {
+          if (Double.isNaN(min)) {
+            min = newValue.doubleValue();
```
lwhite1: It's an infrequent case, so it doesn't necessarily need to be dealt with here, but any comparable type could support min.

core/src/main/java/tech/tablesaw/analytic/AggregateFunctions.java
```diff
@@ -213,41 +233,188 @@ public Double getValue() {
...
+        @Override
+        public void addRightMost(T newValue) {
+          Double doubleValue = newValue.doubleValue();
+          this.sum += doubleValue;
+          queue.add(doubleValue);
+        }
+
+        @Override
+        public void addRightMostMissing() {
+          queue.add(Double.NaN);
```
lwhite1: should we use the missingValueIndicator functions in case an implementation changes

core/src/main/java/tech/tablesaw/analytic/AggregateFunctions.java
```diff
@@ -213,41 +233,188 @@ public Double getValue() {
...
+      };
     }
 
     @Override
-    AggregateFunction<T, Double> slidingFunction() {
-      throw new UnsupportedOperationException("Analytic Function Count Is Not implemented");
+    AggregateFunction<T, Integer> functionForSlidingWindows() {
+      return new AggregateFunction<T, Integer>() {
+        // True if is missing value;
+        private final ArrayDeque<Boolean> queue = new ArrayDeque<>();
```
lwhite1: i don't understand the comment above

core/src/main/java/tech/tablesaw/analytic/AggregateFunctions.java
```diff
@@ -213,41 +233,188 @@ public Double getValue() {
...
+        public void removeLeftMost() {
+          Boolean removedMissingValue = queue.remove();
+          if (removedMissingValue) {
+            missingCount--;
+          }
+        }
+
+        @Override
+        public void addRightMost(T newValue) {
+          queue.add(false);
```
lwhite1: i guess i'm starting to understand. :) but this is a public method, so how do we know that newValue is not missing?

core/src/main/java/tech/tablesaw/analytic/AnalyticQuery.java
```diff
@@ -46,60 +46,102 @@ private AnalyticQuery(
...
   public static FullAnalyticQuerySteps.FromStep query() {
     return new FullQueryBuilder();
   }
 
   /**
-   * Entry point for Numbering Query.
+   * Entry point for the fluent Numbering Query Builder.
+   *
+   * <p>A numbering assigns integer values to each row based on their position within the specified
+   * partition. Numbering queries require Order By.
```
lwhite1: is Numbering different from Rank?  Asking for a friend.

core/src/main/java/tech/tablesaw/analytic/AnalyticQuery.java
```diff
@@ -129,44 +171,74 @@ public String toSqlString() {
...
+   * Executes the query and adds all the calculated columns directly to the source table.
+   *
+   * @throws IllegalArgumentException if any of the calculated columns have the same name as one of
+   *     the columns in the FROM table.
+   */
+  public void executeInPlace() {
+    // Throw an exception if there will be duplicate columns before executing the query.
+    Set<String> existingColumnNames =
+        this.table.columnNames().stream().map(String::toLowerCase).collect(Collectors.toSet());
+    Set<String> newColumnNames =
```
lwhite1: nit, i would leave off the _this._ where it isn't needed.

core/src/main/java/tech/tablesaw/analytic/AnalyticQuery.java
```diff
@@ -211,17 +283,17 @@ public Table execute() {
 
     @Override
     public void executeInPlace() {
-      throw new UnsupportedOperationException();
+      this.build().executeInPlace();
     }
   }
```
lwhite1: Leaving off here for the night. Will come back to this tomorrow.

core/src/main/java/tech/tablesaw/analytic/AggregateFunction.java
ryancerf: It is ok. This is adding the NULL element on the right side of the window.

core/src/main/java/tech/tablesaw/analytic/AggregateFunctions.java
ryancerf: Yes. That should be an easy fix.

core/src/main/java/tech/tablesaw/analytic/AnalyticQuery.java
ryancerf: Rank is a type of numbering function. This implementation includes the numbering functions row number, rank and dense rank.

core/src/main/java/tech/tablesaw/analytic/AggregateFunctions.java
ryancerf: This method has a public modifier because it overrides the interface, but the interface is package private and a user will never see these implementation classes or methods.

core/src/main/java/tech/tablesaw/analytic/AggregateFunctions.java
ryancerf: Good point. Done

core/src/main/java/tech/tablesaw/analytic/AnalyticQueryEngine.java
```diff
@@ -53,15 +68,19 @@ private void processAggregateFunctions(TableSlice slice) {
...
+  /**
+   * Execute all numbering functions for the given slice setting values in the appropriate
+   * destination column.
+   */
   @SuppressWarnings({"unchecked", "rawtypes"})
   private void processNumberingFunctions(TableSlice slice) {
     for (String toColumn : query.getArgumentList().getNumberingFunctions().keySet()) {
       if (rowComparator == null) {
-        throw new IllegalArgumentException("Cannot use Numbering Function without" + " OrderBy");
+        throw new IllegalArgumentException("Cannot use Numbering Function without " + "OrderBy");
```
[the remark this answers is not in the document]
ryancerf: fixed

core/src/main/java/tech/tablesaw/analytic/AnalyticQuerySteps.java
```diff
@@ -5,42 +5,104 @@
...
 
   interface FullAnalyticQuerySteps {
-    interface FromStep extends From<FullAnalyticQuerySteps.PartitionStep> {}
 
-    interface PartitionStep extends Partition<FullAnalyticQuerySteps.OrderOptionalStep> {}
-
-    interface OrderOptionalStep extends OrderOptional<DefineWindow> {}
+    interface FromStep {
+      /**
+       * Set the From Table name.
```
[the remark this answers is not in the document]
ryancerf: It is the From/source table similar to SQL.  It is my_table in `SELECT * FROM my_table`

I improved the comment.

core/src/main/java/tech/tablesaw/analytic/AnalyticQuerySteps.java
```diff
@@ -5,42 +5,104 @@
...
-    interface FromStep extends From<NumberingQuerySteps.PartitionStep> {}
-
-    interface PartitionStep extends Partition<NumberingQuerySteps.OrderRequiredStep> {}
 
-    interface OrderRequiredStep
-        extends AnalyticQuerySteps.OrderRequiredStep<AddNumberingFunction> {}
+    interface FromStep {
+      /**
+       * Set the From Table name.
+       *
```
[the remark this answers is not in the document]
ryancerf: improved the comment here too.

core/src/main/java/tech/tablesaw/analytic/AggregateFunctions.java
ryancerf: Updated comment.

As the window slides and we remove items from the left side we need to know if they are missing or not to maintain a correct count.

core/src/main/java/tech/tablesaw/analytic/AggregateFunction.java
lwhite1: Sorry. I still don't understand how a method called:

_addRIghtMost(T value)_ can be usefully described as "... appending a new missing element"

while

_addRightMostMissing()_ can be described as "...appending a new non-missing element"

The method names state the opposite of the comments. Also, if you're always adding NULL in the first method, why does it have a parameter at all? The implication of addRightMost(T value) is that you're adding some value of type T, rather than null.

core/src/main/java/tech/tablesaw/analytic/AnalyticQuery.java
ryancerf: done

core/src/main/java/tech/tablesaw/analytic/AggregateFunction.java
ryancerf: You are absolutely right. I do not know how I missed that.

ryancerf:
Thanks for the review.

core/src/main/java/tech/tablesaw/analytic/AggregateFunction.java
lwhite1: thats fine. glad we got it sorted.

PR #624, opened 2019-08-22
Provide more helpful error messages when a column cannot be found on Row

### LINKED ISSUE #617: Create Better Column Not Found Error Message for Row
Ran into this today while using the library.

`row.getLong("colDoesNotExist")` throws an NPE. We should create a better error message. Possibly create a custom ColumnNotFoundException (unchecked) and use it project wide.

Current error message:

```
Exception in thread "main" java.lang.NullPointerException
	at tech.tablesaw.api.Row.getLong(Row.java:153)
	at com.ryancerf.tgo.steps.CreateInvoicesStep.createInvoices(CreateInvoicesStep.java:23)
	at com.ryancerf.tgo.Main.main(Main.java:40)
```

This error also occurs when the column name exists on the table, but the type is wrong. It would be nice if we searched the possible column names and gave the user the hint that they probably got the wrong `row.get` method.

If no one has grabbed this I should have time to work on it in a few days.

ryancerf:
Thanks for contributing.

## Description

Fixes #617

New Error Msgs look like:

For missing column:
```
Column col2 is not present in table myTable
```

For using wrong type.
```
Column col1 is of type LOCAL_DATE and cannot be cast to LOCAL_TIME. Use the method for LOCAL_DATE.
```

## Testing

Yes. Added tests.

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/Row.java
+++ b/core/src/main/java/tech/tablesaw/api/Row.java
@@ -4,39 +4,90 @@ import java.time.Instant;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.LocalTime;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.TreeMap;
 import tech.tablesaw.columns.Column;
 import tech.tablesaw.table.TableSlice;
 
 public class Row implements Iterator<Row> {
 
+  /**
+   * Wrap Map of columnname to Column ap and provide helpful error messages to the user when a
+   * column name cannot be found.
+   */
+  private class ColumnMap<T> {
+    private final Map<String, T> columnMap = new HashMap<>();
+    private final ColumnType columnType;
+
+    public ColumnMap(ColumnType columnType) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/tech/tablesaw/api/Row.java
```diff
@@ -4,39 +4,90 @@
...
 import tech.tablesaw.columns.Column;
 import tech.tablesaw.table.TableSlice;
 
 public class Row implements Iterator<Row> {
 
+  /**
+   * Wrap Map of column name to Column map and provide helpful error messages to the user when a
+   * column name cannot be found.
+   */
+  private class ColumnMap<T> {
```
benmccann: make this static?

ryancerf: In order to provide the context in the error messages ColumnMap needs access to the table and list of columns. If we wanted to make it static we could create all the ColumnMaps in Row's constructor, but it is more concise the way it is. Happy to change it if you like.

As far as inner classes go I think they are fine as long as they are private and no object outside the containing class can hold a reference to the inner class.

PR #633, opened 2019-08-25
Undo deprecation missing value methods

lwhite1:
Thanks for contributing.

## Description

What was changed

## Testing

Did you add a unit test?


benmccann:
I feel like it's a bit confusing that one method is `valueIsMissing` and the other is `isMissingValue`. Perhaps they should both be called `isMissingValue`?

PR #634, opened 2019-08-25
Column creation from streams

benmccann:
Thanks for contributing.

## Description

I use Java 8 streams a lot. This makes column creation a bit easier


benmccann:
The reason I didn't add support for floats and shorts is that the Java Stream API only supports map to object, int, long, and double. I suppose for float I could take a `DoubleStream` and for short I could take an `IntStream`, but I wasn't sure if it was worth adding that or not

lwhite1:
Oh right. That was kinda half-assed of them. I think it would be better to
either use larger types, or object types (e.g. Stream<Float>) than to leave
holes in our API.

benmccann:
Ok. I've updated it to use `Stream<Float>`, etc.

Issue #457, opened 2019-02-01
Remove code from documentation and link to example classes


lwhite1:
Code in docs is impossible to maintain.

benmccann:
Maybe we could turn each page into a Jupyter notebook.

Here's an example. You can edit it and run it online by clicking "Execute on Binder" from nbviewer. It does take awhile to launch though
https://nbviewer.jupyter.org/github/benmccann/nbviewer-test/blob/master/Tablesaw.ipynb

One caveat with the example - I'm not sure how to get images and tables to properly render in nbviewer yet. I believe it's supported, but haven't been able to investigate

It would give us the following advantages:
- Viewable online with output. User doesn't need to download and run to see what it looks like (assuming we can get the images fixed)
- Supports markdown. Will help make migration easier and allow for rich descriptions
- Can run online to ensure code works.
  - We would need to update periodically to use latest version of Tablesaw. It's relatively easy to identify breakages when upgrading because the code is runnable unlike our current markdown files
  - It's always clear to the user what version is being demonstrated. It may sometimes be on an old version if we haven't updated it, but it shouldn't ever be broken.

benmccann:
I figured out that you have to press "Widgets" > "Save Notebook Widget State" to have it save with the images and tables. The link above works pretty well now!

benmccann:
You can also embed in sphinx as an alternative to nbviewer: https://nbsphinx.readthedocs.io/en/0.4.2/code-cells.html

benmccann:
Something like [Paradox](https://developer.lightbend.com/docs/paradox/current/) which lets you embed Java in Markdown would be another alternative to ensure the docs compile

lwhite1:
Paradox looks great. If the markdown is referencing code in a file that gets compiled and tested regularly, perhaps we're not doomed after all :)

ryancerf:
I put in a half hearted effort trying to get paradox and sbt to work with maven and gave up. 

I came to the conclusion that I would much rather write my own maven plugin that has similar snippet inclusion functionality rather than fiddle with build tools.

I got a proof of concept working pretty quickly and I think it could get the job done (I might also use it for other projects). It copies the docs directory to target and resolves snippets located in source files. It is almost exactly the same as Paradox's snippet inclusion feature, but that is all it does.

Let me know if you are interested and I can try to figure out how to upload it to maven so we could use it.

benmccann:
[...]

PR #637, opened 2019-08-25
Compile Tutorial docs.

ryancerf:

## Description

This not fully baked, but getting there. #457

* Creates a docs module (so it can contain compiled code)
* Add a plugin with similar functionality to paradox to inject snippets into markdown. 
* Adds an add hoc class that saves the output of the compiled snippets so they can be used in the docs. (E.G. the text output of table.first(10))
* Rewrites the tornado tutorial using using the new tools. (It was not even close to compiling)

For more information:

* See the README of https://github.com/ryancerf/choss-maven-plugin

* Read the README in the docs subdirectory.

## Testing

The plugin is well tested.


docs/README.md
```diff
@@ -1,11 +1,29 @@
-# Tablesaw Documentation
+## Compiled Docs
```
benmccann: docs/README.md is what's rendered at https://jtablesaw.github.io/tablesaw/  Perhaps it'd be possible to reconfigure it, but I'm not sure. Maybe it'd be better to add this to docs/contributing.md?

benmccann: Nevermind. This was the first file I looked at so I hadn't noticed that you were moving everything

ryancerf: Since the docs are now being built what we would actually deploy is in the `docs/dist/` directory. We would need to change the root directory for the docs to from `docs/` to `docs/dist`

The reason you see a diff here is because I moved all our old markdown files to `docs/main/`  and used the docs/README to explain how to build the docs.

docs/src/main/java/tech/tablesaw/docs/Main.java
```diff
@@ -0,0 +1,21 @@
+package tech.tablesaw.docs;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+
+/** Main class that will run all the docs. All Docs classes have to be registered in main. */
+public class Main {
+
+  /** */
```
benmccann: remove empty comment

docs/src/main/java/tech/tablesaw/docs/OutputWriter.java
```diff
@@ -0,0 +1,65 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Saves string snippets to an output file so they can be used in the docs.
+ *
+ * <p>// TODO clean this up. It is a bit of a hack.
```
benmccann: probably don't need the `//` since it's already in a comment

docs/src/test/java/tech/tablesaw/docs/TutorialTest.java
```diff
@@ -0,0 +1,8 @@
+package tech.tablesaw.docs;
+
+public class TutorialTest {
```
benmccann: Maybe leave this file off for now? It doesn't look like it's needed

benmccann:
This looks pretty cool!

ryancerf:
> This looks pretty cool!

Thanks for the idea!

lwhite1:
No problem at all. This is a big win

PR #648, opened 2019-08-28
Stage the built docs so they can be released explicitly

ryancerf:

## Description

Cleaning up the way we build and deploy the docs after fixing yesterdays breakage in a hurry.

This PR builds the docs to a staging location (`docs-src/dist/`) and creates a deploy script so they can be copied to `docs/` explicitly
* Having a staging location for the built docs allows us to choose when to release them.
We can now work on the docs for a new feature and then release them at the same time the
feature is released.


ryancerf:
Good idea. I created #652 to track preparing the docs for a release.

Issue #638, opened 2019-08-26
Better handling of file encodings


lwhite1:
We need to be able to specify an encoding/charset when opening a CSV or other text file

We _should_ support encoding detection via ICU4J (or better, if there is something better)

benmccann:
You can specify. Pass in an `InputStreamReader` which allows specifying the charset:

https://docs.oracle.com/javase/7/docs/api/java/io/InputStreamReader.html#InputStreamReader(java.io.InputStream,%20java.nio.charset.Charset)

lwhite1:
yes, but the process of figuring out what was wrong, what the correct
encoding was, and how to make Tablesaw do the right thing- took me an hour.
To learn how to change the encoding I read CsvReadOptions, then
ReadOptions, then grepped the code. That led me to the Source class, which
led me to the Builders with use streams (It doesn't say why streams have a
source, but files don't. Then I created a reader on the file and passed in
the encoding (which I figured out using ICU4J) and, *voilà, *I had the file
loaded with accents instead of question marks.

Numbers, Excel, and Pandas just did it correctly on the first try.

lwhite1:
The implementation in #638 handles the most common case: using either of these

```
Table.read().csv("filename") 
// or 
Table.read().csv(file)
```

PR #653, opened 2019-09-03
Auto detect text file charsets. Partial implementation of #638

lwhite1:
Thanks for contributing.

## Description

Detects encodings for input text files when possible, and when no encoding is specifically provided as an argument. If the detection confidence is weak, it falls back on the system default charset.

This implementation adds a new dependency to Core, which is undesirable, but getting data into tablesaw correctly is arguably its most important function so I think it's worth it.

## Testing

Did you add a unit test?
No but there are numerous tests that exercise this functionality


aggregate/pom.xml
```diff
@@ -5,7 +5,7 @@
   <parent>
     <groupId>tech.tablesaw</groupId>
     <artifactId>tablesaw-parent</artifactId>
-    <version>0.34.3-SNAPSHOT</version>
+    <version>0.35.1-SNAPSHOT</version>
```
benmccann: This file shouldn't be edited. I'm afraid you probably pulled while I was trying to cut a release earlier today, which failed and was rolled back

lwhite1: ah  didn't realize it was rolled back.  I will unwind this tomorrow.

benmccann: Unfortunately Maven is a bit frustrating in this regard. It'll commit the version bump and push it to Github early in the release process. If it fails anywhere in the process it's already pushed the commit despite the release failing

benmccann: It looks like I got the release out (it takes an hour to show up, so I'll verify later). Unfortunately I had to do a force push so this PR is still showing a diff. Not sure the best way to solve. You might have to open a new PR?

lwhite1: no problem. Thanks for getting it cleared-up.

lwhite1:
replaced with a cleaner PR

PR #654, opened 2019-09-03
partial implementation of #638

lwhite1:
Detects encodings for input text files when possible, and when no encoding is specifically provided as an argument. If the detection confidence is weak, it falls back on the system default charset.

This implementation adds a new dependency to Core, which is undesirable, but getting data into tablesaw correctly is arguably its most important function so I think it's worth it.

Thanks for contributing.

## Description

See above

## Testing

Did you add a unit test?
No but there are numerous tests that exercise this functionality

### DIFF
--- a/core/src/main/java/tech/tablesaw/io/Source.java
+++ b/core/src/main/java/tech/tablesaw/io/Source.java
@@ -1,5 +1,7 @@
 package tech.tablesaw.io;
 
+import com.ibm.icu.text.CharsetDetector;
+import com.ibm.icu.text.CharsetMatch;
 import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.FileInputStream;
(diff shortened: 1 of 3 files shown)


core/src/main/java/tech/tablesaw/io/Source.java
```diff
@@ -88,10 +89,46 @@ public Reader createReader(byte[] cachedBytes) throws IOException {
...
-      scanner.useDelimiter("\\A");
+    try (Scanner scanner = new Scanner(new URL(url).openStream())) {
+      scanner.useDelimiter("\\A"); // start of a string
       return scanner.hasNext() ? scanner.next() : "";
     }
   }
+
+  /**
+   * Returns the likely charset for the given file, if it can be determined. A confidence score is
+   * calculated. If the score is 60 or lower (on a 1 to 100 interval) the system default charset is
```
benmccann: It looks like "60 or lower" should be "less than 60" based on the code

benmccann:
Looks okay to me. Should we make the example into a test so that we don't regress on the new detection functionality in the future?

lwhite1:
Thanks, Ben. I'll fix the doc text and add a test in a separate PR.

Issue #658, opened 2019-09-04
ArrayIndexOutOfBoundsException when reading CSV file


aecio:
This bug is similar to #297 but it is not the same. This happens, when the CSV has a column header with an empty space, eg: `c1,"c2 ","c3"` (note the space in `"c2 "` header).

While reading the file, the method [`selectColumnNames()`](https://github.com/jtablesaw/tablesaw/blob/e0370f73904d3aee92d85a643a748031ad421969/core/src/main/java/tech/tablesaw/io/FileReader.java#L189) (in line `String[] columnNames = selectColumnNames(headerRow, types)`) returns trimmed strings which are then used to search over the original names of the columns in line [columnIndexes[i] = headerRow.indexOf(columnNames[i])](https://github.com/jtablesaw/tablesaw/blob/e0370f73904d3aee92d85a643a748031ad421969/core/src/main/java/tech/tablesaw/io/FileReader.java#L107). Thus, it does not find the correct column index and returns -1, which ultimately causes an index out of bounds exception.

I don't think header names should be trimmed in this case where column headers are delimited by quotes.

```
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: -1
	at java.util.ArrayList.elementData(ArrayList.java:422)
	at java.util.ArrayList.get(ArrayList.java:435)
	at tech.tablesaw.io.AddCellToColumnException.<init>(AddCellToColumnException.java:63)
	at tech.tablesaw.io.FileReader.addRows(FileReader.java:143)
	at tech.tablesaw.io.FileReader.parseRows(FileReader.java:104)
	at tech.tablesaw.io.csv.CsvReader.read(CsvReader.java:89)
	at tech.tablesaw.io.csv.CsvReader.read(CsvReader.java:78)
	at tech.tablesaw.io.DataFrameReader.csv(DataFrameReader.java:156)
	at tech.tablesaw.io.DataFrameReader.csv(DataFrameReader.java:152)
```

Here is a dataset that can be used to reproduce the error: https://finances.worldbank.org/Projects/2017-Climate-Investment-Funds-Scaling-Up-Renewable/vq9a-4dmu
Direct CSV link: https://finances.worldbank.org/api/views/vq9a-4dmu/rows.csv?accessType=DOWNLOAD

lwhite1:
Thanks for the bug report and the file to use when re-creating it. I will
work on this tomorrow.

Given that most of tablesaw relies on referencing columns by name, i think
[...]

PR #659, opened 2019-09-05
fixes ArrayIndexOutOfBoundsException when reading CSV file #658

lwhite1:
trims headers consistently.

Thanks for contributing.

## Description

Trim headers consistently

## Testing

Did you add a unit test? Yes


benmccann:
@lwhite1 Travis is failing

core/src/main/java/tech/tablesaw/io/FileReader.java
```diff
@@ -62,6 +62,7 @@ private String cleanName(String name) {
       String[] headerNames = parser.parseNext();
       // work around issue where Univocity returns null if a column has no header.
       for (int i = 0; i < headerNames.length; i++) {
+        headerNames[i] = headerNames[i].trim();
```
benmccann: It looks like this could cause an NPE since the next line checks if `headerNames[i] == null`

lwhite1: i don't think it's possible

> @lwhite1 Travis is failing

It's the log-length thing again, which I've "fixed" several times by suppressing various kinds of output. I guess we're limited to somewhere 3000-3500 lines of output now or some such. This time the extra output is from issues related to documentation. 

With the sub-projects it seems we will run into this a lot. I'd like to suppress all logging except errors. Is that possible?

benmccann: I just pushed a fix

PR #661, opened 2019-09-10
Fix saw issue

lwhite1:
Thanks for contributing.

## Description

What was changed

## Testing

Did you add a unit test?


saw/src/test/java/tech/tablesaw/io/saw/SawStorageTest.java
```diff
@@ -89,6 +93,19 @@ void testWriteTable() {
     t.sortOn("string"); // exercise the column a bit
   }
 
+  @Test
+  void testWriteTable2() {
```
benmccann: Maybe we could use a more descriptive name like `testWriteBooleanColumn`? (Actually, I've gotten in the habit of leaving `test` off the method name since it seems redundant with the already present `@Test`, so `writeBooleanColumn` is another option). And maybe we should rename `testWriteTable` to `testWriteStringColumn`?

Issue #663, opened 2019-09-17
Backfilling data


benmccann:
I'd like a way to backfill data. Is there any reasonably straightforward way to do this today? If not, I would be happy to add such a method

Use case: I have a column where only every third value is present (quarterly data in a monthly table). I'd like to backfill this column to just use the last value that was present

We'd gotten a previous request for this as well: https://github.com/jtablesaw/tablesaw/issues/401

benmccann:
It looks like Pandas calls these methods `backfill` and `pad` or `bfill` and `ffill`. I like that `bfill` and `ffill` are consistent with each other in terms of naming, but prefer more explanatory names, so would propose the methods to be called `backfill` and `frontfill`

lwhite1:
I think this is a common need in time series especially.  Most libraries in
that space support different methods of interpolation.  It would be nice if
we could put this behind an interface that let us swap in other approaches

benmccann:
> It would be nice if we could put this behind an interface that let us swap in other approaches

Can you clarify?

lwhite1:
I’m on my phone now, but this covers your approach and a couple simple
interpolation algos

lwhite1:
https://towardsdatascience.com/how-to-interpolate-time-series-data-in-apache-spark-and-python-pandas-part-1-pandas-cff54d76a2ea

benmccann:
Are you suggesting that instead of putting methods like `backfill` and `frontfill` directly on the column we group these methods together in another class? I.e. make the API look more like `col.interpolate().backfill()`, `col.interpolate().frontfill()`, `col.interpolate().linear()`,  `col.interpolate().nearest()`, etc?

Issue #401, opened 2018-10-16
 Anything can be done to fill missing value with backfill method or frontfill method?


anphy:
I'm dealing with a bunch of time serial tables with many missing values to be filled with backfill(backward) method like pandas in python, but I can only find method of Class Column to fill them with default value. Any suggestion about this?

lwhite1:
I don’t understand what you mean by forward or backward method. Can you
 explain what you want without comparing to pandas?

anphy:
@lwhite1 
Thank you for your reply, and apologize for my not being clear. I reedited it and make it clearer as follows: backfill(backward) means filling  missing/empty values (plural means many) with the next valid value, and on the contrary,  frontfill(forward) means filling them with the last valid value.

anphy:
I've made a way to do fronfill method, would anyone make a check on it?
```
            String nanStr = String.valueOf(Double.NaN);
            for (Column column : other.columns()) {
                if (column.countMissing() == 0) {
                    continue;
                }
                int row = -1;
                for (Object item : column) {
                    String i = item.toString();
                    if (i.isEmpty() || i == nanStr) {
                        column.set(row + 1, column.get(row));
                    }
                    row++;
                }
            }
```

benmccann:
@anphy I wanted to let you know that we just added this feature and it will be included in the next release

PR #664, opened 2019-09-17
Add interpolation methods

benmccann:
Thanks for contributing.

## Description

Closes https://github.com/jtablesaw/tablesaw/issues/663 https://github.com/jtablesaw/tablesaw/issues/401

## Testing

Added unit tests

### DIFF
--- a/core/src/main/java/tech/tablesaw/interpolation/Interpolator.java
+++ b/core/src/main/java/tech/tablesaw/interpolation/Interpolator.java
@@ -1,0 +1,48 @@
+package tech.tablesaw.interpolation;
+
+import tech.tablesaw.columns.Column;
+
+/**
+ * Creates a new column with missing cells filled based off the value of nearby cells. This class
+ * contains methods that are agnostic of column type.
+ */
+public class Interpolator<T> {
+
+  protected final Column<T> col;
+
+  public Interpolator(Column<T> column) {
+    this.col = column;
+  }
+
+  /** Fills missing values with the next non-missing value */
+  public Column<T> backfill() {
+    Column<T> result = col.copy();
+    T lastVal = null;
+    for (int i = col.size() - 1; i >= 0; i--) {
+      if (col.isMissing(i)) {
+        if (lastVal != null) {
+          result.set(i, lastVal);
+        }
+      } else {
+        lastVal = col.get(i);
+      }
+    }
+    return result;
+  }
+
+  /** Fills missing values with the last non-missing value */
+  public Column<T> frontfill() {
+    Column<T> result = col.copy();
+    T lastVal = null;
+    for (int i = 0; i < col.size(); i++) {
+      if (col.isMissing(i)) {
+        if (lastVal != null) {
+          result.set(i, lastVal);
+        }
+      } else {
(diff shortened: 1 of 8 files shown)


core/src/main/java/tech/tablesaw/interpolation/Interpolator.java
```diff
@@ -0,0 +1,48 @@
...
+
+  public Interpolator(Column<T> column) {
+    this.col = column;
+  }
+
+  /**
+   * Fills missing values with the next non-missing value
+   */
+  public void backfill() {
+    T lastVal = null;
```
lwhite1: couldn't this have been written by sharing the logic from frontFill, but with the iteration reversed?

lwhite1: In general, I think this implementation might have been better written as methods on AbstractColumn. Implementing it as a Interpolation class might make it more difficult to add advanced interpolation methods (like linear, mean, etc) that assume continuous variables (or at least ordered variables). I don't know how we could add linear interpolation to a string column, for example.

It was hard for me to communicate the other day because I was in an Uber, but I was thinking of an abstract interpolate() method that took a function argument, perhaps with these two implementations provided by default as named methods that use the generic interpolate.  It's still just a thought at this point as I haven't considered all the details. 

What do you think?

benmccann: yeah, the logic is the same except for reversed iteration order, but I'm not sure how we'd share more code?

benmccann: @lwhite1 I added a linear interpolation so you could see what it would look like with this type of implementation. I think the benefit of this way is that methods are nicely grouped together and you don't need to import any other class

benmccann:
Please wait to merge this PR. I'd like to rebase this one on https://github.com/jtablesaw/tablesaw/pull/669 assuming that one is merged

benmccann:
One big question I have with this PR is should we create copies of the columns or do the interpolations in place. Doing it in place would be more performant and users could call `copy` themselves if they wish. However, I ended up making a copy because it made it easy to support interpolations on `NumberColumn` because I can create the copy as a `DoubleColumn`. You can't really do a linear interpolation in place on an `IntColumn`, for example, with the current code since it produces decimal interpolations unless we convert to DoubleColumn in the process

core/src/main/java/tech/tablesaw/interpolation/Interpolator.java
lwhite1: I still don't see why it should be a separate class. It will be harder to find

benmccann: The hope was that it'd make things easier to find by making it so that you can see all interpolation methods in a single place. Of course you have to figure out that you need to look under `interpolate` first, but then you'd see all the various methods that are available for filling missing cells. If it's on the column directly then the user would have to look through a much larger set of methods to figure out what interpolation methods are available and that could be a bit overwhelming.

I'm fine with putting them directly on the column if that's clearer though.

PR #666, opened 2019-09-18
Add standard deviation to rolling column

benmccann:
Thanks for contributing.

## Description

I wanted to this reproduce [this chart from a Quora answer](https://qph.fs.quoracdn.net/main-qimg-a1f8ef78c1f847a1743d328fe03fcfb9.webp)

Added both `stdDev` and `standardDeviation` since they were both present in `AggregateFunctions`. They both do the same thing. Should we just have one and deprecate the other or leave both?


lwhite1:
nice enhancement.
I think we should probably deprecate standardDeviation in AggregateFunctions and not include it here.

benmccann:
Done

PR #669, opened 2019-09-19
Remove hundreds of lines of type casts

benmccann:
Thanks for contributing.

## Description

Moved casts from individual column types to `AbstractColumn`

Breaking change: `table.nCol` now returns `NumericColumn` instead of `NumberColumn` because `NumberColumn` was changed to take two type parameters

## Testing

Existing unit tests and compiler


lwhite1:
If I understand correctly, the net result of this change is that instead of getting a _DateColumn_ if I call first(10) on a _DateColumn_, I will get a Column<LocalDate> and a _Column<LocalDate>_ has no date-specific methods on it, so these methods can't be chained anymore, and can't be used in chains with date map functions and date filter functions, like:

`dateCol.sample(10000).year()
`

In that case, I think it's more correct to say that we'd be moving the casting into the client code, rather than removing it. I see the appeal, but I'm not sure it's a net gain.

benmccann:
No, if you call `first(10)` you should still get a `DateColumn`, so chaining should still work. I agree with you we want to return the specific column type still and not just `Column`

PR #671, opened 2019-09-24
Add method to apply function to column

benmccann:
Thanks for contributing.

## Description

Most of our methods return copies of columns. That means whenever I want to edit a column in a table I have to call `replaceColumn`. This makes that easier.

Adds a method to allow you to do:

```
table.replaceColumn("monthlyPriceIndex", col -> col.interpolate().backfill());
```

Instead of:

```
table.replaceColumn("monthlyPriceIndex", table.column("monthlyPriceIndex").interpolate().backfill());
```


benmccann:
Actually, this is not that useful because it only supports untyped columns

PR #675, opened 2019-09-28
Update gettingstarted.md

adamsiembida:

## Description

The name of a function was incorrect in the getting started document. `rowNumber` was fixed to `getRowNumber`.

## Testing

Did you add a unit test? No.

### DIFF
--- a/docs/gettingstarted.md
+++ b/docs/gettingstarted.md
@@ -309,11 +309,11 @@ destinationTable.addRow(43, sourceTable);	// adds row 43 from sourceTable to the
 table.sample(200);				// select 200 rows at random from table 
 ```
 
-You can also perform arbitrary operations on each row in the table.  One way is to just iterate over the rows and work with each column individually, using Row::rowNumber() as the index:
+You can also perform arbitrary operations on each row in the table.  One way is to just iterate over the rows and work with each column individually, using Row::getRowNumber() as the index:
 
 ```java
 for (Row row : table) {
-    System.out.println(column1.get(row.rowNumber())); // etc.
+    System.out.println(column1.get(row.getRowNumber())); // etc.
 }
 ```
 


benmccann:
Thanks for catching! I think that might be the compiled version of the docs. @ryancerf would know best. He was updating them to compile the code snippets. I think you need to update https://github.com/jtablesaw/tablesaw/blob/master/docs-src/main/gettingstarted.md instead

adamsiembida:
@benmccann Thanks for pointing that out. I'm trying to fix that now.

However, I'm a little bit confused about the code snippet part. I see that in https://github.com/jtablesaw/tablesaw/blob/master/docs-src/main/gettingstarted.md

there is the markdown that needs to be fixed:

`You can also perform arbitrary operations on each row in the table.  One way is to just iterate over the rows and
work with each column individually, using Row::rowNumber() as the index`

However I can't find where the code for the code snippet is. In the markdown file I see:

`
@@snip [for_loop](./src/main/java/tech/tablesaw/docs/GettingStarted.java)
`

But when I look that up in `GettingStarted.java`, I see

```
// @@ for_loop
for (Row row : table) {
  System.out.println("On " + row.getDate("date") + ": " + row.getDouble("approval"));
}
// @@ for_loop
```

However in the rendered docs it looks like it results in

```
for (Row row : table) {
    System.out.println(column1.get(row.rowNumber())); // etc.
}
```

The snippet in the java file and rendered file don't seem to match up. I'm probably not looking in the right place, but I can't find where the code snippet is actually taken from.

benmccann:
@ryancerf just set this up recently, so would be the best person to ask

ryancerf:
The TLDR is that gettingstarted.md has been updated in docs-src/ but it has not been released. To release the docs all you have to do is copy the contents of docs-src/dist to to docs/ where github reads the docs from. I am going to close this PR and create another PR to release the docs.

Thanks for contributing. 

FYI. Here is more information on how our custom docs work.

1) Content files (markdown) files in `docs-src/main/`. These contain the content and links to code snippets defined in java src files.

2) src files (java) that hold the compiled code (can be anywhere in the project, but are generally in `docs-src/src/`)

3) the output of running snippets in the src files. E.G. the output of running `table.first(10)`

4) The built docs are in `docs-src/dist/` (have not necessarily released)

5) The released/deployed docs are in `docs/` at the root of the project. (Github automatically serves whatever is in the `docs/` directory in the master branch) To release the docs copy docs/dist/ to docs/. That is what the docs-src/deploy_docs.sh shel script is for.

PR #682, opened 2019-10-10
Reads a specific table from the HTML page if multiple tables are pres…

emilianbold:
…ent based on HtmlReadOptions.Builder.tableIndex

Thanks for contributing.

## Description

Added HtmlReadOptions.Builder.tableIndex to allow selecting which HTML table should be read from a page with multiple tables.

## Testing

No, tested manually.

### DIFF
--- a/html/src/main/java/tech/tablesaw/io/html/HtmlReader.java
+++ b/html/src/main/java/tech/tablesaw/io/html/HtmlReader.java
@@ -42,14 +42,24 @@ public class HtmlReader implements DataReader<HtmlReadOptions> {
       doc = Parser.htmlParser().parseInput(options.source().createReader(null), "");
     }
     Elements tables = doc.select("table");
+    int tableIndex = 0;
     if (tables.size() != 1) {
-      throw new IllegalStateException(
-          "Reading html to table currently works if there is exactly 1 html table on the page. "
-              + " The URL you passed has "
-              + tables.size()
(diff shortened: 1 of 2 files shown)


html/src/main/java/tech/tablesaw/io/html/HtmlReader.java
```diff
@@ -42,14 +42,24 @@ public Table read(HtmlReadOptions options) throws IOException {
...
-              + ". You may file a feature request with the URL if you'd like your pagae to be supported");
+      if (options.tableIndex() != null) {
+        if (options.tableIndex() >= 0 && options.tableIndex() < tables.size()) {
+          tableIndex = options.tableIndex();
+        } else {
+          throw new IndexOutOfBoundsException(
+              "Table index outside bounds. The URL has " + tables.size() + " tables");
+        }
+      } else {
+        throw new IllegalStateException(
```
benmccann: I might change this message now. How about `tables.size() + " tables found. When more than one html table is present on the page you must specify the index of the table to read from."`?

emilianbold:
Sure, we can change that message. Do you want me to update the PR?

I just picked a random page when I wrote a test which happened to have a
second smaller table in there. Seemed like an easy fix.

--emi

joi, 10 oct. 2019, 15:46 Ben McCann <notifications@github.com> a scris:

emilianbold:
I've updated the error message.

PR #683, opened 2019-10-10
Reads a specific sheet from the Excel file based on XlsxReadOptions sheetIndex

emilianbold:
…sheetIndex

Thanks for contributing.

## Description

Adds XlsxReadOptions.tableIndex to select which table to return by default in `read(XlsxReadOptions)`. Somewhat in style with the HtmlReadOptions flag.

I wanted this to be a `sheetIndex` but then noticed some sheets might not have tables so the index wouldn't match...

## Testing

Yes.


benmccann:
There's a typo in "file" in the commit message

benmccann:
There's a typo in "file" in the commit message. Also the commit message says `sheetIndex` but then the code uses `tableIndex`

excel/src/main/java/tech/tablesaw/io/xlsx/XlsxReadOptions.java
```diff
@@ -12,8 +12,12 @@
 
 public class XlsxReadOptions extends ReadOptions {
 
+  /** Which table should be read? Null means no specific index was set. First table has index 0. */
```
benmccann: I think it sounds funny to put a question in the javadoc. How about "The table to read" instead of "Which table should be read?"

benmccann:
I also like the idea of `sheetIndex` instead of `tableIndex`. You could make that work by having `readMultiple` return a list containing `null` values when there's no table in a sheet or a `Map<Integer, Table>` with the sheet number

What do you think @hallvard? I'd like to get your opinion since you wrote this code initially

emilianbold:
You are right, `sheetIndex` is what I originally wanted too. I've updated the commit, fixed the message and added more tests.

Picked the idea of adding `null`s for empty sheets, but I do this in a protected method so the public `readMultiple` method remains unchanged for users.

excel/src/main/java/tech/tablesaw/io/xlsx/XlsxReader.java
```diff
@@ -60,14 +61,47 @@ public static void register(ReaderRegistry registry) {
...
+      if (options.sheetIndex() >= 0 && options.sheetIndex() < tables.size()) {
+        int index = options.sheetIndex();
+
+        Table table = tables.get(index);
+        if (table == null) {
+          throw new IllegalArgumentException("No tables found.");
+        }
+        return table;
+      } else {
+        throw new IndexOutOfBoundsException(
```
benmccann: How about:

```
if (options.sheetIndex() < 0 || options.sheetIndex() >= tables.size()) {
  throw new IndexOutOfBoundsException(
    "Sheet index outside bounds. " + tables.size() + " sheets found.");
}
```

Then the rest of the code doesn't need to be nested in the `if/else` statement

excel/src/main/java/tech/tablesaw/io/xlsx/XlsxReader.java
```diff
@@ -60,14 +61,47 @@ public static void register(ReaderRegistry registry) {
...
+            "Sheet index outside bounds. " + tables.size() + " sheets found.");
+      }
+    }
+    // since no specific sheetIndex asked, return first table
+    Optional<Table> table = tables.stream().filter(t -> t != null).findFirst();
```
benmccann: Could do:

```
return tables.stream().filter(t -> t != null).findFirst()
    .orElseThrow(() -> new IllegalArgumentException("No tables found."));
```

excel/src/main/java/tech/tablesaw/io/xlsx/XlsxReader.java
```diff
@@ -60,14 +61,47 @@ public static void register(ReaderRegistry registry) {
...
   }
 
   public List<Table> readMultiple(XlsxReadOptions options) throws IOException {
+    return readMultiple(options, false);
+  }
+
+  /**
+   * Read at most a table from every sheet.
+   *
+   * @param options
```
benmccann: should either fill out or remove. probably the same with the `@throws` documentation

benmccann:
Great idea! I like it! Just a few minor comments

emilianbold:
OK, updated the PR.

emilianbold:
I think it's looking pretty good. Do I need to change something else?

excel/src/main/java/tech/tablesaw/io/xlsx/XlsxReader.java
```diff
@@ -60,14 +60,42 @@ public static void register(ReaderRegistry registry) {
 
   @Override
   public Table read(XlsxReadOptions options) throws IOException {
-    List<Table> tables = readMultiple(options);
+    List<Table> tables = readMultiple(options, true);
     if (tables.isEmpty()) {
```
benmccann: You probably don't need this check anymore since there are other checks that would cover it as well

emilianbold: :-)

excel/src/main/java/tech/tablesaw/io/xlsx/XlsxReader.java
```diff
@@ -60,14 +60,39 @@ public static void register(ReaderRegistry registry) {
...
+    if (options.sheetIndex() != null) {
+      int index = options.sheetIndex();
+      if (index < 0 || index >= tables.size()) {
+        throw new IndexOutOfBoundsException(
+            "Sheet index outside bounds. " + tables.size() + " sheets found.");
+      }
+
+      Table table = tables.get(index);
+      if (table == null) {
+        throw new IllegalArgumentException("No tables found.");
```
benmccann: Let's change this message to:

    throw new IllegalArgumentException("No table found at index " + index);

emilianbold: Maybe something like `"No table found for sheet index " + index`?

emilianbold: Or `found at sheet index`?

benmccann: sure. "at sheet index" sounds good to me

excel/src/main/java/tech/tablesaw/io/xlsx/XlsxReader.java
```diff
@@ -60,14 +60,39 @@ public static void register(ReaderRegistry registry) {
...
   public Table read(XlsxReadOptions options) throws IOException {
-    List<Table> tables = readMultiple(options);
-    if (tables.isEmpty()) {
-      throw new IllegalArgumentException("No tables found.");
+    List<Table> tables = readMultiple(options, true);
+    if (options.sheetIndex() != null) {
+      int index = options.sheetIndex();
+      if (index < 0 || index >= tables.size()) {
+        throw new IndexOutOfBoundsException(
+            "Sheet index outside bounds. " + tables.size() + " sheets found.");
```
benmccann: sorry. one last thing I just noticed. How about also adding the index here:

    "Sheet index " + index + " outside bounds. " + tables.size() + " sheets found.");

emilianbold: Done. If you want I can also add a test for this:

`assertEquals("Sheet index 5 outside bounds. 2 sheets found.", iobe.getMessage());`

benmccann:
This is great! Thanks for all the help!

PR #686, opened 2019-10-19
Added options to ignore leading and trailing whitespaces in quoted CSV

aecio:

## Description

This PR adds the options `ignoreLeadingWhitespacesInQuotes()` and `ignoreTrailingWhitespacesInQuotes()` in `CsvReadOptions`, and sets them in the underlying univocity CSV parser.

## Testing

Did you add a unit test?

Yes, there are 5 new unit tests.


lwhite1:
@aecio Thanks for the submission and the tests, but I find this PR a bit confusing. 

If I'm reading this correctly, you first added, then removed an option called "trimHeaders" that seemed to do what you wanted. Why is that? 

Also, the Univocity methods you're using, _i think_, "ignore" whitespace for all rows, not just the headers. Is that correct? If so, you're comments in the PR (e.g. "Add options to ignore spaces in quoted header names") don't describe what you implemented. 

Finally, I feel the Univocity method names are sufficiently bad that they shouldn't be propagated to Tablesaw:  If you "ignore" whitespace, does that mean you leave them where you see them? It sounds that way to me, but looking at their code I think setting ignore to **true** causes the whitespace to be trimmed. trimWhitespace... would be much clearer. If you don't provide comments the names have to be unambiguous.

aecio:
I implemented the trim headers option some time ago, and while doing that I realized that univocity had these options to ignore whitespace of quoted values, so I thought it could be better to just use them since other univocity options are already exposed.

I also agree that the option names are somewhat ambiguous. Let me know if you'd like to include any of these options, and if so, how they should be named and then I could update the PR. 

I could add trimHeaders() option again, rename the univocity options, and make them disabled by default.

lwhite1:
> I implemented the trim headers option some time ago, and while doing that I realized that univocity had these options to ignore whitespace of quoted values, so I thought it could be better to just use them since other univocity options are already exposed.
> 
> I also agree that the option names are somewhat ambiguous. Let me know if you'd like to include any of these options, and if so, how they should be named and then I could update the PR.
> 
> I could add trimHeaders() option again, rename the univocity options, and make them disabled by default.

If I understand the Univocity methods correctly, it's impossible to use them and maintain compatibility with the current version, because the current version defaults to:

1. Trim the Headers
2. Don't trim the content rows.

but the Univocity methods will make it either trim both, or trim neither. 
If that's correct they can't be used.

aecio:
Right, I mentioned disabling the univocity options (`ignoreLeadingWhitespacesInQuotes`, `ignoreTrailingWhitespacesInQuotes`) by default to avoid breaking compatibility. It would keep the current behavior unless users turn these options on.

The other possibility is to bring back the trimHeaders() from the first commit, which does not use univocity options. This would allow disabling only the behavior "1. Trim the Headers" while keeping the content intact.

So we could:
1. Add only the univocity options
2. Add only the option to disable trim headers
3. Add both 
4. Add none

I don't have a strong opinion for any of these options, so feel free to close this pull request if you think it is best to to add none.

core/src/main/java/tech/tablesaw/io/csv/CsvReadOptions.java
```diff
@@ -139,6 +151,8 @@ public int maxCharsPerColumn() {
     private Integer maxNumberOfColumns = 10_000;
     private Character commentPrefix;
     private boolean lineSeparatorDetectionEnabled = true;
+    private boolean ignoreTrailingWhitespacesInQuotes = true;
```
benmccann: I would default these to `false` or `null`. That's both more inline with what I'd expect, I think it'd make it easier for users to understand that we just use the univocity defaults, and it'd keep backwards compatibility as discussed on the thread earlier

benmccann:
@aecio thanks for this. I'm not sure what you need to accomplish your goal. If you revert the changes `AbstractColumn.java` and `FileReader.java` will that still work for your use case? It sounds like that might be more amenable to Larry

Also, this PR will need to be rebased

aecio:
I was just surprised when I found out that tablesaw trims whitespace even when those spaces are inside quoted strings (and there is no option to disable it). It is not what I would expect from any CSV parser without enabling an explicit configuration for this behavior. Trimming a quoted string is a modification of the original data and could cause unexpected behavior, e.g., hash functions could generate different hash values, it could make equality comparisons to fail when comparing data read from tablesaw and data read from the original source of data in downstream applications.

All that being said, this is not causing problems (that I'm aware of) in my current use case. So I'm totally fine with just dropping this PR or modifying it in any way that the tablewsaw team would prefer. Unfortunately, I won't be able to work on it in the next few weeks.

benmccann:
While I somewhat understand what you're saying, it only does it in the headers and I don't know if the examples you listed are super relevant in that case. Given that it's not actually causing you any problems and you don't have time to revisit it at the moment I don't know how much it's worth spending time on. I'll close it for now to make it easier to keep track of what needs review, but if you want to revisit and have time to rebase the PR, feel free to reopen it.

PR #691, opened 2019-10-24
Adds plotly config argument. Only config.displayModeBar

emilianbold:
Thanks for contributing.

## Description

Plot.ly has a `config` argument ( https://plot.ly/javascript/configuration-options/ ) as the last argument to `Plotly.plot`. This has many options to customize the UI.

This PR adds the initial way of supporting this `config` object and adds the `displayModeBar` option.

Note that I've made `config` part of the `Layout` object. This isn't how Plot.ly does it but it seems convenient to have a single 'configuration object'. The `config` field is not saved to the context nor serialized to JS by `Layout`. It is the job of `Figure` to do that. If you don't like it I can remove this part.

## Testing

Yes, a basic one.

### DIFF
--- a/jsplot/src/main/java/tech/tablesaw/plotly/components/Config.java
+++ b/jsplot/src/main/java/tech/tablesaw/plotly/components/Config.java
@@ -1,0 +1,58 @@
+package tech.tablesaw.plotly.components;
+
+import com.mitchellbosecke.pebble.error.PebbleException;
+import com.mitchellbosecke.pebble.template.PebbleTemplate;
+import java.io.IOException;
+import java.io.StringWriter;
+import java.io.Writer;
+import java.util.HashMap;
+import java.util.Map;
+
+public class Config extends Component {
+
+  private final Boolean displayModeBar;
+
+  private Config(Builder builder) {
+    this.displayModeBar = builder.displayModeBar;
+  }
+
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  @Override
+  public String asJavascript() {
+    Writer writer = new StringWriter();
+    PebbleTemplate compiledTemplate;
+
+    try {
+      compiledTemplate = engine.getTemplate("config_template.html");
+      compiledTemplate.evaluate(writer, getContext());
+    } catch (PebbleException | IOException e) {
+      e.printStackTrace();
+    }
+    return writer.toString();
+  }
+
+  private Map<String, Object> getContext() {
+    Map<String, Object> context = new HashMap<>();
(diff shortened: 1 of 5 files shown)


emilianbold:
The Travis build shows as pending but seems to have finished successfully.

Would appreciate some feedback on this.

lwhite1:
Hi @emilianbold 

Sorry about the lack of feedback. We only have three active maintainers and I'm the only one who does visualization stuff. Right now I'm on vacation and when I get back on Tuesday I'm buried with work at my real job. (i work on cancer diagnostics so I don't feel right sneaking time out for my open source when we're behind.) Tablesaw needs more collaborators. 

If you'd like to consider becoming a maintainer, I will find time to review your submission and comments today. Otherwise, I will get to it when I can.

Again. sorry it's not quicker. Just a very busy time right now.

emilianbold:
I can become a maintainer: I have the time and a project I work on depends on Tablesaw.

I mostly need to add many things to the visualization part these days.

lwhite1:
> I can become a maintainer: I have the time and a project I work on depends on Tablesaw.
> I mostly need to add many things to the visualization part these days.

That sounds great. I have a bit of time now and will try to at least answer your questions. In general, visualization is the place with the greatest need for work, so this could work out well.

> Plot.ly has a config argument ( https://plot.ly/javascript/configuration-options/ ) as the last argument to Plotly.plot. This has many options to customize the UI. 

I didn't realize this existed. It would be awesome to have this in Tablesaw.  

> Note that I've made config part of the Layout object. This isn't how Plot.ly does it but it seems convenient to have a single 'configuration object'. The config field is not saved to the context nor serialized to JS by Layout. It is the job of Figure to do that. If you don't like it I can remove this part.

I think i prefer to mirror the Plot.ly code for a few reasons: 
- even though they're both configs, they're intended to deal with different plotting needs: layout vs interaction (of course there is some overlap, but the idea makes sense, I think.)
- keeping a design that is close to plot.ly may pay off if the Plot.ly code is restructured. It will be clear where to look for any of our code that needs to change. 
- having the Layout serialize some of its elements but not all of them is likely to confuse people. "I created my figure and am serializing the layout, but..."
- having object A hold B and still having to pass both into the Figure builder

So i would have the figure builder take three objects: layout, config, and trace(s) as you are,  but leave the config and layout objects separate.

Does this sound ok? 

(I'll come back to your other PR a little later tonight)

emilianbold:
Yes, it makes sense to keep layout and config separate. I've pushed a commit that fixes this.

Now only `Figure` and the builder deals with `Config`.

PR #697, opened 2019-10-29
Adds firstAsString aggregate function

emilianbold:
Thanks for contributing.

## Description

Added `AggregateFunctions.firstAsString` which is similar to `AggregateFunctions.first` except it returns a `String` value regardless of the column type.

Not sure if fits the overall architecture as aggregate functions seem to be designed for numeric data only, but it does make sense somehow.

I pondered about this while trying to make unique (parent,child) pairs from two columns and `firstAsString` + `groupBy` seemed the most logical thing: `table.summarize(parentCol, firstAsString).by(col)`.

I don't mind if you reject the PR, I know TOP1 is not an aggregate function in SQL.

## Testing

Yes.

### DIFF
--- a/core/src/test/java/tech/tablesaw/aggregate/AggregateFunctionsTest.java
+++ b/core/src/test/java/tech/tablesaw/aggregate/AggregateFunctionsTest.java
@@ -25,7 +25,7 @@ import static tech.tablesaw.aggregate.AggregateFunctions.countTrue;
 import static tech.tablesaw.aggregate.AggregateFunctions.countUnique;
 import static tech.tablesaw.aggregate.AggregateFunctions.countWithMissing;
 import static tech.tablesaw.aggregate.AggregateFunctions.earliestDate;
-import static tech.tablesaw.aggregate.AggregateFunctions.firstString;
+import static tech.tablesaw.aggregate.AggregateFunctions.firstAsString;
 import static tech.tablesaw.aggregate.AggregateFunctions.latestDate;
 import static tech.tablesaw.aggregate.AggregateFunctions.mean;
 import static tech.tablesaw.aggregate.AggregateFunctions.noneTrue;
@@ -111,7 +111,7 @@ class AggregateFunctionsTest {
   }
 
   @Test
-  void testFirstString() {
+  void testFirstAsString() {
     StringColumn company = StringColumn.create("Company");
     StringColumn sector = StringColumn.create("Sector");
     Table t = Table.create(company, sector);
@@ -125,7 +125,7 @@ class AggregateFunctionsTest {
       r.setString(0, "Facebook");
       r.setString(1, "Technology");
     }
-    Table result = t.summarize("Company", firstString).by("Sector");
+    Table result = t.summarize("Company", firstAsString).by("Sector");
     assertEquals(1, result.rowCount());
     assertEquals("Technology", result.get(0, 0));
     assertEquals("Google", result.get(0, 1));
--- a/core/src/test/java/tech/tablesaw/aggregate/AggregateFunctionsTest.java
+++ b/core/src/test/java/tech/tablesaw/aggregate/AggregateFunctionsTest.java
@@ -25,6 +25,7 @@ import static tech.tablesaw.aggregate.AggregateFunctions.countTrue;
 import static tech.tablesaw.aggregate.AggregateFunctions.countUnique;
 import static tech.tablesaw.aggregate.AggregateFunctions.countWithMissing;
 import static tech.tablesaw.aggregate.AggregateFunctions.earliestDate;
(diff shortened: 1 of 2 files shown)


benmccann:
Top makes me think it's first sorted. Maybe call it firstStrings or something?

Also, I think it'd be nicer to give it a constructor so you can pass in the number and support any number

emilianbold:
`firstString` also works. Note that `summarize` returns a single string, not sure how I could take a number.

benmccann:
Maybe it wouldn't make sense to take a number parameter in this case. I was thrown off by the name `top1` thinking you'd also want a `top5`, `top10`, etc. and didn't want to create so many variants, but then the return type would be different in those cases because you wouldn't be returning `String` anymore, so disregard that thought

emilianbold:
Renamed to `firstString`. Does it make sense?

benmccann:
I just realized that `firstString` makes it sound like it's searching the column for different types and returning the first one that's a string. Maybe `firstAsString` would make more sense?

emilianbold:
But a column has only one type, no? I can rename it to `firstAsString`, but I was just wondering if it really helps.

benmccann:
Yeah, it's true. I think it sounds more natural though, so would probably still prefer it

emilianbold:
OK, renamed to `firstAsString`.

ryancerf:
@emilianbold welcome to the project. Thanks for all the help.

Given how big the library is already and that it is possible for users to write their own aggregate functions I am not crazy about adding this, but I will defer to others for the final decision.

If you do go ahead and submit this please include javadoc.

emilianbold:
I don't mind either way, but I would prefer some expediency. This is a very low-value commit which is taking a lot of back and forth. If it's good, let's merge it, if not let's unceremoniously close the PR.

I'm not attached to it at all, it was just neat that it could be done and I figured it might help others...

Don't want to sound brisk, just trying to limit the number of balls in the air.

core/src/test/java/tech/tablesaw/aggregate/AggregateFunctionsTest.java
```diff
@@ -108,6 +110,27 @@ void testGroupBy() {
     assertEquals(13, result.rowCount());
   }
 
+  @Test
+  void testFirstAsString() {
+    StringColumn company = StringColumn.create("Company");
+    StringColumn sector = StringColumn.create("Sector");
+    Table t = Table.create(company, sector);
+    {
```
benmccann: this syntax is confusing to me. I've never seen curly braces used like this before

emilianbold: :-) It's just a code block that allows me to reuse `Row r` since they have a different scope. Handy for tests.

benmccann: I'd prefer to avoid since it's a non-standard construct. Probably the cleanest thing to do would make the methods chainable so that we can do:
```
t.appendRow().setString(0, "Google").setString(1, "Technology");
t.appendRow().setString(0, "Facebook").setString(1, "Technology");
```

emilianbold: Ha, that is neat!

ryancerf:
@emilianbold 

Thanks for preparing this change, but since we both agree that it is low value I am going to close it.

I absolutely agree that we need to make decisions faster and minimize the number of balls in the air. Sorry for being slow and indecisive on this one. I will try to do better.

emilianbold:
This might be a silly question but I'm more of a chart user then a core user.

So, how would I go about adding an `UNIQUE`-like constraint on an `id` column?

Seems to me a good trick is to go with:

`table.summarize(columnNames, AggregateFunctions.first, firstString).by("id")`

Looks like I need an aggregate function that maps to all `ColumnType`s so I can take just a first pick.

lwhite1:
Not sure what you’re trying to do, but uniqueness constraints are more a DB
function than a dataframe function.

emilianbold:
I'm trying to prepare the data for a chart. Plotly expects unique values somewhere and I was wondering how to use a Table directly (instead of doing this manually). `summarize(col, first).by('id')` seems to do this as it just takes on row per id. But `first` only works for double columns, hence me having to re-invent `firstString`.

ryancerf:
This transformation makes sense, but I do not think it is super common and it is ok to force users to occasionally write their own aggregate functions.

I think this is what you are trying to do in SQL.
```sql
SELECT
  id,
  ANY_VALUE(CAST(col AS STRING)) AS col
FROM table
GROUP BY id;
```

I will probably add this as an example of a custom aggregate function in the docs.

PR #702, opened 2019-10-30
Escapes backslash too when creating Javascript string

emilianbold:
Thanks for contributing.

## Description

Forgot to escape backslash itself (`\`) when I did PR #699

## Testing
Yes.


emilianbold:
So, @lwhite1, can I just merge small PRs like this directly?

You have my email, we can talk over there about "house rules".

benmccann:
It's probably better to use a Json handling library like Jackson which is used other places in the code

emilianbold:
That seems a bit heavy but makes sense. I don't see Jackson in the dependencies though.

benmccann:
It looks like it's used in the `saw` and `json` sub-projects

emilianbold:
OK, using the util from Jackson now.

jsplot/src/main/java/tech/tablesaw/plotly/Utils.java
```diff
@@ -13,18 +14,22 @@ public static String quote(String string) {
...
 
   /** @return a Javascript array of strings (escaped if needed) */
   public static String dataAsString(Object[] data) {
     StringBuilder builder = new StringBuilder("[");
     for (int i = 0; i < data.length; i++) {
       Object o = data[i];
-      builder.append("'");
-      builder.append(escape(String.valueOf(o)));
-      builder.append("'");
+      builder.append("\"");
```
benmccann: I'd probably use Jackson to convert the whole array instead of just the value. But this seems like an improvement in any case

emilianbold: Yes, if you think about it 90% of the output template is a JSON. We could just marshal an object there with Jackson...

emilianbold: FYI, here is a branch where I'm trying to use Jackson instead of PebbleTemplate: https://github.com/emilianbold/tablesaw/commits/emi/jackson

It doesn't look too bad. JS code is indented better. I dislike that it uses quote for member names too (being JSON, not plain JS) but it's manageable.

A lot of work though to finish all those classes...

benmccann: Cool. You don't actually have to change that much though. Jackson can turn a `Map<String, Object>` into JSON correctly

emilianbold: > Jackson can turn a Map<String, Object> into JSON correctly

Heh, I actually didn't notice that. I thought using a (bean) object might come handy for hierarchies. For example `HoverLabel.font` used `Font`. But maybe we just even more Maps as values?

benmccann: I think it also handles a mix of beans and maps just fine. I don't know the jsplot code that well so whatever you think makes the most sense. Feel free to merge this whenever. I don't mean to hold it up

emilianbold: OK, I'll merge this PR and we can talk about Jackson when I do a PR about that.

PR #703, opened 2019-10-30
Adds ReadOptions.percentage which parses percentage-like strings as doubles

emilianbold:
Thanks for contributing.

## Description

Adds flag to parse percentage strings (10%) as doubles (0.1).

## Testing

Yes.

### DIFF
--- a/core/src/main/java/tech/tablesaw/columns/numbers/DoubleParser.java
+++ b/core/src/main/java/tech/tablesaw/columns/numbers/DoubleParser.java
@@ -7,6 +7,8 @@ import tech.tablesaw.io.ReadOptions;
 
 public class DoubleParser extends AbstractColumnParser<Double> {
 
+  private boolean percentage = false;
+
   public DoubleParser(ColumnType columnType) {
     super(columnType);
   }
@@ -16,15 +18,13 @@ public class DoubleParser extends AbstractColumnParser<Double> {
     if (readOptions.missingValueIndicator() != null) {
       missingValueStrings = Lists.newArrayList(readOptions.missingValueIndicator());
     }
+    percentage = readOptions.percentage();
   }
 
   @Override
   public boolean canParse(String s) {
-    if (isMissing(s)) {
-      return true;
-    }
     try {
-      Double.parseDouble(AbstractColumnParser.remove(s, ','));
+      parseDouble(s);
       return true;
     } catch (NumberFormatException e) {
       // it's all part of the plan
@@ -42,6 +42,12 @@ public class DoubleParser extends AbstractColumnParser<Double> {
     if (isMissing(s)) {
       return DoubleColumnType.missingValueIndicator();
     }
-    return Double.parseDouble(AbstractColumnParser.remove(s, ','));
+    boolean isPercentage = false;
+    if (percentage && s.endsWith("%")) {
+      isPercentage = true;
+      s = s.substring(0, s.length() - 1);
+    }
+    double d = Double.parseDouble(AbstractColumnParser.remove(s, ','));
+    return isPercentage ? d / 100.0 : d;
   }
 }
--- a/core/src/main/java/tech/tablesaw/io/ReadOptions.java
+++ b/core/src/main/java/tech/tablesaw/io/ReadOptions.java
@@ -79,6 +79,7 @@ public class ReadOptions {
   protected final DateTimeFormatter dateFormatter;
   protected final DateTimeFormatter dateTimeFormatter;
   protected final DateTimeFormatter timeFormatter;
+  protected final boolean percentage;
 
   protected final boolean header;
 
@@ -104,6 +105,7 @@ public class ReadOptions {
     } else {
       locale = builder.locale;
     }
+    percentage = builder.percentage;
   }
 
   public Source source() {
@@ -169,6 +171,10 @@ public class ReadOptions {
     return DateTimeFormatter.ofPattern(dateFormat, locale);
   }
 
+  public boolean percentage() {
+    return percentage;
+  }
+
   protected static class Builder {
 
     protected final Source source;
@@ -186,6 +192,7 @@ public class ReadOptions {
     protected boolean minimizeColumnSizes = false;
     protected boolean header = true;
     protected int maxCharsPerColumn = 4096;
+    protected boolean percentage = false;
 
     protected Builder() {
       source = null;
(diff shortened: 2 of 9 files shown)


benmccann:
I worry this might be a little bit on the specific side and there will be other transformations that users will ask for as well. What do you think about a more general API to allow the following? (I haven't thought too much about the name)

    readOptions.transform("percentageColumn", s -> s.endsWith("%") ? s.substring(0, s.length() - 1) : s);

core/src/main/java/tech/tablesaw/io/csv/CsvReadOptions.java
```diff
@@ -281,5 +281,14 @@ public Builder minimizeColumnSizes() {
       super.minimizeColumnSizes();
       return this;
     }
+
+    /**
+     * @param b if true, reads strings that appear as percentages (eg. 10%) as numbers (0.1)
+     * @return
```
lwhite1: We should not leave empty @return annotations in the javadoc.

core/src/main/java/tech/tablesaw/columns/numbers/DoubleParser.java
```diff
@@ -42,6 +42,12 @@ public double parseDouble(String s) {
     if (isMissing(s)) {
       return DoubleColumnType.missingValueIndicator();
     }
-    return Double.parseDouble(AbstractColumnParser.remove(s, ','));
+    boolean isPercentage = false;
+    if (percentage && s.endsWith("%")) {
```
lwhite1: hmm. 
So if i use the percentage() option, and the column's data doesn't end in % nothing happens. Is that what we want? What if I have a column where the numbers are percents but their is no trailing % sign? I would think that 10.4 should be converted to .104 even without a trailing percent sign, if the user has requested the transformation. 

Also, should this decision be made on a row-by-row basis? It seems like if the user says this is a percentage column we should treat the entire column as percents, otherwise a row where the percent sign is missing would not get converted. 

I guess the column might contain both 10% and .10, where both are intended to mean 10% but that kind of sloppy data is really something the user should fix before analysis, I think.

lwhite1:
I can see how this is useful, and it's not uncommon for java number parsing code to handle percentages.

It seems that you only implemented this for csv, and not for fixedWidth, JSON or Excel files. I get that CSV support is more important, but should we introduce inconsistencies?

core/src/main/java/tech/tablesaw/io/ReadOptions.java
```diff
@@ -294,6 +301,15 @@ public Builder minimizeColumnSizes() {
       return this;
     }
 
+    /**
+     * @param b if true, reads strings that appear as percentages (eg. 10%) as numbers (0.1)
```
lwhite1: If we don't make any changes to the logic, the code comment here is inadequate, since it doesn't explain what happens if there is no % on a given row. We should try to explain the API completely. 

Also, i didn't read this code carefully enough, but i assume its possible to apply the percentage option to a column of non-numeric data. We should probably state what error will occur if that happens.

core/src/main/java/tech/tablesaw/columns/numbers/DoubleParser.java
emilianbold: This parser is designed to help with the apparently text columns such as 10% which are actually numeric. I guess a column transformation could be done, but it seems quite practical and I've encountered CSV files like this so the practice exists in the wild.

I don't think it's practical to think about malformed files. The user is assumed to know why they picked the percentage parser.

Converting the numeric 10.4 value to 0.104 seems less usual but I should probably add it.

core/src/main/java/tech/tablesaw/io/ReadOptions.java
emilianbold: Hm, actually there's a potential problem here. The percentage flag is not set per column but per reader. So, all columns that end with % are treated as potential percentages. There's no way to say only a specific column is treated like this.

I wonder, how is this done for other column-level reader options?

benmccann: I don't think we have any options that are per-column

core/src/main/java/tech/tablesaw/io/csv/CsvReadOptions.java
emilianbold: Done.

core/src/main/java/tech/tablesaw/columns/numbers/DoubleParser.java
emilianbold: > So if i use the percentage() option, and the column's data doesn't end in % nothing happens. Is that what we want?

Yes. This flag is just telling the reader that % numbers are numbers and not a string. Without a % the normal parsing will happen.

> Also, should this decision be made on a row-by-row basis?

Yes, because there's no other mechanism. We have no per-column flags.

core/src/main/java/tech/tablesaw/io/ReadOptions.java
emilianbold: > If we don't make any changes to the logic, the code comment here is inadequate, since it doesn't explain what happens if there is no % on a given row. 

Nothing should happen since it won't be detected as a percentage. If you mean that data is not clean and some values might have % and some not, then all bets are off anyhow.

emilianbold:
> It seems that you only implemented this for csv, and not for fixedWidth, JSON or Excel files. I get that CSV support is more important, but should we introduce inconsistencies?

Done, see updated `*ReadOptions` classes.

core/src/main/java/tech/tablesaw/columns/numbers/DoubleParser.java
lwhite1: > I don't think it's practical to think about malformed files. The user is assumed to know why they picked the percentage parser.

That's kinda my point. It seems like if they say the column contains percentages, we should shift all the decimal points. Or throw an exception or something consistent. 

In any case, as long as the builder documentation for the operation specifies what happens it probably doesn't matter too much which solution we choose.

lwhite1: >Also, should this decision be made on a row-by-row basis?
>>Yes, because there's no other mechanism. We have no per-column flags.

yeah, but the row logic could be 

if (val ends in % or PERCENT_FLAG)

could it not?

emilianbold: But they don't say a given column contains percentages. They say that if a column looks line 'nnn%' it should be treated as numeric and read as a percentage. The same way that columns that look like a date get read as date_time.

lwhite1:
Closing after deciding to handle percentages as a regular numeric format, rather than as a special option to be turned on/off. 

@emilianbold Thank you for this and your other contributions.

Issue #704, opened 2019-10-30
Column.mapInto does not append


emilianbold:
The `Column.mapInto` documentation claims:

> Maps the function across all rows, appending the results to the provided Column

Except it doesn't seem to append data, just replace data. Particularly, creating an empty column to map something into will fail because `into.set` is called but `into.size() == 0`.

We could claim the name is correct as `into` might suggest replacing values, not appending.

Still, `Column` needs some way to convert the data type too. We have `map(Function<? super T, ? extends T> fun)` which does not allow, eg. a `String` to `double` conversion.

I suggest at least changing the Javadoc and showing an example as to how to covert types.

benmccann:
I'd probably prefer changing the javadoc rather than the behavior since the method's been around awhile

emilianbold:
Fixed by PR #705

PR #705, opened 2019-10-30
Updates the Column.mapInto javadoc and adds two new Column.map methods

emilianbold:
Thanks for contributing.

## Description

This fixed issue #704 by updating the `Column.mapInto` javadoc.

It also adds two new `Column.map` methods which offer a way to append data to a newly created column. The output column is easily created/provided via a `Supplier` or `Function`.

## Testing

Yes.

### DIFF
--- a/core/src/main/java/tech/tablesaw/columns/Column.java
+++ b/core/src/main/java/tech/tablesaw/columns/Column.java
@@ -348,24 +348,6 @@ public interface Column<T> extends Iterable<T>, Comparator<T> {
   }
 
   /**
-   * Maps the function across all rows, appending the results to the supplied Column.
-   *
-   * <p>Example:
-   *
-   * <pre>
-   * DoubleColumn d;
-   * StringColumn s = d.map(String::valueOf, () -> StringColumn.create("x"));
-   * </pre>
-   *
-   * @param fun function to map
-   * @param supplier the supplier for the column where results are appended to
-   * @return the Column with the results
-   */
-  default <R, C extends Column<R>> C map(Function<? super T, ? extends R> fun, Supplier<C> supplier) {
-    return map(fun, (name) -> supplier.get());
-  }
-
-  /**
    * Maps the function across all rows, appending the results to the created Column.
    *
    * <p>Example:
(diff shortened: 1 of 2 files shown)


emilianbold:
Ooh, interestingly `NumericColumn` has a separate `mapInto` that does `append`!

benmccann:
Let's remove the `NumericColumn` version then so it inherits the main one

I don't know that we need the new methods because it should be pretty easy to use the existing methods to accomplish the same thing. E.g. the user could just create the column with an initial size instead of empty and use `mapTo`

emilianbold:
To me something like `StringColumn s = d.map(String::valueOf, StringColumn::create)` looks in line with other stream APIs, eg. [`Stream.toArray(int[]::new)`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/stream/Stream.html#toArray(java.util.function.IntFunction) )

Right now `d.mapInto(String::valueOf, StringColumn.create(d.name(), d.size()))` works but I dislike the manual setting of the size (and, partially, the setting of the name).

I'll remove the `NumericColumn` method too.

benmccann:
Ok. Makes sense. At least for the one taking a creator. What's the use case for the one taking a supplier?

emilianbold:
Actually, the `NumericColumn.mapInto` takes a `DoubleFunction` so we can't remove it. We could add the same logic (`set` instead of `append`) but it might break existing code...

emilianbold:
> What's the use case for the one taking a supplier?

Good point, if we have the creator it's enough.

benmccann:
The new creator method let's you do `StringColumn s = d.map(String::valueOf, StringColumn::create`.  But actually, don't we already have a method that lets you do `StringColumn s = d.map(String::valueOf)`?

emilianbold:
> But actually, don't we already have a method that lets you do `StringColumn s = d.map(String::valueOf)`?

Not as far as I can tell. If `d` is a `DoubleColumn` then `map` will create another `DoubleColumn`. You cannot change the column type with a `map`, only `mapInto`.

PR #709, opened 2019-11-01
Updated Tables Documentation: Section Add and remove Rows

### LINKED ISSUE #708: Documentation: Adding Row to a table, dropping row and Adding row at index
I think Documentation on Rows is missing some basic operations such as how to add new Rows to a table, delete them and even adding rows at a certain index. It will certainly help someone who is interested in doing the same. 

Let me know if this would be helpful, so i can go ahead with a PR.

murtuza-ranapur:
Thanks for contributing.

## Description

Changes in file userguide/tables.md

1. Added documentation to add new Row
2. Added documentation to add new Row at certain index
3. Added documentation to remove Row
4. Added documentation to remove Row based on some criteria

Fixes #708 
## Testing

Not required

### DIFF
--- a/docs-src/main/userguide/tables.md
+++ b/docs-src/main/userguide/tables.md
@@ -171,8 +171,36 @@ Tablesaw supports inner and outer joins between tables.
 
 
 ## Add and remove rows
+###Adding a new Row
 
+To add a new row into the table you simply need to call
+```java
+Row newRow = aTable.appendRow();
+```
+then just set the values of each column in the row
+```java
+newRow.setString("s","value1");
+newRow.setDate("d",LocalDate.now());
+```
 
+###Remove rows from table
+
+To remove rows from table you need to call *dropRows()* on table with indexes that you want to drop
+```java
+Table newTableWithRemainingRows = table.dropRows(1,2);
+``` 
+
+###Remove rows based on given query
+
+Suppose table *aTable* consists of 2 columns 
+1. *IntColumn* named *ages*
+2. *StringColumn* named *names*
+
+then you can remove all the rows with age greater then 20 in the following manner
+
+```java
+Table newTableWithFilteredValues = aTable.dropWhere(ages.isGreaterThan(20)); 
+```
 
 ## Filter
 
--- a/docs-src/main/userguide/tables.md
+++ b/docs-src/main/userguide/tables.md
@@ -185,7 +185,7 @@ newRow.setDate("d",LocalDate.now());
 
 ###Remove rows from table
 
-To remove rows from table you need to call *dropRows()* on table with indexes that you want to drop
+To remove rows call *dropRows()* with indexes that you want to drop
 ```java
 Table newTableWithRemainingRows = table.dropRows(1,2);
 ``` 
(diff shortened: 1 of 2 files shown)


docs/userguide/tables.md
```diff
@@ -171,8 +171,51 @@ Tablesaw supports inner and outer joins between tables.
...
+newRow.setString("s","value1");
+newRow.setDate("d",LocalDate.now());
+Table newTableWithRowAtGivenIndex = top.append(bottom);
+```  
+
+###Remove rows from table
+
+To remove rows from table you need to call *dropRows()* on table with indexes that you want to drop
+```java
+Table newTableWithRemainingRows = table.dropRows(1,2,..);
```
ryancerf: Please remove the ,... so this would compile if copied and pasted.

docs/userguide/tables.md
```diff
@@ -171,8 +171,51 @@ Tablesaw supports inner and outer joins between tables.
...
+```java
+Row newRow = aTable.appendRow();
+```
+then just set the values of each column in the row
+```java
+newRow.setString("s","value1");
+newRow.setDate("d",LocalDate.now());
+```
+
+###Adding row at certain index
```
ryancerf: I do not think this is a very common pattern and would prefer to leave this out of the docs. I think it is more noise than it is worth.

docs/userguide/tables.md
```diff
@@ -171,8 +171,51 @@ Tablesaw supports inner and outer joins between tables.
 
 
 ## Add and remove rows
+###Adding New Row
 
+To add a new row into the table you simply need to call
```
ryancerf: This is a very common pattern. I can not believe we were missing this. Thanks for adding!

docs/userguide/tables.md
murtuza-ranapur: i was also dubious about adding this point, but then i though it was certainly useful to me so it might be helpful to someone else too. Plus i also though it would give user idea of of how to split the tables (As the documentation already has section on combining the tables but not splitting). Anyway i will just leave this part out.

docs/userguide/tables.md
murtuza-ranapur: Thanks @ryancerf actually i would like to contribute more in the documentations, so i think it would be great if i could get some guidelines for the same.

murtuza-ranapur:
Made suggested changes and moved code to docs-src/main/userguide/tables.md

docs-src/main/userguide/tables.md
```diff
@@ -171,8 +171,36 @@ Tablesaw supports inner and outer joins between tables.
...
+```
+then just set the values of each column in the row
+```java
+newRow.setString("s","value1");
+newRow.setDate("d",LocalDate.now());
+```
+
+###Remove rows from table
+
+To remove rows from table you need to call *dropRows()* on table with indexes that you want to drop
```
ryancerf: A bit wordy. Please change to something like.

To remove rows call *dropRows()* with indexes of the rows you want to drop.

murtuza-ranapur: Yeah sure

murtuza-ranapur:
Yeah sure, meanwhile i will keep adding documentation in the missing sections.

PR #710, opened 2019-11-04
Adds custom color scale to Marker

emilianbold:
Thanks for contributing.

## Description

Adds `MarkerBuilder.addColorScale(double value, int r, int g, int b)` and related JS generation code which allows custom (non-palette) color scales.

## Testing

Yes.

### DIFF
--- a/jsplot/src/main/java/tech/tablesaw/plotly/components/Marker.java
+++ b/jsplot/src/main/java/tech/tablesaw/plotly/components/Marker.java
@@ -8,7 +8,9 @@ import com.mitchellbosecke.pebble.template.PebbleTemplate;
 import java.io.IOException;
 import java.io.StringWriter;
 import java.io.Writer;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import tech.tablesaw.api.NumericColumn;
 import tech.tablesaw.plotly.Utils;
@@ -75,6 +77,8 @@ public class Marker extends Component {
   private final Line line;
   private final String[] color;
   private final Palette colorScalePalette;
+  private final List<Double> colorScaleCustomRange;
+  private final List<Color> colorScaleCustomRGB;
   private final boolean cAuto;
   private final double cMin;
   private final double cMax;
@@ -100,6 +104,8 @@ public class Marker extends Component {
     colorArray = builder.colorArray;
     gradient = builder.gradient;
     colorScalePalette = builder.colorScalePalette;
+    colorScaleCustomRange = builder.colorScaleCustomRange;
+    colorScaleCustomRGB = builder.colorScaleCustomRGB;
     cAuto = builder.cAuto;
     cMin = builder.cMin;
     cMax = builder.cMax;
@@ -129,7 +135,9 @@ public class Marker extends Component {
     Map<String, Object> context = new HashMap<>();
     context.put("size", size.length == 1 ? size[0] : Utils.dataAsString(size));
     if (colorScalePalette != null) {
-      context.put("colorScale", colorScalePalette);
+      context.put("colorScale", "'" + colorScalePalette + "'");
+    } else if (colorScaleCustomRange != null && !colorScaleCustomRange.isEmpty()) {
+      context.put("colorScale", asColorScaleString(colorScaleCustomRange, colorScaleCustomRGB));
     }
     if (cAuto != DEFAULT_C_AUTO) context.put("cAuto", cAuto);
     if (color != null && color.length > 0) {
@@ -155,6 +163,39 @@ public class Marker extends Component {
     return context;
   }
 
+  static class Color {
+
+    final int r, g, b;
+
+    public Color(int r, int g, int b) {
+      this.r = r;
+      this.g = g;
+      this.b = b;
+    }
+  }
+
+  private static String asColorScaleString(List<Double> values, List<Color> colors) {
+    StringBuilder sb = new StringBuilder();
+    sb.append("[\n");
+    for (int i = 0; i < values.size(); i++) {
+      Color c = colors.get(i);
+      if (i != 0) {
+        sb.append(",\n");
+      }
+      sb.append("['")
+          .append(values.get(i))
+          .append("', 'rgb(")
+          .append(c.r)
+          .append(",")
+          .append(c.g)
+          .append(",")
+          .append(c.b)
+          .append(")']");
+    }
+    sb.append("\n]");
+    return sb.toString();
+  }
+
   public static class MarkerBuilder {
 
     private double[] size = {6};
@@ -165,6 +206,8 @@ public class Marker extends Component {
 
     private Gradient gradient;
     private Palette colorScalePalette;
+    private List<Double> colorScaleCustomRange;
+    private List<Color> colorScaleCustomRGB;
     private boolean cAuto = DEFAULT_C_AUTO;
     private double cMin;
     private double cMax;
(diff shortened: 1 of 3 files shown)


emilianbold:
Does this project have a mailing list or some other way to talk among developers / users?

This commit seems fine to me and I would merge it myself but it's unclear to me if you are fine with this approach.

benmccann:
I've never been involved with any of the jsplot stuff, so I'll leave this one to you and Larry

emilianbold:
@benmccann I'm a committer now so if it was up to me I'd just push these changes. I just want to make sure you guys are OK with me doing this by myself? I will be careful, of course, not to break existing APIs, etc. and ask for confirmations if it looks complicated. For many smaller things though it would seem faster to just commit.

I just realized GitHub offers no way for committer to talk among eachother. There's no chat, no mailing list. It's only the discussions in the PRs. Surprising, now that I've noticed it.

benmccann:
I'll usually review commits to the core library. I'd ask @lwhite1 what he prefers for the jsplot library since he's been the one maintaining this one

lwhite1:
@emilianbold  It will be weeks-to-months before I'm able to devote any time to this. If @benmccann and @ryancerf  are unavailable, I would say use your judgement

jsplot/src/main/java/tech/tablesaw/plotly/components/Marker.java
```diff
@@ -155,6 +163,39 @@ public String asJavascript() {
     return context;
   }
 
+  static class Color {
+
+    final int r, g, b;
```
ryancerf: What about adding the colorScaleCustomRange value to the Color class? It looks like colors and values are always added at the same time.

That will allow you to only add one field to the MarkerBuilder (not two).

jsplot/src/main/java/tech/tablesaw/plotly/components/Marker.java
```diff
@@ -155,6 +163,39 @@ public String asJavascript() {
...
+
+  private static String asColorScaleString(List<Double> values, List<Color> colors) {
+    StringBuilder sb = new StringBuilder();
+    sb.append("[\n");
+    for (int i = 0; i < values.size(); i++) {
+      Color c = colors.get(i);
+      if (i != 0) {
+        sb.append(",\n");
+      }
+      sb.append("['")
```
ryancerf: If you add value to the color class as mentioned above you could factor this out into a separate method that takes no parameters.

jsplot/src/main/java/tech/tablesaw/plotly/components/Marker.java
```diff
@@ -155,6 +163,39 @@ public String asJavascript() {
...
+          .append(values.get(i))
+          .append("', 'rgb(")
```
ryancerf: This would be more concise with String.format. Something like.

 String.format("[r(%d,%d,%d)]", c.r, c.g, c.b)

jsplot/src/main/java/tech/tablesaw/plotly/components/Marker.java
```diff
@@ -129,7 +135,9 @@ public String asJavascript() {
     Map<String, Object> context = new HashMap<>();
     context.put("size", size.length == 1 ? size[0] : Utils.dataAsString(size));
     if (colorScalePalette != null) {
```
ryancerf: If colorScalePalette always takes precedence over colorScaleCustomRange should we mention that in the javadoc somewhere?

ryancerf:
@emilianbold 

> I'm a committer now so if it was up to me I'd just push these changes. I just want to make sure you guys are OK with me doing this by myself? I will be careful, of course, not to break existing APIs, etc. and ask for confirmations if it looks complicated. For many smaller things though it would seem faster to just commit.

In core (where I have done my work) we generally try to give each other a bit of time to review before submitting(1-2 days). For changes with little API impact in the plotly stuff I think it is fine to take more leeway especially if we are slowing you down and we have already discussed the issue.

For non trivial changes in core please wait for Larry or Ben to review.

jsplot/src/test/java/tech/tablesaw/components/MarkerTest.java
```diff
@@ -16,4 +16,18 @@ public void asJavascript() {
...
+
+  @Test
+  public void testCustomPalette() {
+    Marker palette = Marker.builder().colorScale(Marker.Palette.JET).build();
+    assertTrue(palette.asJavascript().contains("colorscale: 'Jet'"), palette.asJavascript());
+    Marker x = Marker.builder().addColorScale(0, 255, 0, 0).addColorScale(1, 0, 255, 0).build();
+    assertTrue(
+        x.asJavascript()
+            .contains(
+                "colorscale: [\n"
```
benmccann: this test seems a little brittle since it's sensitive to formatting

jsplot/src/main/java/tech/tablesaw/plotly/components/Marker.java
```diff
@@ -155,6 +163,39 @@ public String asJavascript() {
...
+    public Color(int r, int g, int b) {
+      this.r = r;
+      this.g = g;
+      this.b = b;
+    }
+  }
+
+  private static String asColorScaleString(List<Double> values, List<Color> colors) {
+    StringBuilder sb = new StringBuilder();
+    sb.append("[\n");
```
benmccann: it might be cleaner to use a JSON library like Jackson for this

jsplot/src/test/java/tech/tablesaw/components/MarkerTest.java
emilianbold: I've had this PR in my inbox for many days since these minor remarks are, to me, a mountain of work.

I wonder if it's something obvious I'm not seeing about how to make this test less brittle (there are other existing tests sensitive to formatting) or how Jackson would simplify that little method?

The way I see it, to test the output in a non-brittle way would mean to basically *parse* that generated JS. Which isn't obvious how (Jackson, you say, but note we don't have JSON, we have Javascript there) and it would make the test much more complex.

As for introducing Jackson that's a good idea, but, again, nothing easy to hook into. I have not one but *two* branches where I try to see how Jackson might lock for the JS/JSON output.

This one https://github.com/emilianbold/tablesaw/commits/emi/jackson is from late last year and I was trying to use whole objects to serialize via Jackson. Not bad, but a lot of work.

This other one https://github.com/emilianbold/tablesaw/commits/jackson is from today and I'm using the easier way of writing the `Map<String,Object>` via Jackson. This seems better, but, of course, there's a lot of work to migrate everything to Jackson.

In conclusion, I don't know how to move this PR forward. I'll make a PR for the Jackson branch but for this one I don't know how to make it better and the code is starting to become stale. I guess I should just drop the PR.

jsplot/src/test/java/tech/tablesaw/components/MarkerTest.java
```diff
@@ -16,4 +16,18 @@ public void asJavascript() {
...
+                    + "['0.0', 'rgb(255,0,0)'],\n"
```
larshelge: You might want to look into JSONAssert:

https://www.baeldung.com/jsonassert

https://github.com/skyscreamer/JSONassert

lwhite1:
> Does this project have a mailing list or some other way to talk among developers / users?

We have always just used issues and pull requests for conversations.  Are they

emilianbold:
> We have always just used issues and pull requests for conversations. Are they inadequate for some reason?

As a new committer is seemed slow to use issue / PRs as a way to ask global questions about the project. Such as, how do you usually do merges? Or, who generally "owns" which chunk and which areas need a PR from a committer and which areas would allow me to do a commit directly? Basic on-boarding questions so to say.

benmccann:
@emilianbold is this PR something that you're going to come back to?

emilianbold:
I rather forgot what was going on in this PR but from my long comment I see
there the cost/benefit are quite slim on finishing it, unless you
introduced some nicer helpers for the kind of testing(?) that was not
proper in the PR.

I do wish to resume contributing to open source next year...

Hope everything is well with you and Happy Holidays.

--emi

benmccann:
I liked the suggestion of using https://github.com/skyscreamer/JSONassert as a helper for testing. That seems quite straightforward and wouldn't require you to make many changes besides just including and calling the library

Hope all is well with you and happy holidays to you too!

frankwondon:

Issue #696, opened 2019-10-27
Table transpose()?


jackie-h:
I'm looking for a way to transpose a table. Similar to Panda's DataFrame.transpose().
Is there currently an easy way to do this?

benmccann:
I'm not aware of a method for it. We'd be open to a PR :-)

lwhite1:
May not be what you’re looking for but take a look at the support for
pivoting.  IIRC there’s a method on table called pivot.

lwhite1:
Having looked at pivot, it doesn't do simple transposing. We should probably add transpose(). 
The difficulty, of course, is that all the resulting columns must have a single type, which should probably be the "least common denominator" type for the given table, defaulting to StringColumn (or perhaps TextColumn), if the types are disparate.

benmccann:
Closing as implemented

PR #717, opened 2019-11-10
Implement transpose #696

jackie-h:
Thanks for contributing.

## Description

Added an implementation of Transpose. It has the restriction that columns must be of the same type. This is an initial version for feedback

## Testing

Added a unit test for the feature


core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -1125,6 +1125,78 @@ public boolean hasNext() {
...
+   */
+  public Table transpose() {
+    Table transposed = Table.create(this.name);
+    Column firstColumn = this.column(0);
+
+    ColumnType[] types = this.columnTypes();
+    //If all columns are of the same type
+    long distinctColumnTypesCount = Arrays.stream(types).skip(1).distinct().count();
+
+    if (distinctColumnTypesCount == 1) {
```
benmccann: I'd reverse this check:

```
if (distinctColumnTypesCount > 1) {
      throw new IllegalArgumentException(
          "Transpose currently only works where value columns are of the same type");
}
```

Then you can remove a level of indentation from the rest of the code

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -1125,6 +1125,78 @@ public boolean hasNext() {
...
+        int row = i - 1;
+        for (int col = 1; col < this.columnCount(); col++) {
+          //Avoid boxing for primitives
+          if(ColumnType.DOUBLE == column.type()) {
+            ((DoubleColumn)column).append(this.doubleColumn(col).getDouble(row));
+          }
+          else if(ColumnType.FLOAT == column.type()) {
+            ((FloatColumn)column).append(this.floatColumn(col).getFloat(row));
+          }
+          else if(ColumnType.INTEGER == column.type()) {
```
benmccann: There's some formatting here that doesn't match the rest of the code base. If you compile with maven on the command line then it should auto-format it for you

core/src/main/java/tech/tablesaw/api/Table.java
jackie-h: Yep, agree will change

core/src/main/java/tech/tablesaw/api/Table.java
jackie-h: Have reformatted

jackie-h:
Let me know if there are any more comments here. Happy to make further changes if required

benmccann:
Thanks Jackie! This looks great. We're happy to have your contributions! I didn't look super closely at it, but it's very well tested! So I'm happy with it.

We just added a couple new committers and haven't spent a lot of time figuring out the logistics of how we want to work together to review code. I'll leave this open for the time being in case anyone else has comments. But if no one chimes in by the end of the week, feel free to ping me and I'll merge this.

ryancerf:
Thanks for contributing. A few comments.

Why are you using the first column in the table as the column labels? Can you please explain a little more of the use case? Is this documented anywhere in the javadoc?

The tests should not equate strings. We do not want the tests to rely on the current implementation of `table.print()`. It makes the tests brittle. Instead you can convert the columns to arrays and use assertArraysEqual

jackie-h:
@ryancerf for the use case, I explained some of it in the issue I raised here #696. Transpose is a common feature required when working with tables. It's an implementation of a [matrix transpose operation](https://en.wikipedia.org/wiki/Transpose) and can be found as a feature in [Excel](https://support.office.com/en-us/article/transpose-function-ed039415-ed8a-4a81-93e9-4b6dfac76027) too. I stumbled upon this as a a missing feature of Tablesaw whilst trying to port [this Python code](https://towardsdatascience.com/report-time-execution-prediction-with-keras-and-tensorflow-8c9d9a889237) to Java. 
I've been doing an exploratory exercise to see if it is possible to reproduce this Python example in Java.
You can see how it is shaping up [here](https://github.com/jackie-h/ReportTimesML/blob/master/src/main/java/test/ReportTimesML/App.java) if interested.

I'll change the tests.

jackie-h:
@benmccann - would be great to have an update on whether this looks good enough to merge. Thanks in advance.

core/src/test/java/tech/tablesaw/table/TableTransposeTest.java
```diff
@@ -0,0 +1,193 @@
...
+import tech.tablesaw.api.LongColumn;
+import tech.tablesaw.api.StringColumn;
+import tech.tablesaw.api.Table;
+
+public class TableTransposeTest {
+
+  @Test
+  void transposeDoubles() {
+    StringColumn label = StringColumn.create("label").append("row1").append("row2").append("row3");
+    DoubleColumn value = DoubleColumn.create("value1").append(1.0).append(1.1).append(1.2);
```
benmccann: It might be a little simpler to do something like 

    DoubleColumn.create("value1", new double[] {1.0, 1.1, 1.2});

core/src/test/java/tech/tablesaw/table/TableTransposeTest.java
```diff
@@ -0,0 +1,193 @@
...
+    DoubleColumn value2 = DoubleColumn.create("value2").append(2.0).append(2.1).append(2.2);
+
+    Table testTable = Table.create("Data");
```
benmccann: I think you can do

    Table.create("Data", label, value, value2)

core/src/test/java/tech/tablesaw/table/TableTransposeTest.java
```diff
@@ -0,0 +1,193 @@
...
+    testTable.addColumns(label);
+    testTable.addColumns(value);
+    testTable.addColumns(value2);
+    Table result = testTable.transpose();
+
+    assertTableContents(
```
benmccann: This is a little hard to read to me because it's essentially transposed. What if we used `assertTablesEqual` instead? (see comment above regarding existence of that method)

core/src/test/java/tech/tablesaw/TableAssertions.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.util.Arrays;
+import tech.tablesaw.api.Row;
+import tech.tablesaw.api.Table;
+
+public final class TableAssertions {
+
+  private TableAssertions() {}
+
+  public static void assertTableContents(Object[][] expected, Table actual) {
```
benmccann: It looks like we have a method `assertTablesEquals` in `SortTest` that could be moved to this class

core/src/test/java/tech/tablesaw/table/TableTransposeTest.java
jackie-h: I looked at assertTablesEqual before I implemented this, but it requires the developer to construct a Table for the expected result. That is a lot of extra boilerplate that I would like to avoid. I find this to be much clearer and quicker to use. 
Please note, we are testing transpose, so the results look transposed because they are, rather than because assertTableContents is expecting input in an illogical transposed format. For another feature I think it would look less confusing.

core/src/test/java/tech/tablesaw/table/TableTransposeTest.java
jackie-h: Agree. Have changed it throughout the test

core/src/test/java/tech/tablesaw/table/TableTransposeTest.java
jackie-h: Agree. Have changed it throughout the test

core/src/test/java/tech/tablesaw/TableAssertions.java
jackie-h: I suggest that would be best tracked on a new issue as it's a separate refactoring

benmccann:
@jackie-h thanks. I think this is really close. However, something happened and the PR is now showing a ton of unrelated commits, which makes it not really possible to do a final review and merge. Is there some way you could fix that?

jackie-h:
Yep, sorry about, I noticed too and am in progress of trying to fix

jackie-h:
@benmccann  fixed the PR. Please take another look

jackie-h:
@benmccann PR is fixed now, please take a look

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -1135,6 +1135,70 @@ public boolean hasNext() {
...
+   * label | row1 | row2 | row3 |<br>
+   * -------------------------------------<br>
+   * value1 | 1 | 1.1 | 1.2 |<br>
+   * value2 | 2 | 2.1 | 2.2 |<br>
+   *
+   * @return The transposed table
+   */
+  public Table transpose() {
+    Table transposed = Table.create(this.name);
+    Column firstColumn = this.column(0);
```
benmccann: This probably won't work on an empty table. It'd be nice to handle that case a bit better by just returning `transposed` first if there are no columns

Also, this variable isn't used until awhile later. It'd be nicer to declare it right before it's used. Or actually I'd probably just inline it:

```
for (String label : this.column(0).asStringColumn()) {
```

core/src/test/java/tech/tablesaw/table/TableTransposeTest.java
```diff
@@ -0,0 +1,167 @@
...
+    StringColumn label = StringColumn.create("label", new String[]{ "row1", "row2", "row3"});
+    LongColumn value = LongColumn.create("value1", new long[]{ 1, 2, 3 });
+    LongColumn value2 = LongColumn.create("value2", new long[] { 4, 5, 6});
+
+    Table testTable = Table.create("Data", label, value, value2);
+    Table result = testTable.transpose();
+
+    assertTableContents(
+        new Object[][] {
+            {"label", "row1", "row2", "row3"},
```
benmccann: I would not expect the column names to end up in the contents

jackie-h: This is the correct implementation of transpose in my view. Everything is kept, even column names, but flipped.  The first row (the headings) becomes the first column. 
This is consistent with Python pandas and Excel
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.transpose.html
If someone has a use case where they want to get rid of that first column, they can drop it in a subsequent step

jackie-h:
@benmccann The code has been like that since the beginning of this PR, it would be a much nicer experience contributing if all comments were given upfront. Please confirm that these are the last comments you have and that you will merge this change. My time to revisit this relatively small change is not infinite so I'd like be able to close this out in one final go

core/src/main/java/tech/tablesaw/api/Table.java
jackie-h: Good catch. I've added a test for the empty table and made the change

core/src/test/java/tech/tablesaw/table/TableTransposeTest.java
```diff
@@ -0,0 +1,175 @@
...
+            {"value2", 2.0, 2.1, 2.2},
+        },
+        result);
+
+    assertEquals(testTable.print(), result.transpose().print());
+  }
+
+  @Test
+  void transposeFloats() {
+    StringColumn label = StringColumn.create("label", new String[]{ "row1", "row2", "row3"});
```
benmccann: the formatting is inconsistent between this line and the next. we have an autoformatting plugin so that you don't have to worry about formatting. if you're rebased against the latest master and run `mvn compile` once on the command line it will install a git pre-commit hook that will take care of any formatting on commit

core/src/test/java/tech/tablesaw/TableAssertions.java
```diff
@@ -0,0 +1,49 @@
...
+
+    if (expected.length == 0) {
+      //empty table
+      assertEquals(0, actual.rowCount(), "Expected an empty table");
+      assertEquals(0, actual.columnCount(), "Expected a table with no columns");
+    } else {
+      assertEquals(
+          expected.length - 1, actual.rowCount(), "Table does not have the same number of rows");
+
+      assertArrayEquals(
```
benmccann: This would be really confusing if you called this method thinking that it was just asserting the table contents were equal to the given array

core/src/test/java/tech/tablesaw/table/TableTransposeTest.java
benmccann: I don't think it really is consistent with Pandas. Pandas doesn't move the labels into the dataframe contents. Instead it has an additional concept of row names and swaps the column and row names.

I think we could also go the other way and say if someone has a usecase where they want the labels to be in the table contents they can add it in a subsequent step. It'd be slightly more code, but not too much more and I think it'd be much more expected. It looks like @ryancerf was also confused by this https://github.com/jtablesaw/tablesaw/pull/717#issuecomment-552737591.

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -1135,6 +1135,73 @@ public boolean hasNext() {
...
+    for (int i = 1; i < this.columnCount(); i++) {
+      Column<?> columnToTranspose = this.column(i);
+      labelColumn.append(columnToTranspose.name());
+    }
+
+    for (int i = 1; i < transposed.columnCount(); i++) {
+      Column<?> column = transposed.column(i);
+      int row = i - 1;
+      for (int col = 1; col < this.columnCount(); col++) {
+        // Avoid boxing for primitives
```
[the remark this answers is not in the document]
jackie-h: I considered doing it that way, but I think it will be very inefficient memory/performance wise.
Transposing a whole table is a big operation, so I think that matters.

core/src/test/java/tech/tablesaw/table/TableTransposeTest.java
jackie-h: FYI - The auto-formatting plugin doesn't work out of the box anymore. It seems to be since this change here to switch formatters:
https://github.com/jtablesaw/tablesaw/commit/646fbd494c75ca70f33b87157acfdd0a4cdf7b68
It fails locally with this error:
[ERROR] Failed to execute goal com.cosium.code:maven-git-code-format:1.38:install-hooks (install-formatter-hook) on project tablesaw-core: java.io.IOException: Cannot run program "/Applications/IntelliJ IDEA CE.app/Contents/plugins/maven/lib/maven3/bin/mvn": error=13, Permission denied -> [Help 1]
That is why I couldn't run it. The previous one did work seamlessly. Some googling suggests that I may need to install maven separately but I haven't go to the bottom of that yet. I'm using IntelliJ with its bundled maven.

benmccann: Ah. I made that switch to help Intellij users, but looks like it's hard to find a good solution :-p  The previous one only ran if you ran `mvn`, which a lot of people never did because they just used the IDE

core/src/main/java/tech/tablesaw/api/Table.java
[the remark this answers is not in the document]
benmccann: I'm not following. Why would it use more memory or have worse performance?

core/src/test/java/tech/tablesaw/TableAssertions.java
jackie-h: Ok, it seems we have different mental pictures of what "contents" means. To me "contents" is "everything the table contains", including the headers, so I'd expect contents to include the headers. I would have called it assertRowValues if it was just the rows. 
I'll try to find a name that is less prone to misunderstandings and perhaps separate out the header assertions.

core/src/test/java/tech/tablesaw/table/TableTransposeTest.java
benmccann: Using `assertTablesEqual` is actually fewer lines of code because you no longer need to separately check the table name:

```
    assertTablesEquals(
        Table.create(testTable.name()
            DoubleColumn.create("value1", new double[] {1.0, 1.1, 1.2}),
            DoubleColumn.create("value2", new double[] {2.0, 2.1, 2.2})
        ),
        result);
```

jackie-h: It is more boilerplate though - more tokens, doing all the additional Column.create()/Table.create()'s drove me crazy. The version without that is much easier to eyeball as you scan through the test, as it looks much more like a CSV. I find it is more intuitive to see the headers in a row at the top, as they would be if the table was displayed.
More fundamentally, in my view that is a more brittle test, as it is using the class we are trying to test (Table) for the expected values, and if the internals of the class break we are more susceptible to the test still telling us everything is fine mistakenly. For instance, I wouldn't construct an ArrayList of expected values to check the contents of an ArrayList.

jackie-h:
Closed by accident, due to bad git foo.

core/src/test/java/tech/tablesaw/table/TableTransposeTest.java
```diff
@@ -0,0 +1,175 @@
...
+    assertTableContents( new Object[][]{}, result);
+  }
+
+  @Test
+  void transposeDoubles() {
+    StringColumn label = StringColumn.create("label", new String[]{ "row1", "row2", "row3"});
+    DoubleColumn value = DoubleColumn.create("value1", new double[]{ 1.0, 1.1, 1.2 });
+    DoubleColumn value2 = DoubleColumn.create("value2", new double[]{ 2.0, 2.1, 2.2});
+
+    Table testTable = Table.create("Data", label, value, value2);
```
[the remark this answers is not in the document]
jackie-h: Ok. Have changed it

core/src/test/java/tech/tablesaw/TableAssertions.java
jackie-h: Took care of this here #743

core/src/main/java/tech/tablesaw/api/Table.java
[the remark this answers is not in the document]
jackie-h: Having looked at this more closely, I take back that comment. I'd tried to do something previously using existing methods but it would have resized the data structure inside the columns when it appended, which would be inefficient.

Do you already have other cases in mind that could reuse that method ? I'd typically take the view that until it is needed elsewhere it shouldn't be on the public API, but I do agree it might be useful there.

I'd suggest that it might be better to name it appendFrom.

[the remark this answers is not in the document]
benmccann: That's fair. I'm fine leaving this internal for now. We do something somewhat similar in `DataFrameJoiner`, but it looks different enough that it couldn't easily be shared

core/src/test/java/tech/tablesaw/TableAssertions.java
```diff
@@ -27,4 +30,48 @@ public static void assertTablesEquals(Table compareWith, Table sortedTable) {
...
     }
   }
+
+  public static void assertTableEquals(
+      String[] expectedHeaders, Object[][] expectedRowValues, Table actual) {
+    assertColumnNamesEquals(expectedHeaders, actual);
+    assertRowValuesEquals(expectedRowValues, actual);
+  }
+
+  public static void assertColumnNamesEquals(String[] expectedColumnNames, Table actual) {
```
benmccann: It looks like these methods aren't used outside of this class, so you could make them private. Also, it'd be nice to check that the table names are equal I think

benmccann:
The one thing I'm still really stuck up on is swapping the column names with the first column. I took a look at your app and see how that makes sense for your use case. I'm concerned that it's too specialized for your use case though.

There's a lot of other cases where it wouldn't make very much sense. E.g. imagine you have table that's entirely populated by doubles. Now you transpose it. You will lose the first column from the table contents because it will become column labels. And it will be really weird to have doubles as column labels instead of just doing something like "Col 1", "Col 2", etc.

We potentially could consider adding a row labels feature like Pandas has. I would be open to that as a feature. Then it would work both as you proposed and as I was expecting. I'm somewhat curious what @lwhite1 would think about that as the original library author since it's a somewhat core change.

jackie-h:
I think it is important in many use cases that transpose() is a reversible operation, it shouldn't lose information. Dropping the column labels loses information. I'd expect it to respect that mathematical property and to be able to call transpose() again to reverse the operation and be back where you started. 
For practical purposes though, I can see that it is sometimes desirable and sometimes not, depending on what data is in the table. My view is that it is very easy to drop the column afterwards if you don't need it (a single op) but harder to transpose and add it if it isn't there to start with.

I've refactored to provide both options here which is hopefully a better compromise. I think the key thing is what should the default be. I've gone in your direction there but I think the default really should be reversible (the opposite), but perhaps row labels are a much better way to accomplish that at a later stage.

My use case was basically very similar to calling the new implementation Table.summary() and then calling transpose(). It makes a lot of sense there because the first column is a set of String labels that nicely end up as column headings.
 
For your table that is entirely doubles, that is a good point, perhaps more options are required, or perhaps a transposeRowData() or similar that is more explicitly only transposing the values.
Another approach would be to be more nuanced, and check if the first column is a String column or not inside transpose.
I'll think about how best to make that use case easy and most likely make further changes here.

core/src/test/java/tech/tablesaw/TableAssertions.java
jackie-h: I've changed my approach here, as I think I see what you were getting at initially. The initial Table is created using a column-wise approach - i.e. Column.create(), so if eyeballing both the old and new tables, it was hard to see the difference. 
I've added a convenience method that allows comparing in that column-wise fashion but without forcing the person to create the expected results as a Table. 
I think it would be easier to visualise the tables in the test if both were the other way around (row based construction) but I do agree that it is better to be consistent. 
Let me know what you think about this.

benmccann:
Cool. Thanks for updating the `transpose` method. That seems much more amenable to me.

For comparing the tables, I think it could be a little hard to tell when you're reading the code whether the `Object[][]` is supposed to be columns or rows. Would doing the following reduce boilerplate enough do you think? You could do it for the initial table creation too if you also thought that was less boilerplate
```
    Table.read().string(
        "label, row1, row2, row3\n" +
        "value1, 1.0, 1.1, 1.2\n" +
        "value2, 2.0, 2.1, 2.2\n", "csv")
```

You'd raised a concern that we wouldn't want to use `Table` to test `Table`, but I take a finer-grained view of it. We're just testing `Table.transpose` here. I'm not worried about whether getting a value from a cell in the table works correctly because those have their own tests, so we would know if something broke there.

Thanks for sticking in there on this one. I'll try to cut a new release for you not too long after merging this (though I might wait for another PR or two if they seem close enough)

jackie-h:
On the `Table.read().string()` suggestion - I'd far prefer a `Table.create(Object[][])` method that lets me do that as an array. I'd prefer not to have to pass something in that has to be parsed later (csv). 
I can see that both of these might struggle with typing the columns appropriately though.
Perhaps it would be nice to have convenience methods on Table for creating with more specific types, such as `Table.create(String name, String[] column labels, double[][] values)`,  for when the values are all doubles.

benmccann:
Here's another idea. Right now you have to do `DoubleColumn.create("value1", new double[] {1.0, 1.1, 1.2})`. If we changed `DoubleColumn.create` to take `double... arr` instead of `double[] arr` then you could instead do `DoubleColumn.create("value1", 1.0, 1.1, 1.2)`. The one place you'd get in trouble is creating an `IntColumn` with one element because you'd end up specifying the initial size, but that tradeoff might be worth it for the conciseness is would enable everywhere else.

> I'd far prefer a Table.create(Object[][]) method that lets me do that as an array

Unfortunately, it'd be a lot of work because you'd have to figure out the types. The `Table.read().string()` method for parsing a CSV already handles this which is why I suggested it, but I agree with you it's a little brittle and not type safe. It might be possible to refactor `ColumnTypeDetector` to work with `Object` in addition to `String`, but I'm not sure it's worth it.

> Perhaps it would be nice to have convenience methods on Table for creating with more specific types, such as Table.create(String name, String[] column labels, double[][] values), for when the values are all doubles.

It seems a little specific. You'd have to create lots of them for all the different types and still wouldn't handle mixed types. Plus, I still worry that you couldn't easily tell from looking at the code whether it was an array of columns or array of rows

jackie-h:
> If we changed DoubleColumn.create to take double... arr instead of double[] arr then you could instead do DoubleColumn.create("value1", 1.0, 1.1, 1.2). 

I like that idea. I tried it locally in the test and it looks a lot cleaner.

benmccann:
Cool! Glad we figured something out! Will you implement that?

jackie-h:
The good thing about allowing a TableAssertions.assertTablesEquals() method that takes the Object[][] is that you don't have those problems with having to do the type conversion. 
On the comment 
> I think it could be a little hard to tell when you're reading the code whether the Object[][] is supposed to be columns or rows
I think that having really clear naming of the method parameter and docs could avoid that, as if you are unsure, you can read it.

core/src/test/java/tech/tablesaw/table/TableTransposeTest.java
jackie-h: In case others see this, using a separate install of Maven worked here.
 
FWIW - I'm not sure I like the fact that it runs Maven now on every commit as Maven is a bit slow. I'd probably prefer instructions for contributors for how to set-up IntelliJ's formatting to match Google's so it just happens with IntelliJ magic. I appreciate that not everyone will do that though which may create more work for reviewers. Perhaps the CI can police that though, I've seen that done on other projects.

core/src/test/java/tech/tablesaw/table/TableTransposeTest.java
```diff
@@ -0,0 +1,288 @@
...
+  @Test
+  void transposeFloats() {
+    Table testTable =
+        Table.create(
+            TABLE_NAME,
+            StringColumn.create("label", new String[] {"row1", "row2", "row3"}),
+            FloatColumn.create("value1", new float[] {1.0f, 1.1f, 1.2f}),
+            FloatColumn.create("value2", new float[] {2.0f, 2.1f, 2.2f}));
+    assertEquals(
+        testTable.print(),
```
benmccann: Can we call `assertTableEquals` instead of checking the output of `print`?

core/src/test/java/tech/tablesaw/table/TableTransposeTest.java
```diff
@@ -0,0 +1,288 @@
...
+        new Object[][] {
+          {"0", 1.0, 2.0},
+          {"1", DoubleColumnType.missingValueIndicator(), 2.1},
+          {"2", 1.2, 2.2},
+        },
+        result);
+  }
+
+  @Test
+  void transposeFloats() {
```
benmccann: The test name doesn't really match what it's testing. It seems like it's testing that you can `transpose` twice and get back the original value

core/src/test/java/tech/tablesaw/table/TableTransposeTest.java
jackie-h: They are floating point numbers so equality is tricky, they need to be compared within a tolerance.
Print does some formatting which helps avoid that which is why this ended up like this.

benmccann:
Hey @jackie-h just checking in on this... Were you planning to implement the varargs change? No worries if you just haven't had time. I just wanted to make sure I wasn't holding you up

jackie-h:
> Hey @jackie-h just checking in on this... Were you planning to implement the varargs change? No worries if you just haven't had time. I just wanted to make sure I wasn't holding you up

I think it is a good change to make, but I wasn't planning to pick it up, as I've run out of free hours I can really put into this and didn't want to rebase yet again. My plan was to aim to close this out without it. If I have time at a later date perhaps I could pick it up. 
I do still prefer my assertEquals() which has the Object[][] but I understand that your not keen, so to resolve that I moved it out of the main TableAssertions class so that it can be local to the test only.

Any tips on how to run the NPath complexity locally to understand the last failure ? I have been running SonarLint locally. That showed the complexity issues, but they were shown as resolved when I removed the large if/else block and replaced it with append.

core/src/test/java/tech/tablesaw/table/TableTransposeTest.java
jackie-h: Have changed the assertions to address this

core/src/test/java/tech/tablesaw/table/TableTransposeTest.java
jackie-h: Have changed assertions to address this and removed the prints. I pulled to float values out into variables, which should allow them to be compared safely.

jackie-h:
You can ignore the last part of that comment. I installed the PMD plugin for intellij and it seems to be reporting the same error for NPath, so I'm sorted.

core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -1135,6 +1136,118 @@ public boolean hasNext() {
...
+
+  private Table transpose(
+      Table transposed,
+      ColumnType resultColumnType,
+      IntFunction<String> columnNameExtractor,
+      int startingColumn) {
+
+    for (int row = 0; row < this.rowCount(); row++) {
+      String columnName = columnNameExtractor.apply(row);
+      Column column = resultColumnType.create(columnName);
```
benmccann: This will result in a compilation warning. You should either address or suppress

jackie-h:
FYI - The Quality Review results seem incorrect here. They seem to be failing off an older version. They are highlighting that transpose() is on line 1179, but it isn't on that line anymore.
There was a high NPath complexity > 200, but I've verified with the PMD plugin locally that in the latest refactor it no longer exists.

jackie-h:
> I think it's much harder to follow through five different methods than a single method. I would just revert this commit honestly

This does make the default path transpose() much easier to debug, as it skips the extra complexity entirely. 
There was a high NPath complexity > 200, which seemed like a fair criticism (booleans are hard to follow), so I refactored it out.
If you don't care about the NPath complexity, it would be better not to include it in those checks, or to have commented earlier that I could ignore it. I've just spent 2 hours trying to make this simpler and am now quite happy with the result.

jackie-h:
> Is there any reason you changed "Transpose" to "This operation"? I thought it was more explicit before

I suspected we would have other uses for this method inside Table, so I made it more general

jackie-h:
> I went ahead and pushed the change to `create`. Can you update the tests to use it now?

As I mentioned above, I really didn't want to have to rebase and merge again and revisit the tests again, which is why I didn't do that change. Are you now putting that refactoring as a blocker to accepting this?

benmccann:
I've been pretty consistent in saying that we should compare tables in the tests. I think there's been more work put into trying to avoid that than just doing it at this point. I really don't think it'd take long based on the last version of the code I'd looked at. But now the problem is that the code has changed so much since then.

I'm really a bit frustrated that the code is changing so much so quickly without prior discussion. It makes it really hard to review. I think this PR is substantially further away from being in a mergable state now. Can you drop the last 7 commits from this PR and go back to where we were before?

core/src/main/java/tech/tablesaw/api/Table.java
jackie-h: I had to remove the <?> so that I could make use of append(). 
This issue could be fixed by fixing the generics on the ColumnType class. ColumnType.create() shouldn't really be returning <?>, instead it could be typed to return <T> as the actual column type. 
I started looking at that over the weekend, but backed out of it because it was going to be quite a big refactor.

jackie-h:
> I've been pretty consistent in saying that we should compare tables in the tests. I think there's been more work put into trying to avoid that than just doing it at this point. I really don't think it'd take long based on the last version of the code I'd looked at. But now the problem is that the code has changed so much since then.
> 

Whilst that varargs change addressed part of my concerns about boilerplate, it still leaves all the excess boilerplate of the Column.create()'s. I've also been consistent in my views that comparing the Object[][] is easier for testing and as it saves boilerplate. I respected your earlier wishes by moving it into the private class so no-one else will reuse it, it wasn't much rework, just a move.
That seemed to be a reasonable compromise to me, especially given this codebase as whole isn't generally consistently comparing Tables in the testing. 

> I'm really a bit frustrated that the code is changing so much so quickly without prior discussion. It makes it really hard to review. I think this PR is substantially further away from being in a mergable state now. Can you drop the last 7 commits from this PR and go back to where we were before?

That doesn't seem fair, it's been 4 days until I had time to work on this again, and I only pushed 2 commits today. 
To explain-
I made the first commit to address the review comment about the float test.
My other commit was to reduce the NPath complexity, which had been flagged as an issue, with a big red cross on the pull request. It moves code around into methods, but it isn't a difference in logic, and it could only be done in one go to reach the goal of reduced NPath complexity.
I'd concluded that was why it wasn't merged over the weekend (the test feedback seemed resolved to me, as I'd moved it out of the TableAssertions class and made it private to compromise. I didn't expect that to be a showstopper).
I do think code quality is important, especially complexity so it seemed fair to look at fixing it. 
I'm not sure how I was supposed to guess that it was not important to having this accepted. A comment on the weekend to that effect would have helped. I can revert that part to match your preference now that I know. 

FWIW -  I do think there is some merit here to that refactor in the improved understandability of the default path as it skips the chunk , and it opens up a possibility in future if someone wants to supply their own naming convention for the columns. I can see both sides to that though.

benmccann:
Don't worry about it. I think it's close enough I'll just finish up the cleanup and merge it. I was mainly getting frustrated by how much back and forth this was taking. I know you are too. Hopefully we can make this process smoother in the future. I'm considering replacing Codacy with Sonar. I have Sonar running against PRs now, but I guess it didn't run against this one because it was open awhile ago.

Thanks again for all your help and sorry this wasn't the smoothest process. I hope you'll contribute more in the future and that you'll share how you used this afterwards. I'm very interested to see your project

jackie-h:
OK - I'm happy to revert the last commit if that helps (NPath refactor), that should be back closer to what you were expecting

benmccann:
Merged here: https://github.com/jtablesaw/tablesaw/commit/1b01eaf5c94c8a51d09be7fe2c080a78dc9a03e1

Thanks again for the new feature

benmccann:
I've released 0.37.1 so that you can use the new features: https://github.com/jtablesaw/tablesaw/releases/tag/v0.37.1

Issue #707, opened 2019-11-01
Suggestion: Table.summary() should ideally return a Table


jackie-h:
It would be nice to have a feature on the Table that mirrors the Pandas Dataframe.describe() feature.
Column.summary() is the equivalent on a column, but it would be nice if Table.summary() brought the column summaries together into a table that can then be used as a Table (similar to Dataframe.describe()).
Today Table.summary() returns a string, but instead you could do Table.summary().print() to get the String version and keep that feature. 
I'm looking for feedback on this suggestion. If it sounds like a good one, I might attempt the pull request.
For context - I am translating some Python ML code into Java, and looking for equivalent methods

benmccann:
That sounds reasonable to me. @lwhite1 is more familiar with the `summary` method than I am, so I'd probably want his feedback as well

benmccann:
Closing as implemented. I really like this change!

PR #718, opened 2019-11-10
Change Relation.summary() to return a Table #707

jackie-h:
Thanks for contributing.

- [ x ] Tick to sign-off your agreement to the [Developer Certificate of Origin (DCO) 1.1](https://developercertificate.org)

## Description

Changed Relation.summary() method to return a Table instead of a String. 
This is especially useful for numerical tables, where you can then see all the statistics in one table. This is similar to describe() on the Pandas DataFrame.
There are a few options for handling different column types, here I've merged into one table, but we could also return a List<Table> and summarise by type. The test for StringColumn's with DoubleColumn's shows the challenge here

## Testing

Added two unit tests


emilianbold:
Why not add a new method called `describe` and leave the existing `summary` in place?

core/src/main/java/tech/tablesaw/table/Relation.java
```diff
@@ -240,19 +240,55 @@ public Table structure() {
...
-    StringBuilder builder = new StringBuilder();
-    builder
-        .append(System.lineSeparator())
-        .append("Table summary for: ")
-        .append(name())
-        .append(System.lineSeparator());
-    for (Column<?> column : columns()) {
-      builder.append(column.summary().print());
-      builder.append(System.lineSeparator());
+  /*public Table summary() {
```
benmccann: We prefer to delete code rather than comment it out because it can confuse other people who come later and wonder why it's here and commented out

core/src/test/java/tech/tablesaw/table/TableSummaryTest.java
```diff
@@ -0,0 +1,68 @@
...
+    testTable.addColumns(label);
+    testTable.addColumns(value);
+    testTable.addColumns(label2);
+    testTable.addColumns(value2);
+    Table result = testTable.summary();
+    assertEquals(
+        "                                       Data                                       \n"
+            + " Summary   |  label  |         value1         |  label2  |        value2         |\n"
+            + "----------------------------------------------------------------------------------\n"
+            + "     row1  |      1  |                        |       1  |                       |\n"
```
benmccann: I like the statistics! That's very cool!

I'm not sure about "row1", "row2", etc. It seems to me that it could make the table quite messy. It also doesn't appear that Pandas includes it in their output, so I think could also be unexpected or confusing for users coming from Python

core/src/main/java/tech/tablesaw/table/Relation.java
jackie-h: @emilianbold My general view is that it is better to have consistent API's across the codebase. The summary() API on column returns a Table so it feels more natural to me to do the same here. It is easy enough to get the String representation from that by calling the existing print() methods.
If the preference is to rename all the summary() methods to describe(), including on the columns, I'd support that too.
Personally I don't find the current String representation of summary() useful, as it seems to lazily dump column stats individually which is somewhat hard to read.
I'm happy to defer to the maintainers here though, @benmccann,  @lwhite1 let me know which you prefer. 

@benmccann Agree, was not my intent to leave in commented out code. I'll fix, apologies.

core/src/test/java/tech/tablesaw/table/TableSummaryTest.java
jackie-h: Thanks! I just checked the behaviour in pandas and it seems to be like this:

"Use the parameter, DataFrame.describe(include = 'all') to get a summary of all the columns when the dataframe has mixed column types. The default behavior is to only provide a summary for the numerical columns."

[DataFrame.describe()](https://pandas.pydata.org/pandas-docs/version/0.17.1/generated/pandas.DataFrame.describe.html)

Perhaps that is a better idea, to add a few overloaded methods that allow you to ask for summaries of specific column types, and to mirror that default behaviour.

benmccann: Even with `include = 'all'` the output includes only summary statistics:

```
>>> df.describe(include='all')
        categorical  numeric object
count            3      3.0      3
unique           3      NaN      3
top              f      NaN      c
freq             1      NaN      1
mean           NaN      2.0    NaN
std            NaN      1.0    NaN
min            NaN      1.0    NaN
25%            NaN      1.5    NaN
50%            NaN      2.0    NaN
75%            NaN      2.5    NaN
max            NaN      3.0    NaN
```

core/src/main/java/tech/tablesaw/table/Relation.java
emilianbold: You have to think a bit about backwards compatibility too. Existing code might be out there that assumes the return type is string. That code will not compile anymore. That was my angle against changing the return type so easily.

benmccann: We've never been particularly strict about backwards compatibility since the library is pre-1.0. And I imagine in this case that if you do something like `System.out.println(table.summary());` that it won't really matter that the return type changed

My main question was whether we'd also want a row in the summary table that prints the column type since the old output included and it could be useful in some circumstances

emilianbold: > We've never been particularly strict about backwards compatibility since the library is pre-1.0.

Good to know. This allows more flexibility in other parts too.

core/src/test/java/tech/tablesaw/table/TableSummaryTest.java
jackie-h: Good point, those statistics seem to be a better representation of what a summary should be for a StringColumn, so I'd suggest we change the StringColumn.summary() method to return statistics. This is a breaking change but if someone was relying on it, they can easily switch to use StringColumn.countByCategory() directly

jackie-h: I've tried that out in this latest set of changes. Once we finalize they feature I'll also update the tests so the don't assert on .print() and assert values

jackie-h: We could also introduce a new method statistics() on Relation and Column and deprecate the summary() methods if that seems clearer.

core/src/test/java/tech/tablesaw/table/TableSummaryTest.java
```diff
@@ -0,0 +1,66 @@
...
+            + " Std. Dev  |   0.09999999999999998  |   0.1000000000000002  |",
+        result.print());
+  }
+
+  @Test
+  public void summaryMixedTypes() {
+    StringColumn label =
+        StringColumn.create("label").append("yellow").append("green").append("blue");
+    DoubleColumn value = DoubleColumn.create("value1").append(1.0).append(1.1).append(1.2);
+    StringColumn label2 = StringColumn.create("label2").append("red").append("red").append("red");
```
benmccann: Maybe instead of having a second string and double column we could show some other types like `BooleanColumn` and `DateColumn`?

jackie-h:
I've decided to break this up into a few smaller PR's. First one is here #742

benmccann:
@jackie-h the other PR looks good. just one minor comment on it. It looks like lots of extra commits have ended up on this PR as well

benmccann:
I'll take another look at this one after you rebase it against https://github.com/jtablesaw/tablesaw/pull/742

core/src/test/java/tech/tablesaw/table/TableSummaryTest.java
```diff
@@ -0,0 +1,73 @@
...
+            DoubleColumn.create("value1", new double[] {1.0, 1.1, 1.2}),
+            BooleanColumn.create("truthy", new boolean[] {true, false, true}),
+            DateColumn.create(
+                "dates",
+                new LocalDate[] {
+                  LocalDate.of(2001, 1, 1), LocalDate.of(2002, 1, 1), LocalDate.of(2001, 1, 1)
+                }));
+    Table result = testTable.summary();
+    assertEquals(
+        "                                   Data                                    \n"
```
benmccann: it's a little messy that the output is so different for each column type. e.g. why "Missing" only for dates and why not "Count" for boolean column?  this is a good start though

benmccann:
Thanks again for this!!

PR #720, opened 2019-11-20
Adding support for reading Apache ORC file format

### LINKED ISSUE #620: Consider supporting Orc as a file format
Orc is an apache column store (C++) focused on Hadoop, but has a Java library that can be used for the local file system.

Rumor has it, that it is better documented than Parquet (see #221). Parquet is also a column store focused on Hadoop, from Apache, but written in Java. Unfortunately, using it on a local file system has not been a focus, and seems to be possible, but not to be well documented. It would be nice to support one or the other.

murtuza-ranapur:
Thanks for contributing.

## Description

1. Added support for reading Apache ORC file format 
 Fixes #620 

## Testing

Unit Test cases added

## Note
Will raise another PR for writing Table to ORC file format as soon as i am done


murtuza-ranapur:
There are no switch cases in the code without breaks (Other than the fall through one), still the code review is failing. Can anyone help?

PR #724, opened 2019-11-22
Added support for reading and writing Apache ORC file format

### LINKED ISSUE #620: Consider supporting Orc as a file format

murtuza-ranapur:
Thanks for contributing.

## Description

1. Added support for reading and writing Apache ORC file format 
 Fixes #620 

## Testing

Unit Test cases added


benmccann:
Just FYI, you can update an existing pull request and don't need to open a new one

murtuza-ranapur:
@benmccann, noted. Actually its my first contribution, at least in terms of code. I was really looking for some help here, actually i am pretty sure that i have address all 3 issues raised by Codacy. I have made use temporary variable instead of reassigning incoming parameter and have added breaks even in default ones. Still the review is failing for the same point. I am bit confused here as to what to change in the code to get it cleared by Codacy.

benmccann:
Some of the formatting doesn't match what we do elsewhere. Can you run `mvn compile` on the command line? It will automatically reformat your code

murtuza-ranapur:
@benmccann did the formatting. Something strange is going on here though, PR review is still failing for a switch case which longer exist in the code. Am i missing some step here?

benmccann:
I'll close and reopen this PR to see if it gets Codacy to pick up the most recent code. If it doesn't then let's just ignore Codacy

murtuza-ranapur:
Its still the same. Failing for the same point and some of the lines which don't exist anymore.

benmccann:
Don't worry about Codacy then.

I'm curious if you have any plans to add an `OrcWriter`?

I didn't look super closely at this PR, but it looks fairly good to me. My main question would be whether we want to include this in the core project or make it a separate project (possibly under the jtablesaw org)

murtuza-ranapur:
Yes, i will soon add the writer code. Actually i had thought of raising another PR for the same, so that i could inculcate the inputs that i would received for the read operations(Regarding type selection and design choices), and avoid any major rework. But now that i have some positive feedback i ll just go ahead and add the same in this PR only.

murtuza-ranapur:
Added `OrcWriter`

murtuza-ranapur:
Hello @benmccann one question, why doesn't `Destination` class have `File` property? Orc requires its own Writer to write the data, which takes a file path as argument(ORC even has its own `Path` class). So only valid input to OrcWriter is a builder with a output file path. And though `Destination` has constructor which take `File`, its doesn't store file object unlike the `Source` class. 

I my case `OrcWriter::write(Table table, Destination dest)` function will fail in all cases as `Destination` doesn't provide output file Path. Would it be okay to add `File` property in `Destination` class?

Currently for Orc only working method  to write is `OrcWriter::write(Table table, OrcWriteOptions options)`.

benmccann:
I suppose it'd be okay to add `File` to `Destination`. I don't recall why `Destination` and `Source` might be different except that maybe it hasn't been necessary until now

murtuza-ranapur:
I have added a new constructor in `Destination` which takes file path as input and sets `File` property. Though it did help me make `OrcWriter::write(Table table, Destination dest)` work as users can set the file in destination and use it. But now there are two constructors in `Destination` one which takes `File` argument and creates `FileOutputStream` and one which takes `String` path and just sets the `File`. I know it has become more confusing now as both constructor exhibits completely different behavior. I had though of removing `FileOutputStream` construction portion from `File` parameter constructor but that would have broken a lot of functionalities. Some shortcut functions in `DataFrameWriter` such as `toFile` would have completely stopped working. 

I think according to current structure an ideal `DataWriter` implementation is the one which handles writing of the `Table` by itself using `Writer`, which is true for html and Json, but implementation which depends on third party API to write (Such as ORC and even Excel, though its writer part is absent) may only work with some specific configuration.

I understand the purpose of `toFile`, `toWriter`, `toString` etc. functions in the `DataFrameWriter`, is to configure data writer with some sensible defaults but i still think moving strictly to Builder based implementation would give more flexibility.

Let me know your thoughts.

lwhite1:
@benmccann Is there a limit to the number of maven sub-projects? I feel like I've seen some issues in the past like that.

benmccann:
I'm not aware of a limit to the number of maven sub-projects, but am not a heavy Maven user.

I think my main question with this PR is whether it's something that should live in this repo (other than the changes to `Destination`). We'd be committing to maintaining it and are already stretched pretty thin and I don't have any experience with Orc. It's pretty well-separated as a separate Maven project, so there's no really compelling reason to make it part of this repo except for discoverability, but I think we could solve that just as well by including it in the documentation. So I'd lean towards taking the changes to `Destination` and then adding some docs pointing to a new repo containing this plugin.

lwhite1:
> I think my main question with this PR is whether it's something that should live in this repo (other than the changes to `Destination`). We'd be committing to maintaining it and are already stretched pretty thin and I don't have any experience with Orc.

If we're saying we should put it in a different repo so we can ignore it, we should probably just not add it at all.  Someone will want changes or fixes at some point regardless of where it lives.

benmccann:
To clarify, I don't think we should add new repos under the jtablesaw org, but should let the maintainers create them. Anyone can create new repos with additional reader/writer plugins in their own personal repos or repos in their organization and they'll be the ones responsible for maintaining. We can add to the readme or docs to help with discoverability though

lwhite1:
closing.

Issue #728, opened 2019-11-28
add method unique(String columnName)


J4nJ4nsen:
a 'nice-to-have' feature would be a method on `tablesaw.api.Table`, which returns the unique values of some column given by `columnName` like it is already integrated in `tablesaw.column.Column`.

I just forked the repo and am actually working on that...

benmccann:
You can call `table.column(columnName).unique()`

PR #729, opened 2019-11-28
Table extension (regarding Issue #728)

J4nJ4nsen:
added method unique(String columnName) like that in columns.Column

Thanks for contributing.

## Description

What was changed

## Testing

Did you add a unit test?


benmccann:
It seems like it could make the `Table` interface quite large to duplicate every method from `Column` on `Table`

Also, I think it's confusing that `Table.unique` as you're proposing here works differently than `Column.unique` since it sorts the result

benmccann:
It's better just to call `table.column(columnName).unique()` than copy the column methods to the table

PR #735, opened 2020-01-08
Upgrade to Smile 2.0.0

benmccann:
Thanks for contributing.

## Description

Upgrades to Smile 2.0

## Testing

Updated the existing unit tests. Deleted a few since there are less cases to cover now

### DIFF
--- a/core/src/main/java/tech/tablesaw/conversion/smile/SmileConverter.java
+++ b/core/src/main/java/tech/tablesaw/conversion/smile/SmileConverter.java
@@ -1,16 +1,18 @@
 package tech.tablesaw.conversion.smile;
 
-import java.text.ParseException;
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
-import java.util.stream.IntStream;
-import smile.data.Attribute;
-import smile.data.AttributeDataset;
-import smile.data.NominalAttribute;
-import smile.data.NumericAttribute;
+import smile.data.DataFrame;
+import smile.data.Tuple;
+import smile.data.type.DataType;
+import smile.data.type.DataTypes;
+import smile.data.type.StructField;
+import smile.data.type.StructType;
 import tech.tablesaw.api.ColumnType;
-import tech.tablesaw.api.NumericColumn;
-import tech.tablesaw.api.StringColumn;
+import tech.tablesaw.api.InstantColumn;
 import tech.tablesaw.columns.Column;
 import tech.tablesaw.table.Relation;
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/tech/tablesaw/conversion/smile/SmileConverter.java
```diff
@@ -22,125 +24,62 @@ public SmileConverter(Relation table) {
...
+  public DataFrame toDataFrame(StructType schema) {
+    List<Tuple> rows = new ArrayList<>();
+    int colCount = table.columnCount();
+    for (int rowIndex = 0; rowIndex < table.rowCount(); rowIndex++) {
+      Object[] row = new Object[colCount];
+      for (int colIndex = 0; colIndex < colCount; colIndex++) {
+        Column<?> col = table.column(colIndex);
+        if (!col.isMissing(rowIndex)) {
+          row[colIndex] =
+              col.type().equals(ColumnType.INSTANT)
```
benmccann: @haifengl this was the only thing I wasn't sure about in creating Tablesaw's Smile integration. There's no `InstantDataType` in Smile, so I converted the `Instant` to `LocalDateTime`. Does that seem like the right approach to you? Or would it be better for Smile to have an `InstantDataType`?

core/src/main/java/tech/tablesaw/conversion/smile/SmileConverter.java
```diff
@@ -22,125 +24,62 @@ public SmileConverter(Relation table) {
...
-  }
 
-  private NominalAttribute colAsNominalAttribute(Column<?> col) {
-    Column<?> unique = col.unique();
-    return new NominalAttribute(
-        col.name(),
-        unique
-            .mapInto(o -> o.toString(), StringColumn.create(col.name(), unique.size()))
-            .asObjectArray());
+    schema = schema.boxed(rows);
```
benmccann: @haifengl I was wondering if this made sense. I copied it from `CSV.java`, but should we not use the `schema` that's already available? I was curious why `CSV.read` does this instead of using the user-provided schema. Won't we lose column names, etc.?

core/src/main/java/tech/tablesaw/conversion/smile/SmileConverter.java
haifengl: You can wrap anything into a `DataFrame`'s Object column data type. For most use case, it should be fine to convert it to LocalDataTime too.

core/src/main/java/tech/tablesaw/conversion/smile/SmileConverter.java
haifengl: It is in case that missing values exist but we didn't detect it during inferSchema. For user provided schema, it should be nop (if the user provides the correct schema).

PR #738, opened 2020-01-12
Use memory mapping for reading files

aecio:
Thanks for contributing.

## Description

Implemented two new features:
- Reading files via memory mapping
- Single-pass row sampling while reading CSVs

I ended-up implementing these features for a project where I needed to read large files. Please let me know if you are interested in merging any of these features and if you'd like to see any other changes before merging.

## Testing

Did you add a unit test?

Yes. And all existing tests are still passing after the changes.

### DIFF
--- a/core/src/main/java/tech/tablesaw/io/ByteBufferInputStream.java
+++ b/core/src/main/java/tech/tablesaw/io/ByteBufferInputStream.java
@@ -1,0 +1,40 @@
+package tech.tablesaw.io;
+
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+
+public class ByteBufferInputStream extends InputStream {
+
+  private final ByteBuffer bb;
+
+  public ByteBufferInputStream(ByteBuffer buffer) {
+    this.bb = buffer;
+  }
+
+  @Override
+  public int read() {
+    if (bb.remaining() == 0) {
+      return -1;
+    }
+    return bb.get();
+  }
+
+  @Override
+  public int read(byte b[]) {
+    return read(b, 0, b.length);
+  }
+
+  @Override
+  public int read(byte b[], int off, int len) {
+    int length = Math.min(bb.remaining(), len);
+    if (length == 0) {
+      return -1;
+    }
+    bb.get(b, off, length);
+    return length;
+  }
+
+  public int available() {
+    return bb.remaining();
+  }
+}
--- a/core/src/main/java/tech/tablesaw/io/FileReader.java
+++ b/core/src/main/java/tech/tablesaw/io/FileReader.java
@@ -30,9 +30,9 @@ public abstract class FileReader {
    * the ColumnType inference logic. These types may not be correct.
    */
   public ColumnType[] getColumnTypes(
-      Reader reader, ReadOptions options, int linesToSkip, AbstractParser<?> parser) {
+      ReadOptions options, int linesToSkip, AbstractParser<?> parser) {
 
-    parser.beginParsing(reader);
+    parser.beginParsing(options.source().createReader());
 
     for (int i = 0; i < linesToSkip; i++) {
       parser.parseNext();
--- a/core/src/main/java/tech/tablesaw/io/Source.java
+++ b/core/src/main/java/tech/tablesaw/io/Source.java
@@ -1,121 +1,139 @@
 package tech.tablesaw.io;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.io.ByteStreams;
 import com.ibm.icu.text.CharsetDetector;
 import com.ibm.icu.text.CharsetMatch;
-import java.io.ByteArrayInputStream;
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.io.RandomAccessFile;
 import java.io.Reader;
-import java.io.StringReader;
 import java.net.URL;
+import java.nio.ByteBuffer;
+import java.nio.MappedByteBuffer;
+import java.nio.channels.FileChannel.MapMode;
 import java.nio.charset.Charset;
-import java.util.Scanner;
 
 public class Source {
 
-  // we always have one of these (file, reader, or inputStream)
-  protected final File file;
-  protected final Reader reader;
-  protected final InputStream inputStream;
-  protected final Charset charset;
+  protected final ByteBuffer byteBuffer;
+  protected Charset charset;
 
-  public Source(File file) {
-    this(file, getCharSet(file));
+  public Source(File file, Charset charset) throws IOException {
+    this(getMappedByteBuffer(file), charset);
   }
 
-  public Source(File file, Charset charset) {
-    this.file = file;
-    this.reader = null;
-    this.inputStream = null;
+  public Source(File file) throws IOException {
+    this(getMappedByteBuffer(file));
+  }
+
+  private static MappedByteBuffer getMappedByteBuffer(File file) throws IOException {
+    try (RandomAccessFile raf = new RandomAccessFile(file, "r")) {
+      return raf.getChannel().map(MapMode.READ_ONLY, 0, raf.length());
+    }
+  }
+
+  public Source(InputStream inputStream, Charset charset) throws IOException {
+    this(ByteStreams.toByteArray(inputStream), charset);
+  }
+
+  public Source(InputStream inputStream) throws IOException {
+    this(ByteStreams.toByteArray(inputStream));
+  }
+
+  public Source(byte[] bytes, Charset charset) {
+    this(ByteBuffer.wrap(bytes), charset);
+  }
+
+  public Source(byte[] bytes) {
+    this(ByteBuffer.wrap(bytes));
+  }
+
+  public Source(ByteBuffer byteBuffer) {
+    this.byteBuffer = byteBuffer;
+    this.charset = getCharSet(byteBuffer);
+  }
+
+  public Source(ByteBuffer byteBuffer, Charset charset) {
+    this.byteBuffer = byteBuffer;
     this.charset = charset;
   }
 
-  public Source(Reader reader) {
-    this.file = null;
-    this.reader = reader;
-    this.inputStream = null;
-    this.charset = null;
-  }
-
-  public Source(InputStream inputStream) {
-    this(inputStream, Charset.defaultCharset());
-  }
-
-  public Source(InputStream inputStream, Charset charset) {
-    this.file = null;
-    this.reader = null;
-    this.inputStream = inputStream;
-    this.charset = charset;
+  public static Source fromString(String s, Charset charset) {
+    return new Source(s.getBytes(charset), charset);
   }
 
   public static Source fromString(String s) {
-    return new Source(new StringReader(s));
+    return new Source(s.getBytes());
   }
 
   public static Source fromUrl(String url) throws IOException {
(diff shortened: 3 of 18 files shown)


core/src/main/java/tech/tablesaw/api/BooleanColumn.java
```diff
@@ -495,6 +495,18 @@ public BooleanColumn set(Selection rowSelection, boolean newValue) {
     return this;
   }
 
+  @Override
+  public Column<Boolean> setObj(int row, Object obj) {
```
benmccann: I don't see `setObj` called at all in this PR. Can we drop it from the PR? If it's necessary for you let's consider in a different PR to keep this one from getting too large

core/src/main/java/tech/tablesaw/io/ByteBufferInputStream.java
```diff
@@ -0,0 +1,40 @@
+package tech.tablesaw.io;
+
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+
+public class ByteBufferInputStream extends InputStream {
+
+  private ByteBuffer bb;
```
benmccann: this can be `final`

core/src/main/java/tech/tablesaw/io/FileReader.java
```diff
@@ -64,7 +67,9 @@ private String cleanName(String name) {
     return name.trim();
   }
 
-  /** Returns the column names for each column in the source. */
+  /**
```
benmccann: We have an autoformatting plugin setup. If you compile on the command line just once it will automatically run on git commit. The fact that your formatting is changing makes me think the formatting plugin hasn't run on this PR

core/src/main/java/tech/tablesaw/io/Source.java
```diff
@@ -1,121 +1,143 @@
...
+  }
+
+  private static MappedByteBuffer getMappedByteBuffer(File file) throws IOException {
+    try (RandomAccessFile raf = new RandomAccessFile(file, "r")) {
+      return raf.getChannel().map(MapMode.READ_ONLY, 0, raf.length());
+    }
+  }
+
+  public Source(InputStream inputStream, Charset charset) throws IOException {
+    this(ByteStreams.toByteArray(inputStream), charset);
```
benmccann: This will read all bytes from the stream into an array, which seems less efficient than streaming processing

benmccann:
There's a lot here. Is it possible to split the two new features into two PRs instead? That would make it easier to understand and review

core/src/main/java/tech/tablesaw/io/Source.java
aecio: Do you mean memory or time efficiency? It would be good to measure the impact of this change. Is there a benchmark available for tablesaw?

But note that this is not different from the previous implementation. If you look at the previous code from `getReaderAndColumnTypes()`, which was duplicated in both `CsvReader` and `FixedWidthReader`, you'll see that it caches the whole stream into a byte array of any input source different from a file:
```
bytesCache = CharStreams.toString(reader).getBytes();
```
Indeed, this line crashes the JVM for files larger than ~2GB because it tries to allocate an array with an index larger than an int can hold. This happened to me while reading a file via a `FileInputStream`.

Now, InputStreams are read once to a byte buffer and reused whenever necessary later (this still has the previous problem). Files are read directly from a MappedByteBuffer, so they are directly read from the OS cache when necessary. In my opinion, this simplifies the logic from `Source` class and their consumers which do not need to keep track of the type of source and close the underlying resources when necessary.

core/src/main/java/tech/tablesaw/io/FileReader.java
aecio: Was this set up recently? These commits are more than a month old (I was not able to make the PR earlier). I just rebased the code recently. Let's see if it works when I split this PR into separate PRs.

aecio:
Done. I extracted row sampling to PR #739 and this PR is now related to memory mapping. This also adds some minor constructors that allow the user to provide the charset manually when the correct charset is known (related to issue #638). Let me know you'd prefer these removed from this PR as well.

core/src/main/java/tech/tablesaw/io/Source.java
benmccann: I believe that before it would only cache for those two file types if the user did not supply the column types and it looks like it will now happen for all types under all circumstances. Since you said the read ahead may crash if the input is larger than 2GB I think I'd rather avoid it happening in a wider variety of circumstances.

lwhite1: > But note that this is not different from the previous implementation. If you look at the previous code from getReaderAndColumnTypes(), which was duplicated in both CsvReader and FixedWidthReader, you'll see that it caches the whole stream into a byte array of any input source different from a file:

hmm. I think @benmccann is correct. My understanding is that the reader _sometimes_ reads the whole file into memory, but that that isn't necessarily the case. Would this change mean always reading the file into memory?

> Indeed, this line causes the JVM to throw OOM for files larger than ~2GB because it tries to allocate an array with an index larger than an int can hold. 

I'm not sure I'm getting this. Where is there an index that's based on the file size? I know we would throw an error for more than Integer.max _rows_, but wasn't aware of a file size issue.

lwhite1:
> Implemented two new features:

> Reading files via memory mapping
> I ended-up implementing these features for a project where I needed to read large files. 

Could you help me understand what problems this change solves? You asked about memory versus time efficiency. Where do you see improvement over the current approach, and under what circumstances?

core/src/main/java/tech/tablesaw/io/Source.java
aecio: > hmm. I think @benmccann is correct. My understanding is that the reader _sometimes_ reads the whole file into memory, but that that isn't necessarily the case.

Yeah, my understanding is also that the reader only caches the whole file sometimes. More specifically, it caches when an `InputStream` or a `Source`  is provided as input instead of a `File`:
```
      if (options.source().file() == null) {	
        bytesCache = CharStreams.toString(reader).getBytes();	
        // create a new reader since we just exhausted the existing one	
        reader = options.source().createReader(bytesCache);	
      }
```

> Would this change mean always reading the file into memory?

No. This only reads the file into memory in the same situations as before. When an `InputStream` or a `Source` is provided, the data is cached in a byte array wrapped in a `ByteBuffer`. When a `File` is provided, the file is read directly using memory mapping via a `MappedByteBuffer`. This allows the rest of the code to always deal with the same interface.



> I'm not sure I'm getting this. Where is there an index that's based on the file size? I know we would throw an error for more than Integer.max _rows_, but wasn't aware of a file size issue.

In the cases where caching happens, the method `CharStreams.toString(reader)` reads the whole file into a single `String`, and strings have a maximum size due to array indexing issues mentioned before.

If I recall correctly, I got an OOM error when reading a large file using a `FileInputStream`, which forces `tablesaw` to read and cache the entire file. Debugging this issue motivated me to create this pull request. Only later I realized that providing a `File` instance was enough to solve my problem.

aecio:
> Could you help me understand what problems this change solves? You asked about memory versus time efficiency. Where do you see improvement over the current approach, and under what circumstances?

Originally, I wanted to solve the OOM mentioned in the other thread, however, only later while digging into the codebase to debug my problem, I realized the problem only happened because I was loading data using an `InputStream`. So, in fact, this PR solves no problem, it became only a refactoring.

However, because this change uses memory mapping it can potentially impact performance. I tend to believe that using memory mapping would increase IO performance because it tells the OS to load the file into memory if there is memory available. For CSV files, for example, the file is read multiple times (for type detection, CSV parsing), so the reader would likely hit the OS cache when there is enough memory available.

Previously the file was read using:
```
new InputStreamReader(new FileInputStream(file), charset)
```
In this PR, depending on the reader, it uses something like:
```
new ByteBufferInputStream(byteBuffer)
```
or
```
new InputStreamReader(new ByteBufferInputStream(byteBuffer))
```

Honestly, it is hard for me to tell which version performs better without a benchmark.

To fix the OOM error and avoid caching in memory, one could probably write the `InputStream` (or `Source`) to a temporary memory-mapped file and then use the `MappedByteBuffer` to read it multiple times.

core/src/main/java/tech/tablesaw/io/Source.java
benmccann: > it caches when an InputStream or a Source is provided as input instead of a File

I don't think that's quite right. It will cache it when an `InputStream` is provided, but I don't think it necessarily will when a `Source` is provided. A `Source` is just a container for some type of input. If the underlying input for the `Source` is a `File` then it won't read it into memory and cache it

aecio: You're right, what I meant was a `Reader`, not a `Source`. I updated the comment to reflect that.

aecio: Additionally, the caching happens immediately when the `Source` container is created, which is why the constructors had to change to throw an `IOException`.

lwhite1:
@aecio Thank you for the detailed replies. You make a great point about the benchmark. It would be really nice to have one that covered file reading and joining. 

OK, so if i understand correctly. This PR 
- uses memory mapping
- refactors out some duplicate code.
- would avoid an error that occurs if you read a file with more than MaxInt characters into a string
- has an unknown impact on time and memory performance.
Does that sum it up?

I had some old code that compared file reading performance in a separate project so I did a quick and dirty test reading files using the current approach and the memory mapping approach. I didn't see much difference except when reading files without sampling, although I only ran each option once, and my separate project is a Kotlin project which might have made some difference. When reading without sampling the MM version seemed to perform noticeably better.  
I put the results at the bottom if you're interested.

I think I would go ahead and merge this if @benmccann doesn't have any remaining concerns. What do you think, Ben?

**15 mb file** 

_with memory mapping_
Read with known column types: 2684
Read without known column types: 3015
Read without known column types: 3280  (No Sampling)

_without memory mapping_
Read with known column types: 2470
Read without known column types: 3273
Read without known column types: 3923 (No Sampling)

**500 mb file** 

_with memory mapping_
Read with known column types: 27248
Read without known column types: 26711

_without memory mapping_
Read with known column types: 28824
Read without known column types: 28568

benmccann:
I don't think my main concern has been addressed yet. Previously we would read the input into memory only when using the CSVReader and sampling to determine the column types. Now it will be done even when the user provides the column types and for all file types. This may consume a lot of memory and even lead to out of memory errors and crash the program, which the user could previously avoid by specifying the column types. I also don't see the benefit of doing this for the other file types. E.g. some file types like the newly proposed orc reader provide a schema that give you the file types. In this case why read everything into memory?

aecio:
@lwhite1 I think your points sums it up well, except for the issue related to loading large files into a string. As is, this PR does not fix that yet.  Also, thanks for the benchmark. I understand that it is not thorough, but it seems that memory mapping indeed improves in some cases.

@benmccann I understand your concern. Note that this problem can still be avoided by providing a `File` as a source instead of an `InputStream` when possible. Anyway, I think this issue can be fixed while still keeping memory mapping. I'm not sure when I'll be able to take a closer look at this, but I'll try to find some time.

lwhite1:
I think maybe we should close this. I don't think either @benmccann or I was completely sold on the benefits. We'd really need benchmarks to show a significant difference before introducing a big change like this.

aecio:
Closing is fine for me unless someone wants to take it on. It may improve speed, but it clearly needs more work and I won't be able to do more work on it in the next few months.

benmccann:
Ok, I'll go ahead and close for now

PR #739, opened 2020-01-13
Implement single-pass random sampling of rows while reading input

aecio:
Thanks for contributing.

## Description

Implemented a single-pass row random sampling while reading CSVs using the reservoir sampling algorithm. This PR was extracted from PR #738.

## Testing

Did you add a unit test?

Yes.

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/BooleanColumn.java
+++ b/core/src/main/java/tech/tablesaw/api/BooleanColumn.java
@@ -493,6 +493,18 @@ public class BooleanColumn extends AbstractColumn<BooleanColumn, Boolean>
     return this;
   }
 
+  @Override
+  public Column<Boolean> setObj(int row, Object obj) {
+    if (obj == null) {
+      return setMissing(row);
+    }
+    if (!(obj instanceof Boolean)) {
+      throw new IllegalArgumentException(
+          "Cannot append " + obj.getClass().getName() + " to BooleanColumn");
+    }
+    return set(row, (Boolean) obj);
+  }
+
   public double getDouble(int row) {
     return getByte(row);
   }
--- a/core/src/main/java/tech/tablesaw/api/BooleanColumn.java
+++ b/core/src/main/java/tech/tablesaw/api/BooleanColumn.java
@@ -16,7 +16,6 @@ package tech.tablesaw.api;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
-import com.google.common.base.Preconditions;
 import it.unimi.dsi.fastutil.booleans.BooleanIterable;
 import it.unimi.dsi.fastutil.booleans.BooleanIterator;
 import it.unimi.dsi.fastutil.booleans.BooleanOpenHashSet;
@@ -459,9 +458,8 @@ public class BooleanColumn extends AbstractColumn<BooleanColumn, Boolean>
   }
 
   @Override
-  public Column<Boolean> setValue(int row, String newValue, AbstractColumnParser<?> parser) {
-    Preconditions.checkArgument(!parser.isMissing(newValue));
-    return set(row, parser.parseByte(newValue));
+  public Column<Boolean> set(int row, String newValue, AbstractColumnParser<?> parser) {
+    return parser.isMissing(newValue) ? setMissing(row) : set(row, parser.parseByte(newValue));
   }
 
   @Override
--- a/core/src/main/java/tech/tablesaw/api/BooleanColumn.java
+++ b/core/src/main/java/tech/tablesaw/api/BooleanColumn.java
@@ -16,6 +16,7 @@ package tech.tablesaw.api;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
+import com.google.common.base.Preconditions;
 import it.unimi.dsi.fastutil.booleans.BooleanIterable;
 import it.unimi.dsi.fastutil.booleans.BooleanIterator;
 import it.unimi.dsi.fastutil.booleans.BooleanOpenHashSet;
@@ -447,13 +448,20 @@ public class BooleanColumn extends AbstractColumn<BooleanColumn, Boolean>
     return this;
   }
 
-  private void set(int i, byte b) {
+  private BooleanColumn set(int i, byte b) {
     data.set(i, b);
+    return this;
   }
 
   @Override
   public BooleanColumn set(int i, Boolean val) {
     return set(i, val.booleanValue());
+  }
+
+  @Override
+  public Column<Boolean> setValue(int row, String newValue, AbstractColumnParser<?> parser) {
+    Preconditions.checkArgument(!parser.isMissing(newValue));
+    return set(row, parser.parseByte(newValue));
   }
 
   @Override
--- a/core/src/main/java/tech/tablesaw/api/BooleanColumn.java
+++ b/core/src/main/java/tech/tablesaw/api/BooleanColumn.java
@@ -458,8 +458,8 @@ public class BooleanColumn extends AbstractColumn<BooleanColumn, Boolean>
   }
 
   @Override
-  public Column<Boolean> set(int row, String newValue, AbstractColumnParser<?> parser) {
-    return parser.isMissing(newValue) ? setMissing(row) : set(row, parser.parseByte(newValue));
+  public Column<Boolean> set(int row, String stringValue, AbstractColumnParser<?> parser) {
+    return set(row, parser.parseByte(stringValue));
   }
 
   @Override
--- a/core/src/main/java/tech/tablesaw/api/BooleanColumn.java
+++ b/core/src/main/java/tech/tablesaw/api/BooleanColumn.java
@@ -493,18 +493,6 @@ public class BooleanColumn extends AbstractColumn<BooleanColumn, Boolean>
     return this;
   }
 
-  @Override
-  public Column<Boolean> setObj(int row, Object obj) {
-    if (obj == null) {
-      return setMissing(row);
-    }
-    if (!(obj instanceof Boolean)) {
-      throw new IllegalArgumentException(
-          "Cannot append " + obj.getClass().getName() + " to BooleanColumn");
-    }
-    return set(row, (Boolean) obj);
-  }
-
   public double getDouble(int row) {
     return getByte(row);
   }
--- a/core/src/main/java/tech/tablesaw/columns/Column.java
+++ b/core/src/main/java/tech/tablesaw/columns/Column.java
@@ -565,11 +565,13 @@ public interface Column<T> extends Iterable<T>, Comparator<T> {
   Column<T> set(int row, T value);
 
   default Column<T> set(int row, String newValue, AbstractColumnParser<?> parser) {
-    setObj(row, parser.parse(newValue));
-    return this;
+    AbstractColumnParser<T> typedParser = (AbstractColumnParser<T>) parser;
+    if (parser.isMissing(newValue)) {
+      return setMissing(row);
+    } else {
+      return set(row, typedParser.parse(newValue));
+    }
   }
-
-  Column<T> setObj(int row, Object value);
 
   Column<T> set(int row, Column<T> sourceColumn, int sourceRow);
 
--- a/core/src/main/java/tech/tablesaw/columns/Column.java
+++ b/core/src/main/java/tech/tablesaw/columns/Column.java
@@ -564,9 +564,9 @@ public interface Column<T> extends Iterable<T>, Comparator<T> {
 
   Column<T> set(int row, T value);
 
-  default Column<T> set(int row, String newValue, AbstractColumnParser<?> parser) {
+  default Column<T> set(int row, String stringValue, AbstractColumnParser<?> parser) {
     AbstractColumnParser<T> typedParser = (AbstractColumnParser<T>) parser;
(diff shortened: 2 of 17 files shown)


core/src/main/java/tech/tablesaw/io/csv/CsvReadOptions.java
```diff
@@ -283,5 +286,11 @@ public Builder minimizeColumnSizes() {
       super.minimizeColumnSizes();
       return this;
     }
+
+    /** {@inheritDoc} */
```
benmccann: We use `@Override` instead of `{@inheritDoc}`

core/src/main/java/tech/tablesaw/io/ReadOptions.java
```diff
@@ -75,6 +75,7 @@
   protected final String missingValueIndicator;
   protected final boolean minimizeColumnSizes;
   protected final int maxCharsPerColumn;
+  protected final int sampleRowsIfGreaterThan;
```
benmccann: I don't think it's appropriate to put this in `ReadOptions` because it doesn't apply to all readers. It would be confusing if you tried to set this on an `HtmlReader` or `JsonReader` for example and then it did nothing. I would put it just on `CsvReadOptions` for now

core/src/main/java/tech/tablesaw/io/ReadOptions.java
```diff
@@ -294,6 +301,15 @@ public Builder minimizeColumnSizes() {
       return this;
     }
 
+    /**
+     * If the files contains more than numSamples, performs sampling using the single-pass reservoir
+     * sampling algorithm (https://en.wikipedia.org/wiki/Reservoir_sampling).
+     */
+    public Builder sampleRowsIfGreaterThan(int numSamples) {
```
benmccann: This name's a bit long. Can we come up with something shorter? How about `sampleSize`?

aecio: Sound good, but if the file has fewer rows than `numSamples`, the actual sample size will be smaller than `numSamples`. Does `maxSampleSize` sounds more descriptive for you?

core/src/main/java/tech/tablesaw/io/ReadOptions.java
aecio: Agreed. I somehow thought this would work for other readers. I'll fix this.

aecio: The problem here is that `FileReader` only has access to `ReadOptions`, not `CsvReadOptions`. So I need to either implement sampling for all readers or introduce a new intermediary class `FileReaderOptions` in the options class hierarchy that is extended by `CsvReadOptions` and `FixedWidthReadOptions`.

core/src/main/java/tech/tablesaw/io/csv/CsvReadOptions.java
aecio: Thanks, I'll fix it.

core/src/main/java/tech/tablesaw/io/ReadOptions.java
benmccann: Really? Can you explain more? I thought that if we passed in `numSamples` here that was greater than the number of rows in the file then the entire file would be read into the table

core/src/main/java/tech/tablesaw/io/ReadOptions.java
benmccann: Perhaps you could just add a parameter to the `FileReader` method instead? I think introducing a `FileReadOptions` might end up being a lot of code and also be confusing (e.g. you could read json or html from a file, but they wouldn't use `FileReadOptions`

core/src/main/java/tech/tablesaw/io/ReadOptions.java
aecio: Yes. If `numSamples` is smaller than the number of rows in the file than exactly `numSamples` are returned. If the file has `n` rows, and and `numSamples` > `n`, then `n` random samples are returned. Therefore, `numSamples` is the maximum number of samples returned since it is possible the number of rows in the file is less than `numSamples`.

benmccann: Ok, I think we had different connotations in mind for "samples". In my mind, there's no sampling if all records are returned, which is why `sampleSize` makes sense to me. `maxSampleSize` to me sounds confusing because it makes it sound like there might be sampling, but it's not clear how much.

aecio: You're right, effectively there is no sampling, but there could be oversampling to increase the samples to match the selected sample size. Nevertheless, I'm fine with `sampleSize`. I changed the code accordingly.

core/src/main/java/tech/tablesaw/io/ReadOptions.java
aecio: I followed your suggestion and moved sampling to CsvReadOptions using a method parameter.

core/src/main/java/tech/tablesaw/api/BooleanColumn.java
```diff
@@ -493,6 +493,18 @@ public BooleanColumn set(Selection rowSelection, boolean newValue) {
...
   }
 
+  @Override
+  public Column<Boolean> setObj(int row, Object obj) {
+    if (obj == null) {
+      return setMissing(row);
+    }
+    if (!(obj instanceof Boolean)) {
+      throw new IllegalArgumentException(
+          "Cannot append " + obj.getClass().getName() + " to BooleanColumn");
```
benmccann: "append" might not be quite the right word since we're not in the `append` method. Maybe "add" or something else?

core/src/main/java/tech/tablesaw/columns/Column.java
```diff
@@ -564,6 +564,13 @@ default boolean noneMatch(Predicate<? super T> test) {
 
   Column<T> set(int row, T value);
 
+  default Column<T> set(int row, String newValue, AbstractColumnParser<?> parser) {
+    setObj(row, parser.parse(newValue));
```
benmccann: The reason `appendCell` is implemented on each column type is to avoid auto-boxing. It's maybe a little ugly and I'm not sure how I feel about it, but we should probably do the same here to keep them consistent

core/src/main/java/tech/tablesaw/io/FileReader.java
```diff
@@ -22,8 +23,7 @@
 public abstract class FileReader {
 
   private static Logger logger = LoggerFactory.getLogger(FileReader.class);
-
-  public FileReader() {}
+  public static final int UNLIMITED_SAMPLE_SIZE = -1;
```
benmccann: this can be private

core/src/main/java/tech/tablesaw/io/FileReader.java
```diff
@@ -168,20 +180,44 @@ private void addRows(
...
 
-      // append each column that we're including (not skipping)
-      int cellIndex = 0;
-      for (int columnIndex : columnIndexes) {
-        Column<?> column = table.column(cellIndex);
-        AbstractColumnParser<?> parser = parserMap.get(column.name());
-        try {
-          String value = nextLine[columnIndex];
+      int samplesCount = table.rowCount();
+      if (sampleSize == -1 || samplesCount < sampleSize) {
```
benmccann: I might do `sampleSize < 0` rather than `sampleSize == -1` to be a bit more tolerant of input values

aecio: Fixed.

core/src/main/java/tech/tablesaw/io/FileReader.java
aecio: fixed.

core/src/main/java/tech/tablesaw/columns/Column.java
```diff
@@ -564,6 +564,15 @@ default boolean noneMatch(Predicate<? super T> test) {
 
   Column<T> set(int row, T value);
 
+  default Column<T> set(int row, String newValue, AbstractColumnParser<?> parser) {
+    AbstractColumnParser<T> typedParser = (AbstractColumnParser<T>) parser;
+    if (parser.isMissing(newValue)) {
+      return setMissing(row);
+    } else {
+      return set(row, typedParser.parse(newValue));
```
benmccann: this is still going to do autoboxing because it calls the generic `parse` instead of `parseDouble`, `parseFloat`, etc.

core/src/main/java/tech/tablesaw/columns/Column.java
aecio: Sorry, I'm seeing this comment just now. I ended up finding a way to completely remove all `setObj()` implementations at the cost of adding a typecast from `AbstractColumnParser<?>` to `AbstractColumnParser<T>`. Let me know how you feel about this and if it relates to this concern.

Also, can you please elaborate on where auto-boxing happens here?

benmccann: Not sure if you saw my comment on the new version, but the reason the auto-boxing happens is because `parse` returns an Object which must then be converted to a primitive. Every Column has its own implementation of `appendCell` to avoid this. I think we'd want to do the same here for consistency. @lwhite1 implemented this part, so may be able to chime in as well though he's pretty busy at the moment, so we probably shouldn't wait for him

core/src/main/java/tech/tablesaw/api/BooleanColumn.java
```diff
@@ -447,15 +448,22 @@ public BooleanColumn set(int i, boolean b) {
...
   }
 
   @Override
   public BooleanColumn set(int i, Boolean val) {
     return set(i, val.booleanValue());
   }
 
+  @Override
+  public Column<Boolean> setValue(int row, String newValue, AbstractColumnParser<?> parser) {
+    Preconditions.checkArgument(!parser.isMissing(newValue));
```
benmccann: I think I'd change the implementation of this method to:

```
  return parser.isMissing(newValue) ? setMissing(row) : set(row, parser.parseByte(newValue));
```

core/src/main/java/tech/tablesaw/columns/Column.java
aecio: Hi @benmccann, can you please take another look? 
I noticed that not all column types use primitive type parsing, so I modified the default implementation to delegate parsing to each specific column implementation which then uses primitive parsing. A default method is still provided to avoid code duplication for column types that don't use primitive types (dates and strings column types).

core/src/main/java/tech/tablesaw/api/BooleanColumn.java
```diff
@@ -447,15 +448,22 @@ public BooleanColumn set(int i, boolean b) {
...
+    return this;
   }
 
   @Override
   public BooleanColumn set(int i, Boolean val) {
     return set(i, val.booleanValue());
   }
 
+  @Override
+  public Column<Boolean> setValue(int row, String newValue, AbstractColumnParser<?> parser) {
```
benmccann: Can we just call this `set` instead of `setValue`?

core/src/main/java/tech/tablesaw/columns/Column.java
benmccann: I like the idea of providing a default method so you don't need to implement it on every column. I'd probably just override `set` rather than delegating to a new `setValue` method though to keep the API cleaner for users

core/src/main/java/tech/tablesaw/columns/Column.java
aecio: Implemented primitive type parsing for Column types that need it.

core/src/main/java/tech/tablesaw/columns/Column.java
aecio: On second thought, it seems that the missing value check on `setValue()` implementations is not needed, right? It seems that parsers do handle missing values correctly.

benmccann: Yes, that's true actually

core/src/main/java/tech/tablesaw/api/BooleanColumn.java
```diff
@@ -447,15 +447,21 @@ public BooleanColumn set(int i, boolean b) {
...
   }
 
   @Override
   public BooleanColumn set(int i, Boolean val) {
     return set(i, val.booleanValue());
   }
 
+  @Override
+  public Column<Boolean> set(int row, String newValue, AbstractColumnParser<?> parser) {
+    return parser.isMissing(newValue) ? setMissing(row) : set(row, parser.parseByte(newValue));
```
benmccann: I make the comment before you noticed that parser actually handles missing values. So you could simplify this to just `return set(row, parser.parseByte(newValue));`

aecio: Ok, I will simplify all methods to `return set(row, parser.parseByte(newValue));`.

core/src/main/java/tech/tablesaw/api/BooleanColumn.java
aecio: done.

core/src/main/java/tech/tablesaw/api/BooleanColumn.java
aecio: done.

aecio:
Thanks for the careful review!

PR #742, opened 2020-01-18
Change StringColumn summary to a more useful fixed format #718

jackie-h:
Thanks for contributing.

- [ X] Tick to sign-off your agreement to the [Developer Certificate of Origin (DCO) 1.1](https://developercertificate.org)

## Description

This is a pre-requisite for the larger change to change Relation.summary() to return a Table.
This changes the summary for the StringColumn to provide statistics in a fixed format. This is similar to the Python pandas implementation.

## Testing

Added a test for summary

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/StringColumn.java
+++ b/core/src/main/java/tech/tablesaw/api/StringColumn.java
@@ -182,7 +182,25 @@ public class StringColumn extends AbstractStringColumn<StringColumn> {
 
   @Override
   public Table summary() {
-    return countByCategory();
+    Table summary = Table.create(this.name());
+    StringColumn measure = StringColumn.create("Measure");
+    StringColumn value = StringColumn.create("Value");
+    summary.addColumns(measure);
+    summary.addColumns(value);
+
+    measure.append("Count");
+    value.append(String.valueOf(size()));
+
+    measure.append("Unique");
+    value.append(String.valueOf(countUnique()));
+
+    Table countByCategory = countByCategory().sortDescendingOn("Count");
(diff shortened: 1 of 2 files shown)


core/src/main/java/tech/tablesaw/api/StringColumn.java
```diff
@@ -182,7 +182,25 @@ public String get(int rowIndex) {
...
+    value.append(String.valueOf(size()));
+
+    measure.append("Unique");
+    value.append(String.valueOf(countUnique()));
+
+    Table countByCategory = countByCategory().sortDescendingOn("Count");
+    measure.append("Top");
+    value.append(countByCategory.stringColumn("Category").getString(0));
+
+    measure.append("Top Freq.");
```
benmccann: I'd probably call this just `"Freq"` for consistency with pandas and also because `"Top Freq."` sounds like it's the frequency of the `"Top"` item

jackie-h: I'm not following this comment. This value is the frequency (number of occurrences) of the "Top" item, both here and in pandas.  
From the pandas docs - "The top is the most common value. The freq is the most common value's frequency."
I added the "Top"  because I find the naming in pandas confusing. It seemed better to be explicit and it makes it clearer that those values are a pair/go together when you view the result.

benmccann: Ok, looks good. Thanks for clarifying!

benmccann:
Thanks for this! I think the new summary implementation will be very cool!

Issue #747, opened 2020-01-30
Column type detection strategy for numbers ending with .0


larshelge:
First of all, thanks for an excellent library!

This issue relates to the table column type detection. My use-case is having TableSaw detect the column types of a CSV data file. That information is used to create a table in AWS Redshift and later ingest the CSV file into the table. I am mapping TableSaw `INTEGER` column type to Redshift `INTEGER` data type.

I came across a corner case in which TableSaw detects numbers ending with ".0", such as "1613.0" or "962.0", as integer column type. This is arguably correct, but fails in my use-case, as AWS Redshift does not accept such values as integers and fails during data load.

Would this case be worthy of a configurable read option, to control whether numbers with zero value decimals will be mapped to integer or floating point types? 

`ReadOptions.zeroDecimalNumberAsFloat` or a more succinct name would be great.

Tablesaw version 0.37.3.

benmccann:
This looks like a duplicate of https://github.com/jtablesaw/tablesaw/issues/732

So Redshift and TableSaw are both reading from the CSV and you need the column types to match? It might help to explain the use case the more. Why does every single cell end in ".0"?

larshelge:
Thanks for the quick reply. It is _not quite_ the same problem, although related. `732` is about the sample phase returning the wrong result, i.e. integer when the data column contains double values with arbitrary decimal values.

This issue is about the sampling arguably returning the correct result (integer), but in the case of matching to other systems such as Redshift, should preferably return a floating type (i.e. double). I would argue that `100.0` should be considered a double, but that might be an opinion.

That said, since there seems to be a use-case for treating `100.0` as a double, perhaps it could deserve an option; either a specific setting such as `zeroDecimalsAsFloatingPoint` or a "mode" setting such as `ReadOptions.parsingMode(STRICT | LENIENT)`. 

My preference would be to change the behavior but I realize that is not backwards compatible and debatable. Second preference would be an option.

To answer your question - I don't know. We are building a data platform where users may upload their own CSV files and so we need to handle less-than-perfect data files. Thanks again!

larshelge:
Created PR for `ReadOptions.ignoreZeroDecimals` here #748.

benmccann:
Closed by https://github.com/jtablesaw/tablesaw/pull/748

Issue #732, opened 2019-12-09
CSV import: IntParser removes decimal when value ends with ".0" during parsing


szabolcskarolyi:
We have a case when our input csv has a column that has many values of "0.0" and some real double values.
Problem is that during the sample phase the columnType is set to Integer because the sample values pass in the parse test due to the check at:
https://github.com/jtablesaw/tablesaw/blob/master/core/src/main/java/tech/tablesaw/columns/numbers/IntParser.java#L28

Wondering if that check could be removed or be optional.

We came up with a solution, where we created a custom IntColumnType and IntParser, trying to set columnTypesToDetect in ReadOptions, however it looks like the setter is missing in CsvReadOptions.Builder so the default columnTypesToDetect list cannot be overriden.
We used reflection, but would like to avoid this solution.
Ideal would be to handle "x.0" values as double instead.

We would appreciate your help :)
Szabolcs

benmccann:
If I understand correctly, it seems to me like the issue isn't that 0.0 is being treated as an integer, but that the real double values are not being encountered during sampling. You could either disable sampling or you could specify that the column is a double column.

szabolcskarolyi:
You are right, but we cannot do either. We have like 200+ columns that makes difficult to specify column types and disable sampling leads to memory issues as our csvs are quite large.

benmccann:
What if we made it so that you can specify just some column types by name or index?

lwhite1:
@szabolcskarolyi 
> You are right, but we cannot do either.

I'm not sure why having 200 columns prevents your specifying column types. Unless the columns change constantly, it's not a lot of work.

lwhite1:
> What if we made it so that you can specify just some column types by name or index?

This seems like a good idea, but I don't think I would do it by index, since the main complaint I've seen about providing column type defs is for files that change their shape over time.

benmccann:
Closed by https://github.com/jtablesaw/tablesaw/pull/748

PR #748, opened 2020-01-30
Add ignoreZeroDecimal to ReadOptions

larshelge:

## Description

Introduces new option `ignoreZeroDecimal` to `ReadOptions`. This option controls whether a numeric value ending with ".0" may be considered an integer (or short, long). Default value is `true` and retains the current behavior. If set to false, values ending with ".0" will be not be considered an integer and instead be considered a floating point.

Since different software/systems have different opinions on this matter, an option to control this behavior could be justified. Will be useful when using TableSaw together with other systems.

The logic could be centralized in `AbstractColumnParser` if that is preferred. Open for other variable naming suggestions.

Relates to issue #747.

## Testing

Tests added to `CsvReaderTest` and `StringUtilsTest`.


larshelge:
Fully appreciating that people are busy, I was just wondering whether you may consider this patch, or if I should start working on an external work-around?

core/src/main/java/tech/tablesaw/columns/numbers/IntParser.java
```diff
@@ -7,6 +7,8 @@
 
 public class IntParser extends AbstractColumnParser<Integer> {
 
+  private boolean zeroDecimalAsFloat;
```
benmccann: this could be made `final` (same in the other parsers)

`zeroDecimalAsFloat` makes me think of the type `float`, which might be misleading. How about something like `ignoreZeroDecimal`?

core/src/main/java/tech/tablesaw/columns/numbers/IntParser.java
```diff
@@ -25,7 +28,7 @@ public boolean canParse(String str) {
     }
     String s = str;
     try {
-      if (s.endsWith(".0")) {
+      if (!zeroDecimalAsFloat && s.endsWith(".0")) {
```
benmccann: It's a little weird that it only works with a single 0. What if it was `.00`? I know you didn't write that part, but it might be a good opportunity to clean it up

core/src/main/java/tech/tablesaw/columns/numbers/IntParser.java
larshelge: Agreed. Went back and forth there myself. Updated it now to `ignoreZeroDecimal`.

larshelge: Updated boolean variables to be `final` now.

In order to centralize the default value (`true`) and avoid maintaining it in several places, I added a constant to `ReadOptions`. Let me know if that is acceptable.

core/src/main/java/tech/tablesaw/columns/numbers/IntParser.java
larshelge: Agreed. I have updated the PR now.

I added a function to `StringUtils` so that we can centralize the logic. I opted for a regex with a compiled pattern for performance reasons. I combined the check for zero decimal with the remove operation as it is simpler and performance will be the same.

This made the `FixedWidthReaderTest#testDataTypeDetection` test fail which is legitimate as the test data contains data with two zero decimals, so I updated the test.

larshelge: Let me know if the util method is better off in another util class or if I should create a new one.

larshelge:
Would it be possible to kindly get a timeline for when a review of this PR could take place? I am happy to change the approach to what the reviewer find most appropriate.

lwhite1:
I guess I don't see why you can't read the file as a double or float column and then convert it to an int column. The asIntColumn() method uses a cast 

```
  public IntColumn asIntColumn() {
    IntColumn result = IntColumn.create(name());
    for (double d : data) {
      if (DoubleColumnType.valueIsMissing(d)) {
        result.appendMissing();
      } else {
        result.append((int) d);
      }
    }
    return result;
  }

```
DoubleColumn also implements asShortColumn and asLongColumn.

One advantage to doing this work here as opposed to in the reader is that most numeric operators return a double column, which you may want to convert to an integral value.

benmccann:
Yeah, I'd debated that as well. You could generally convert after the fact, but it's probably enough extra work that I didn't think this approach was too bad.

Also, there might be circumstances where the current approach makes you lose information. Right now we always strip ".0" and treat as an int. But what if you had a data source where some columns had trailing zeros and others didn't and you want to retain that information? Then I think you'd need to turn off the automatic stripping of ".0". I think that's sort of what was being suggested in https://github.com/jtablesaw/tablesaw/issues/747

larshelge:
Many thanks for the comments, really appreciate it.

Just wanted to mention that the original use-case for #747 is _data type detection_, where for various reasons one would like to consider numbers ending in zero-decimals only as type double.

Converting the column to int after the fact is not really an option as we do not know the data type of the data files/columns up front, as they are uploaded by users. We are using TableSaw to
inform us what data type each column has.

Note that this PR will also coincidentally solve the problem of #732.

core/src/main/java/tech/tablesaw/columns/numbers/IntParser.java
lwhite1: This looks good to me.

core/src/main/java/tech/tablesaw/columns/numbers/IntParser.java
lwhite1: This seems fine. Would you mind adding the test?

core/src/test/java/tech/tablesaw/io/csv/CsvReaderTest.java
```diff
@@ -173,6 +173,64 @@ public void testDataTypeDetection() throws IOException {
     assertArrayEquals(bus_types, columnTypes);
   }
 
+  @Test
+  public void testNumberDetectionRetainZeroDecimals() throws IOException {
+    Reader reader = new FileReader("../data/immunization.csv");
+    CsvReadOptions options =
+        CsvReadOptions.builder(reader).header(true).sample(false).ignoreZeroDecimal(true).build();
```
larshelge: The auto-formatter renders this on one line, originally one line per operation.

core/src/test/java/tech/tablesaw/io/csv/CsvReaderTest.java
```diff
@@ -173,6 +173,64 @@ public void testDataTypeDetection() throws IOException {
...
+    Table table = Table.read().csv(options);
+    assertEquals(INTEGER, table.column(3).type());
+    assertEquals(INTEGER, table.column(7).type());
+  }
+
+  @Test
+  public void testNumberDetectionIgnoreZeroDecimals() throws IOException {
+    Reader reader = new FileReader("../data/immunization.csv");
+    CsvReadOptions options =
+        CsvReadOptions.builder(reader).header(true).sample(false).ignoreZeroDecimal(false).build();
```
larshelge: The auto-formatter renders this on one line, originally I had one line per operation.

core/src/main/java/tech/tablesaw/columns/numbers/IntParser.java
larshelge: I have added tests now. These include checks for the column type detection feature and for getting columns after reading the data into a table. There are tests asserting behavior when `ignoreZeroDecimal` is disabled as well as enabled.

Let me know if the tests look appropriate.

larshelge:
I have added tests now, see `CsvReaderTest.java`.

core/src/test/java/tech/tablesaw/io/csv/CsvReaderTest.java
benmccann: I would just leave it like this. The autoformatter allows us to not have to worry about formatting .Even if it's slightly suboptimal in some cases I think it's a win to be able to just ignore formatting

larshelge: Got it @benmccann, no problem, thanks.

larshelge:
I have added tests. @lwhite1 would it be possible to kindly ask for a review?

core/src/test/java/tech/tablesaw/io/csv/CsvReaderTest.java
```diff
@@ -716,6 +774,7 @@ public void testReadCsvWithRowSampling() throws IOException {
   }
 
   @Test
+  @SuppressWarnings("unchecked")
```
benmccann: Can you remove this line? I just fixed it so that it's no longer necessary

core/src/test/java/tech/tablesaw/util/StringUtilsTest.java
```diff
@@ -0,0 +1,43 @@
...
+  public void testRemoveZeroDecimal() {
+    // Assert zero decimals being removed
+    assertEquals("246", StringUtils.removeZeroDecimal("246.0"));
+    assertEquals("146", StringUtils.removeZeroDecimal("146.00"));
+    assertEquals("357", StringUtils.removeZeroDecimal("357.000"));
+    assertEquals("347", StringUtils.removeZeroDecimal("347.0000"));
+
+    // Assert no change to input value
+    assertEquals("468", StringUtils.removeZeroDecimal("468"));
+    assertEquals("24.", StringUtils.removeZeroDecimal("24."));
```
benmccann: This assertion is a bit funny to me because "24." doesn't seem like a valid value. I'd either remove the check or check that it's "24"

core/src/test/java/tech/tablesaw/io/csv/CsvReaderTest.java
larshelge: Fixed now.

core/src/test/java/tech/tablesaw/util/StringUtilsTest.java
larshelge: Agreed. Fixed now. Changed to "24".

larshelge:
I have added tests. @lwhite1 would it be possible to kindly ask for a review?

core/src/test/java/tech/tablesaw/io/csv/CsvReaderTest.java
```diff
@@ -173,6 +173,64 @@ public void testDataTypeDetection() throws IOException {
     assertArrayEquals(bus_types, columnTypes);
   }
 
+  @Test
+  public void testNumberTypeDetectionIgnoreZeroDecimal() throws IOException {
```
benmccann, line 177: This test and the next are both testing the same thing. (Same with the two tests with `ignoreZeroDecimal(false)`). It's a bit duplicative to have both, so I think it'd be nicer to have just one of them

larshelge: Thanks. I have removed two tests now (the ones based on `Table.read()`). Please let me know if anything else should be adjusted.

PR #753, opened 2020-02-07
Approach to use Jackson for jsplot

emilianbold:
Thanks for contributing.

## Description

Some of the plotly JSON structures are generated using Jackson. If this approach looks OK, I could use it in more places.

## Testing

No, just fixed the existing tests after the changes.


jsplot/pom.xml
```diff
@@ -90,6 +90,10 @@
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-core</artifactId>
```
benmccann: you could probably remove this since it will get pulled in by `jackson-databind`

jsplot/src/main/java/tech/tablesaw/plotly/components/Component.java
```diff
@@ -12,11 +15,33 @@
 public abstract class Component {
 
   protected final PebbleEngine engine = TemplateUtils.getNewEngine();
+  protected final ObjectMapper mapper = new ObjectMapper();
```
benmccann: you could arguably make these variables `static`

jsplot/src/main/java/tech/tablesaw/plotly/components/Component.java
```diff
@@ -12,11 +15,33 @@
...
+  {
+    mapper.enable(SerializationFeature.INDENT_OUTPUT);
+    mapper.setSerializationInclusion(Include.NON_NULL);
+    mapper.enable(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS);
+  }
+
+  @Deprecated
   public abstract String asJavascript();
 
+  @Deprecated
```
benmccann: You could probably just remove these methods rather than deprecating. I doubt many people are extending these classes vs using them directly

jsplot/src/main/java/tech/tablesaw/plotly/components/Component.java
```diff
@@ -12,11 +15,33 @@
...
+  protected Map<String, Object> getJSONContext() {
+    return null;
+  }
+
+  public String asJSON() {
+    StringWriter w = new StringWriter();
+    try {
+      mapper.writeValue(w, getJSONContext());
+    } catch (IOException ioe) {
+    }
```
benmccann: You might want to `throw new UncheckedIOException(ioe);` here

jsplot/pom.xml
```diff
@@ -90,6 +90,10 @@
...
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
         <dependency>
             <groupId>io.pebbletemplates</groupId>
```
benmccann: can this be removed now?

jsplot/src/main/java/tech/tablesaw/plotly/components/Component.java
emilianbold: Right now we have components generating JS with PebbleEngine and Jackson. So it makes sense to deprecate this until PebbleEngine can be removed entirely.

jsplot/pom.xml
emilianbold: Not yet... still many components using it.

emilianbold:
Do you usually 'squash and merge'? Because it touches more files and keeping the individual commits looks better to me.

benmccann:
Yeah, we use squash and merge. It's a lot easier to review the commit history to draft release notes, etc. to have a single commit than to have multiple for the same feature. I'm not sure why you say it'd touch more files. It should only touch files that are touched in one of the commits.

emilianbold:
Fine by me.

It only touched the files modified by the commits (of course...) but to me it makes sense to see changes related to `HoverLabel` separate from changes related to the `Font` class.

In this case the squashed commit is not so complex, but I assume there are situations where squash is not used, correct?

benmccann:
I suppose if a PR had only two commits and they were addressing different concerns we might not use squash, but if commits are dealing with the same topic we'll generally squash them

PR #758, opened 2020-02-14
Sheet name

### LINKED ISSUE #756: Storing/Using sheet name when reading excel
At present (0.37.3), when we read excel the name associated with table object is the file name. Can we change this behavior and instead save the sheet name (atleast when reading multi sheet excel)?

I had a use case to search by sheet name and had to make changes in the current XlsxReader for this. 

### Current
```
 book1.xlsx  
 A  |   B   |
-------------
 1  |   56  |
 2  |   78  |
 3  |  100  |
 4  |  122  |
 5  |  144  |
 6  |  166  |

 book1.xlsx 
 C  |  D   |
------------
 4  |  12  |
 5  |  32  |
 6  |  52  |
 7  |  72  |
```

### Proposed
```
 sheet1
 A  |   B   |
-------------
 1  |   56  |
 2  |   78  |
 3  |  100  |
 4  |  122  |
 5  |  144  |
 6  |  166  |

 sheet2 
 C  |  D   |
------------
 4  |  12  |
 5  |  32  |
 6  |  52  |
 7  |  72  |
```

R1j1t:
Thanks for contributing.

## Description

closes #756 

This commit stores the sheet name as part of `Table.name()`. Following are the TODO
- [ ]  ~~add sheet selection by name~~

Edit:
Removed sheet selection by name for this PR

## Testing

Did you add a unit test?
I added a test to check name in `Table`


excel/src/test/java/tech/tablesaw/io/xlsx/XlsxReaderTest.java
```diff
@@ -135,6 +135,7 @@ public void testSheetIndex() throws IOException {
             .read(XlsxReadOptions.builder("../data/multiplesheets.xlsx").sheetIndex(1).build());
     assertNotNull(table, "No table read from multiplesheets.xlsx");
     assertColumnValues(table.stringColumn("stringcol"), "John", "Doe");
+    assertEquals("multiplesheets.xlsx#Sheet2",table.name(),"table name is different");
```
benmccann: This doesn't match the formatting we use elsewhere. I.e. there should be a space after the commas. If you compile on the command line once with `mvn compile` it will install a git pre-commit hook that should automatically do the formatting for you.

I might remove the last parameter `"table name is different"` since I don't think it adds much to the default message

excel/src/main/java/tech/tablesaw/io/xlsx/XlsxReader.java
```diff
@@ -58,6 +58,7 @@ public static void register(ReaderRegistry registry) {
     registry.registerOptions(XlsxReadOptions.class, INSTANCE);
   }
 
+  // TODO Read by sheet name
```
benmccann: Let's leave out the TODO. There's tons of features we could add in the future and I don't know if putting a TODO for each one is the right way to track them

R1j1t: Okay sure! I got busy with other work and so could not start this. Let me remove the TODO. By the way, this library is very useful. I used to work with python for data science but recently started using Java. And libraries like these help me focus on the core part rather than data loading. Please let me know on how else can I contribute to your roadmap. Will try to contribute as much as possible from my end.

excel/src/test/java/tech/tablesaw/io/xlsx/XlsxReaderTest.java
R1j1t: sounds good, sorry about that.

excel/src/main/java/tech/tablesaw/io/xlsx/XlsxReader.java
benmccann: We don't really have much of a roadmap. I've just been adding things as I encounter a need for them. But I've slowed down a lot recently since most of the stuff I need the most is present now

R1j1t: Somehow it is good. Maybe we can get some feature inspiration from pandas or matplotlib, just a small suggestion

excel/src/test/java/tech/tablesaw/io/xlsx/XlsxReaderTest.java
R1j1t: I did `mvn compile` but not sure if pre-commit was configured, so I manually made the change. Also removing TODO from the PR description

Issue #775, opened 2020-03-07
Calling .unique then .sortAscending results in all 0s.


g-r-t:
I encountered an issue that seems to be related to #586. However, #586 was marked as closed in August 2019, and I'm experiencing this issue using the latest 0.37.3 jar from Maven Central. 

The issue is when .sortAscending is called on a DoubleColumn produced by a call to .unique, the result is all 0s. 

I created a short, simple example that illustrates the issue:
```
        double[] ds = { 5, 4, 3, 2, 1, 5, 4, 3, 2, 1 };
        DoubleColumn unq = DoubleColumn.create("All 0s", ds).unique();
        System.out.println(Arrays.toString(unq.asDoubleArray()));
        unq.sortAscending();
        System.out.println(Arrays.toString(unq.asDoubleArray()));

        System.out.println();
        
        // Must make a copy here, otherwise sorting will expand with 0s.
        unq = DoubleColumn.create("All 0s", ds).unique().copy();
        System.out.println(Arrays.toString(unq.asDoubleArray()));
        unq.sortAscending();
        System.out.println(Arrays.toString(unq.asDoubleArray()));
```
The output from running this is:
```
[1.0, 4.0, 3.0, 5.0, 2.0]
[0.0, 0.0, 0.0, 0.0, 0.0]

[1.0, 4.0, 3.0, 5.0, 2.0]
[1.0, 2.0, 3.0, 4.0, 5.0]
```

As featured in the example code, it's possible to work around the issue by calling .copy on the DoubleColumn produced by the call to .unique.

benmccann:
Thanks. I've added a test for this here: https://github.com/jtablesaw/tablesaw/commit/6febe7d061ca28dd5913eb2bcd5a99668fdec554

I'm guessing that the issue is with `sortAscending`. It seems to work if I change the code there:

```
  @Override
  public DoubleColumn sortAscending() {
    double[] arr = asDoubleArray(); 
    DoubleArrays.parallelQuickSort(arr);
    return DoubleColumn.create(name(), arr);
  }
```

Though the way I changed the code changes the signature and requires the other methods to be updated as well. I'm not sure I'll have time to fix, but I'd be happy to review a PR if you want to take a stab at it.

emilianbold:
I just looked into this.

The problem is the way `data.elements()` is used. `DoubleArrayList.elements()` exposes the *internal* data.

[...]

PR #778, opened 2020-03-10
Sorts column data using fastutil .sort methods

emilianbold:
Thanks for contributing.

## Description
Sorts the fastutil collections using the proper `sort` method and avoids accessing the backing array via `.elements()` since it will generally have invalid data or zeros in there.

Closes issue #775

## Testing

Two tests, and everything existing still runs which is promising.

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/BooleanColumn.java
+++ b/core/src/main/java/tech/tablesaw/api/BooleanColumn.java
@@ -23,7 +23,6 @@ import it.unimi.dsi.fastutil.booleans.BooleanSet;
 import it.unimi.dsi.fastutil.bytes.Byte2IntMap;
 import it.unimi.dsi.fastutil.bytes.Byte2IntOpenHashMap;
 import it.unimi.dsi.fastutil.bytes.ByteArrayList;
-import it.unimi.dsi.fastutil.bytes.ByteArrays;
 import it.unimi.dsi.fastutil.bytes.ByteComparator;
 import it.unimi.dsi.fastutil.bytes.ByteIterator;
 import it.unimi.dsi.fastutil.bytes.ByteListIterator;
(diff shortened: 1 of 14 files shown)


core/src/test/java/tech/tablesaw/api/DoubleColumnTest.java
```diff
@@ -30,6 +29,14 @@ public void unique() {
     assertArrayEquals(arr, new double[] {1.0, 2.0, 3.0, 4.0, 5.0});
   }
 
+  @Test
+  public void createThenSortAscending() {
```
benmccann: How is this different than the existing `sortAscending` test?

core/src/main/java/tech/tablesaw/api/BooleanColumn.java
```diff
@@ -297,12 +296,12 @@ public BooleanColumn copy() {
 
   @Override
   public void sortAscending() {
-    ByteArrays.mergeSort(data.elements());
+    data.sort(descendingByteComparator.reversed() /* or ByteComparators.NATURAL_COMPARATOR */);
```
benmccann: This is kind of a double negative. I'd prefer the `ByteComparators.NATURAL_COMPARATOR` option rather than reversing a reverse comparator

core/src/main/java/tech/tablesaw/api/BooleanColumn.java
```diff
@@ -297,12 +296,12 @@ public BooleanColumn copy() {
...
   }
 
   @Override
   public void sortDescending() {
-    ByteArrays.mergeSort(data.elements(), descendingByteComparator);
+    data.sort(descendingByteComparator);
```
benmccann: Perhaps we should remove `descendingByteComparator` and use `ByteComparators.OPPOSITE_COMPARATOR` here for consistency if we change `sortAscending`

core/src/test/java/tech/tablesaw/api/DoubleColumnTest.java
emilianbold: When you create an empty column like DoubleColumn.create("test") it actually creates a `DEFAULT_ARRAY_SIZE` array which has zeros. It showcases the bug nicely.

core/src/main/java/tech/tablesaw/api/BooleanColumn.java
emilianbold: Yeah, I'd use the natural comparators all over the place. But incrementally it made sense in a way to use whatever comparator you had there (just in case there was some specific logic -- I didn't really check them all).

I will update the PR to use those comparators.

emilianbold:
See the updated PR.

benmccann:
Thanks for the fix!

PR #781, opened 2020-03-16
expose more config for bubble plot

rayeaster:
expose more configurable options for bubble plot.

### DIFF
--- a/jsplot/src/main/java/tech/tablesaw/plotly/api/BubblePlot.java
+++ b/jsplot/src/main/java/tech/tablesaw/plotly/api/BubblePlot.java
@@ -2,9 +2,12 @@ package tech.tablesaw.plotly.api;
 
 import java.util.List;
 import tech.tablesaw.api.Table;
+import tech.tablesaw.columns.Column;
 import tech.tablesaw.plotly.components.Figure;
 import tech.tablesaw.plotly.components.Layout;
 import tech.tablesaw.plotly.components.Marker;
+import tech.tablesaw.plotly.components.Marker.MarkerBuilder;
+import tech.tablesaw.plotly.components.Marker.SizeMode;
 import tech.tablesaw.plotly.traces.ScatterTrace;
 import tech.tablesaw.table.TableSliceGroup;
 
@@ -36,22 +39,42 @@ public class BubblePlot {
     }
     return new Figure(layout, traces);
   }
+  
+  public static Figure create(
+	      String title, Table table, 
+	      String xCol, Column xColumn,
+	      String yCol, Column yColumn,
+	      String sizeColumn, 
+	      double[] color, SizeMode sizeMode, Double opacity) {
+		Layout layout = Layout.builder(title, xCol, yCol).build();
+
+		Marker marker = null;
+		MarkerBuilder builder = Marker.builder();
+		if (sizeColumn != null) {
+			builder.size(table.numberColumn(sizeColumn));
+		}
+		if (opacity != null) {
+//		   builder.opacity(.75);
+			builder.opacity(opacity);
+		}
+		if (color != null) {
+			builder.color(color);
+		}
+		if (sizeMode != null) {
+			builder.sizeMode(sizeMode);
+		}
+		marker = builder.build();
+
+		xColumn = (xColumn == null) ? table.numberColumn(xCol) : xColumn;
+		yColumn = (yColumn == null) ? table.numberColumn(yCol) : yColumn;
+
+		ScatterTrace trace = ScatterTrace.builder(xColumn, yColumn).marker(marker).build();
+		return new Figure(layout, trace);
+  }
 
   public static Figure create(
       String title, Table table, String xCol, String yCol, String sizeColumn) {
-    Layout layout = Layout.builder(title, xCol, yCol).build();
-
-    Marker marker =
-        Marker.builder()
-            .size(table.numberColumn(sizeColumn))
-            // .opacity(.75)
-            .build();
-
-    ScatterTrace trace =
-        ScatterTrace.builder(table.numberColumn(xCol), table.numberColumn(yCol))
-            .marker(marker)
-            .build();
-    return new Figure(layout, trace);
+        return create(title, table, xCol, null, yCol, null, sizeColumn, null, null, null);
   }
 
   public static Figure create(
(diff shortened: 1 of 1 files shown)


jsplot/src/main/java/tech/tablesaw/plotly/api/BubblePlot.java
```diff
@@ -36,22 +39,42 @@ public static Figure create(
...
+	      double[] color, SizeMode sizeMode, Double opacity) {
+		Layout layout = Layout.builder(title, xCol, yCol).build();
+
+		Marker marker = null;
+		MarkerBuilder builder = Marker.builder();
+		if (sizeColumn != null) {
+			builder.size(table.numberColumn(sizeColumn));
+		}
+		if (opacity != null) {
+//		   builder.opacity(.75);
```
benmccann: I would suggest removing this line

rayeaster:
> It looks like you use tabs where the rest of the code uses spaces and some other formatting that's not consistent. If you run `mvn compile` one time it will install a git pre-commit hook so that whenever you commit code the code will be automatically formatted and you don't have to worry about formatting

will try that, thanks

jsplot/src/main/java/tech/tablesaw/plotly/api/BubblePlot.java
rayeaster: sure, will do, thanks

benmccann:
Hey @rayeaster I'm not sure the formatting has been updated yet. It still looks like it uses tabs instead of spaces

rayeaster:
> Hey @rayeaster I'm not sure the formatting has been updated yet. It still looks like it uses tabs instead of spaces

now it should looks good with the google formatter. Thanks for taking on review

benmccann:
Thanks for that!

I'm concerned that the plugin didn't work automatically for you on Windows. It should work there. I saw that they include a few fixes for Windows in some of the recent releases, so I've upgraded it to the latest release: https://github.com/jtablesaw/tablesaw/commit/9d5fce1401d61743fb4cbc368fe92d8dc1f933da. Would you be able to test it again and let me know if it works for you now or what error message you get if it doesn't? I don't want other users to run into this problem. Thanks!

rayeaster:
> Thanks for that!
> 
> I'm concerned that the plugin didn't work automatically for you on Windows. It should work there. I saw that they include a few fixes for Windows in some of the recent releases, so I've upgraded it to the latest release: [9d5fce1](https://github.com/jtablesaw/tablesaw/commit/9d5fce1401d61743fb4cbc368fe92d8dc1f933da). Would you be able to test it again and let me know if it works for you now or what error message you get if it doesn't? I don't want other users to run into this problem. Thanks!

yes, `mvn compile` works now for me in Windows. Thanks

jsplot/src/main/java/tech/tablesaw/plotly/api/BubblePlot.java
```diff
@@ -38,22 +41,46 @@ public static Figure create(
   }
 
   public static Figure create(
-      String title, Table table, String xCol, String yCol, String sizeColumn) {
+      String title,
+      Table table,
+      String xCol,
```
benmccann, line 46: It seems potentially confusing that there's a parameter `xCol` and `xColumn` and only one is used. It's not clear which takes precedence. Can you remove one of the two parameters? (same with `yCol` / `yColumn`). That would also probably fix the Codacy warning that's being generated

emilianbold: I agree with the above. I've kept postponing commenting on this PR because it seems to me adding more and more `create` methods is not a good solution in a way; the other is that this method seems something that could be perhaps `private` and expose to the users something more clean so they don't have to think about the difference between `xCol` and `xColumn`. Might be simple of getting rid of the `table` argument altogether and just send columns.

rayeaster: @benmccann @emilianbold I updated PR with above comments. Thanks for review.

jsplot/src/main/java/tech/tablesaw/plotly/api/BubblePlot.java
```diff
@@ -38,22 +42,43 @@ public static Figure create(
...
-            .build();
-
-    ScatterTrace trace =
-        ScatterTrace.builder(table.numberColumn(xCol), table.numberColumn(yCol))
-            .marker(marker)
-            .build();
+      String title,
+      Column xColumn,
+      Column yColumn,
+      NumericColumn sizeColumn,
```
emilianbold, line 48: Having columns directly looks much neater to me.

jsplot/src/main/java/tech/tablesaw/plotly/api/BubblePlot.java
```diff
@@ -38,22 +42,43 @@ public static Figure create(
...
+      double[] color,
+      SizeMode sizeMode,
+      Double opacity) {
+
+    Layout layout = Layout.builder(title, xColumn.name(), yColumn.name()).build();
+
+    Marker marker = null;
+    MarkerBuilder builder = Marker.builder();
+    if (sizeColumn != null) {
```
emilianbold, line 57: But you should add a quick javadoc about which parameters are nullable and which not. For example here sizeColumn is nullable, but in the other method sizeCol cannot be null.

jsplot/src/main/java/tech/tablesaw/plotly/api/BubblePlot.java
```diff
@@ -38,22 +42,43 @@ public static Figure create(
...
+    marker = builder.build();
+
+    ScatterTrace trace = ScatterTrace.builder(xColumn, yColumn).marker(marker).build();
     return new Figure(layout, trace);
   }
 
+  public static Figure create(String title, Table table, String xCol, String yCol, String sizeCol) {
+    NumericColumn xColumn = table.numberColumn(xCol);
+    NumericColumn yColumn = table.numberColumn(yCol);
+    NumericColumn sizeColumn = table.numberColumn(sizeCol);
```
emilianbold, line 78: ... and actually here sizeCol could be allowed to be null and just init sizeColumn to null in that case.

emilianbold:
This looks good to me, unless this PR gets some other feedback in the following few days I will  merge it.

benmccann:
Seems fine to me

emilianbold:
With 2 +1s no need to wait.

PR #783, opened 2020-03-20
Fix behavior of INNER and LEFT_OUTER joins with empty left table

vanderzee-anl-gov:
Thanks for contributing.

## Description

Current behavior of tablesaw behaves like a right outer join or full outer join should behave if the left table is empty. Based on review of the source code, this appears to be the case because the normal termination for left outer join and inner join is within a loop over the rows of the first table. This commit handles the special case of an empty left table for left outer join and inner join, correctly returning an empty table with the appropriate columns.

## Testing

Although it should be fairly simple to test this, I did not take the time to understand your testing framework and add a unit test.

### DIFF
--- a/core/src/test/java/tech/tablesaw/joining/DataFrameJoinerTest.java
+++ b/core/src/test/java/tech/tablesaw/joining/DataFrameJoinerTest.java
@@ -2,12 +2,16 @@ package tech.tablesaw.joining;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
-import com.google.common.base.Joiner;
 import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
+
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
+
+import com.google.common.base.Joiner;
+
+import tech.tablesaw.api.StringColumn;
 import tech.tablesaw.api.Table;
 import tech.tablesaw.api.TextColumn;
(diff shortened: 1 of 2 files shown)


benmccann:
> Although it should be fairly simple to test this, I did not take the time to understand your testing framework and add a unit test.

Like you said, I think it should be pretty easy. You can add a test to `DataFrameJoinerTest` and then you can run the tests for just that class with `cd core` and then `mvn -Dtest=DataFrameJoinerTest test`. Let me know if you have any particular questions

vanderzee-anl-gov:
I haven't used maven much before, but I have now added tests that illustrate the difference in behavior with and without my suggested block of code. They fail without my code and pass when it is included.

core/src/test/java/tech/tablesaw/joining/DataFrameJoinerTest.java
```diff
@@ -1288,4 +1292,28 @@ public void differentColumnTypes() {
           table1.joinOn("ID").inner(table2);
         });
   }
+
+  @Test
+  public void innerJoinEmptyLeftTable() {
+    StringColumn animalColumn = StringColumn.create("Animal");
```
benmccann: You could condense the first four lines into two:

```
    Table leftTable = Table.create(StringColumn.create("Animal"));
    Table joined = leftTable.joinOn("Animal").inner(ANIMAL_NAMES);
```

Issue #786, opened 2020-04-04
Column indexOf method


islaterm:
I was trying to get the first index that matched a certain value (like Java List's indexOf method) and I couldn't find anything like that in the column's implementation.

Is that functionality not implemented?

For now I'm transforming the column into a List to use the method, but I think it's too much work for something that shouldn't be so complicated.

If it's actually a missing feature I think that the same implementation as Java Lists would work:

```java
public int indexOf(Object o) {
  Objects.requireNonNull(o);
  for (int i = 0, s = size(); i < s; i++) {
    if (o.equals(get(i)) {
      return i;
    }
  }
  return -1;
}
```

benmccann:
If you just want to find if the value is present, we do have a method `anyMatch`. But if you actually want the index returned that's not a method we have yet. Feel free to send a PR if that's a method you'd like still

islaterm:
Ok! I'll make a PR then.

PR #787, opened 2020-04-11
Added implementation and tests for column's indexOf method.

islaterm:
Should solve #786

Thanks for contributing.

## Description

Added an indexOf method to all column types.
Works the same way as Java's ``List::indexOf(Object)``.

## Testing

A simple test for basic behaviour was added to the ``ColumnTest`` class.

### DIFF
--- a/core/src/main/java/tech/tablesaw/columns/Column.java
+++ b/core/src/main/java/tech/tablesaw/columns/Column.java
@@ -303,7 +303,7 @@ public interface Column<T> extends Iterable<T>, Comparator<T> {
         o1 = o2;
       }
     }
-    return (first ? Optional.empty() : Optional.of(o1));
+    return (first ? Optional.<T>empty() : Optional.<T>of(o1));
   }
 
   /**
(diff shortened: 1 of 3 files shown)


core/src/main/java/tech/tablesaw/columns/Column.java
```diff
@@ -80,7 +81,9 @@ default int countUnique() {
   /**
    * Returns a string representation of the value at the given row.
    *
-   * @param row The index of the row.
+   * @param row
```
benmccann, line 84: can you revert the formatting changes in this file?

If you run `mvn compile` one time it will install a pre-commit hook so that when you run `git commit` it will automatically format the code to match the rest of the codebase

islaterm: Sure!
My bad, I forgot I got my IDE configured to format the code when I save.

benmccann:
There are still a number of formatting changes in this PR that should be reverted. I would expect the Google Code formatter pre-commit hook should do that for you automatically

islaterm:
I did that, but it seems that it didn't fix all of the issues.
Ok, I will manually revert the formatting changes to make sure that the code stays like before.

benmccann:
Thanks!

PR #806, opened 2020-05-17
added responsive attr in config

### LINKED ISSUE #794: responsive chart attribute in config
Plotly has an out of the box responsive flag: https://plotly.com/javascript/configuration-options/#making-a-responsive-chart

We just need to add a boolean `responsive` to the Config class.

avirlrma:
fixes #794

Thanks for contributing.

## Description
added a responsive attribute to config to match Plotty.

## Testing

Yes

### DIFF
--- a/jsplot/src/main/java/tech/tablesaw/plotly/components/Config.java
+++ b/jsplot/src/main/java/tech/tablesaw/plotly/components/Config.java
@@ -19,7 +19,7 @@ public class Config extends Component {
 
   @Override
   public String asJavascript() {
-    return asJSON();
+    return "var config = " + asJSON();
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


emilianbold:
The patch looks fine but it also would have been a good opportunity to get rid of the config_template.html and use JSON like many of the other components are starting do.

Do you think you could look into that too? I can merge without, but the config is so small I think the JSON would be easy. Just make sure where the config is used so you get the JSON text.

avirlrma:
@emilianbold Hey,
I think I have made the changes you asked for, please review.
Thanks.

emilianbold:
Hm, your commit looks good but I suspect config is a bit more tricky since it becomes a separate var on the JS side: `var config = {...}`

I wonder why no test is failing in this case?

emilianbold:
So, if you have the time, it would be interesting to add a test that fails for this situation (missing `var config` somewhere).

The overall fix might be as 'simple' as making:

```java
public String asJavascript() {
 return "var config = " + asJSON();
}
```

... although it is a bit ugly. Other ideas welcome.

avirlrma:
@emilianbold I have made the changes. I overlooked it during the first commit.
Have also added a test for the same.

PR #810, opened 2020-06-01
Replace commons-math3 library with org.hipparchus library

axkr:

## Description

In case someone else would also like to use the hipparchus.org library instead of commons-math3 with tablesaw:
- https://github.com/Hipparchus-Math/hipparchus

You can achieve it with this PR.

What was changed

- the PR uses the same main Java math library as in my projects. 
- reduces the size of libraries by approx: 1 MB
- commons-math3 is only used for `test` scope in `pom.xml`

## Testing

Did you add a unit test?
- no


benmccann:
Thanks. I'm not that concerned with library size though. I'm more concerned about using a library that's been tested by lots and lots of people

PR #818, opened 2020-07-12
exposed the thread pool size variable so clients can set appropriate …

lwhite1:
…values

Thanks for contributing.

## Description

What was changed?

see above.

## Testing

Did you add a unit test?

no.

### DIFF
--- a/saw/src/main/java/tech/tablesaw/io/saw/SawReader.java
+++ b/saw/src/main/java/tech/tablesaw/io/saw/SawReader.java
@@ -74,8 +74,22 @@ public class SawReader {
    * @throws UncheckedIOException wrapping an IOException if the file cannot be read
    */
   public static Table readTable(File file) {
+    return readTable(file, READER_POOL_SIZE);
+  }
 
-    ExecutorService executorService = Executors.newFixedThreadPool(READER_POOL_SIZE);
+  /**
+   * Reads a tablesaw table into memory
+   *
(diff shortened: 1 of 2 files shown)


benmccann:
Did you notice any difference in performance? I wonder if we could get the same benefit by making the default based off the number of CPUs on the machine

lwhite1:
> Did you notice any difference in performance? I wonder if we could get the same benefit by making the default based off the number of CPUs on the machine

I did some testing on my laptop, which has 4 CPUs and a single disk. The pool size did not have much of an impact, but I think there are enough potential variables that it's worth exposing. A person may: 

- have multiple disks accessed via separate channels as is an option on AWS. 
- be reading across a network
- have other jobs going on simultaneously

etc. Also, the rowcount, column types, and number of columns in the table may have an effect. I only tested with one large table. 

I wasn't aware that you could detect the number of CPUs available in Java. I will look into that also

lwhite1:
@benmccann I googled getting number of cores in Java and found helpful info on stack-exchange: https://stackoverflow.com/questions/4759570/finding-number-of-cores-in-java

It seems it's not entirely trivial to do properly. The result from `Runtime.getRuntime().availableProcessors()` is the number of logical cores allocated to the JVM, and the value may change while your program is running. Alternately, you could shell-out and get the physical cores from your OS, but you may not have access to them. I think this is better left to the client code.

If you're ok with the changes I will go ahead and merge this.

benmccann:
Ok, sounds fine

PR #819, opened 2020-07-13
Merged from Master, added a comment, and removed unused code

lwhite1:
Thanks for contributing.

## Description

What was changed?
Merged from Master, added a comment, and removed unused code

## Testing

Did you add a unit test?
No, because the change was removing unused code. Existing tests pass.

### DIFF
--- a/saw/src/main/java/tech/tablesaw/io/saw/ColumnMetadata.java
+++ b/saw/src/main/java/tech/tablesaw/io/saw/ColumnMetadata.java
@@ -14,37 +14,10 @@
 
 package tech.tablesaw.io.saw;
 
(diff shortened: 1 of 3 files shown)


saw/src/main/java/tech/tablesaw/io/saw/ColumnMetadata.java
```diff
@@ -55,7 +27,7 @@
 @Beta
 public class ColumnMetadata {
 
-  private static final ObjectMapper objectMapper = new ObjectMapper();
+  // private static final ObjectMapper objectMapper = new ObjectMapper();
```
benmccann, line 30: we should probably remove this if it's unused instead of commenting it out

Issue #827, opened 2020-07-18
Broken Java Docs link


krystiangorecki:
1. "Conclusion" section of "Getting started with Tablesaw" page contains broken link to "Java Docs".
https://jtablesaw.github.io/tablesaw/gettingstarted#conclusion
2. "Exploring tables" section of "Getting started with Tablesaw" page contains broken link to "plotting".
https://jtablesaw.github.io/tablesaw/gettingstarted.html#exploring-tables

benmccann:
Thanks so much for reporting! If you have any time to address, we'd welcome a PR for this

PR #828, opened 2020-07-19
Fixed javadoc links #827.

krystiangorecki:
Thanks for contributing.

## Description

Few broken links in documentation have been updated. #827

### DIFF
--- a/docs-src/main/gettingstarted.md
+++ b/docs-src/main/gettingstarted.md
@@ -256,7 +256,7 @@ rows by default, but you can also use *table.printAll()*, or *table.print(n)* to
 
 Of course, this is just the beginning of exploratory data analysis. You can also use numeric and visual tools to
 explore your data. These facilities are described in the documentation on statistics and
-[plotting](https://jtablesaw.github.io/tablesaw/userguide/Introduction_to_plotting), respectively.
+[plotting](https://jtablesaw.github.io/tablesaw/userguide/Introduction_to_Plotting.html), respectively.
 
 #### Working with a table's columns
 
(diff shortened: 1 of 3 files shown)


docs-src/main/gettingstarted.md
```diff
@@ -256,7 +256,7 @@ rows by default, but you can also use *table.printAll()*, or *table.print(n)* to
 
 Of course, this is just the beginning of exploratory data analysis. You can also use numeric and visual tools to
 explore your data. These facilities are described in the documentation on statistics and
-[plotting](https://jtablesaw.github.io/tablesaw/userguide/Introduction_to_plotting), respectively.
```
benmccann, line 259: I think the best thing to do here would be to make it a relative link. I.e. use `userguide/Introduction_to_plotting.md` (it will transform the extension when publishing). It looks like none of the existing links are relative, which is really surprising to me. Let's test it with this one and then maybe I'll update the rest

krystiangorecki: Let me do this but I need some advice. I don't know how to test if new links work so you'll have to publish the change. Will another commit make it into this PR or should I close this one and submit another?
Also some javadocs use different URLs. Which one is better?
https://www.javadoc.io/doc/tech.tablesaw/tablesaw-core/latest/tech/tablesaw/columns/dates/DateMapFunctions.html
http://static.javadoc.io/tech.tablesaw/tablesaw-core/latest/tech/tablesaw/columns/dates/DateMapFunctions.html

benmccann: Adding another commit to this PR would be best. You can just do `git push` after adding the new commit and it would show up here

The second URL redirects to the first one, so I'd use the first to avoid the redirect

After you make any changes to the docs you can run `./docs-src/deploy_docs.sh` and it will build some files you can commit as well. Then the docs will be automatically updated when the PR is merged

PR #877, opened 2021-01-26
implemented fix for ambiguous method call exception

lwhite1:
Thanks for contributing.

## Description

Breaking change. The old primitive version did not work (with most compilers). The old methods using object varargs isIn(Number... numbers) and isNotIn(Number... numbers) were replaced with Collection<Number> versions. 

This will break any code that used the Number varargs version, although it had a shaky implementation that probably didn't work for anything other than DoubleColumn. 

Moved the primitive version for double... to the DoubleColumn class.
Added similar primitive methods for the other numeric column types
Added tests
Fixed a test broken by the change

## Testing

Did you add a unit test?
Yes

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/DoubleColumn.java
+++ b/core/src/main/java/tech/tablesaw/api/DoubleColumn.java
@@ -1,13 +1,7 @@
 package tech.tablesaw.api;
 
 import com.google.common.base.Preconditions;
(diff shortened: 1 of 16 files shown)


lwhite1:
Regarding the code duplication complaint from Sonar, the code consists of two short methods that are implemented by both IntColumn and ShortColumn, but nothing else. There is no way to eliminate the duplication that wouldn't make the code harder to maintain, IMO.

PR #879, opened 2021-01-26
Use existing charset when available for creating reader

benmccann:
Thanks for contributing.

### DIFF
--- a/core/src/main/java/tech/tablesaw/io/ReadOptions.java
+++ b/core/src/main/java/tech/tablesaw/io/ReadOptions.java
@@ -31,6 +31,7 @@ import com.google.common.collect.Lists;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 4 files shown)


lwhite1 (approved):
So the idea is to use an InputStreamReader that has been instantiated with the desired Charset?  This looks good to me. It does leave open the question of how to manage this for a File data source, since file doesn't have a charset.  Any thoughts on that?

lwhite1:
Actually, I just realized that the UnivocityParser has parse methods for InputStream and File that take a Charset argument. Is this bypassing that mechanism?

benmccann:
You can use an `InputStreamReader` for a `File` as well. I didn't look at where we instantiate the `UnivocityParser`, but the place that was giving us trouble in the bug report was the column type detection method

lwhite1:
Ok. Let's give it a try

PR #881, opened 2021-01-31
Implementation of control of NumberFormats when writing csv files.

deveborn:
Thanks for contributing.

## Description

Added support for apply numberformats when writing CSV-files. Main logic in CsvWriter and options settings in CsvWriteOptions. My first pull request ever (so I recommend you to review carefully). Any kind of feedback is welcome. 

## Testing

Did you add a unit test?
Yes I did. But first time ever making a test. Check if it looks ok.

### DIFF
--- a/core/src/main/java/tech/tablesaw/io/csv/CsvWriteOptions.java
+++ b/core/src/main/java/tech/tablesaw/io/csv/CsvWriteOptions.java
@@ -5,6 +5,7 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.io.Writer;
 import java.nio.file.Paths;
+import java.text.NumberFormat;
 import java.time.format.DateTimeFormatter;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


deveborn:
These test failings... is that something I should address before further review? What is the preferred JDK version to code for? I guess the failings has to do with the use of Map.of(...) during setup, which was not available in JAVA 8.

lwhite1:
I believe we still code to 8

lwhite1:
Having given this greater consideration, I think @benmccann had the right idea. There should be one way to control output formatting for every column. Even if it's more verbose, it's more important to provide a consistent interface for all column types. 

I will close this and implement that solution.

deveborn:
Larry,
The provided implementation offer that. A map can be used where each column
can have a unique number formatter!  Let me know if you still want me to do
something about this. I have been waiting for guidelines on how to proceed.
But I have full respect if you do not want to use this code for some reason.

Den mån 29 mars 2021 kl 13:46 skrev Larry White ***@***.***>:

lwhite1:
As Ben noted, this only addresses numeric output.  It would be better to
cover all the column types with a single approach I think. The date
formatting code I recently added should be removed as well I think.

deveborn:
Ok. I get it. You want something more like an interface where it is up to
the API user to implement the details. Complement this code with some kind
of StringFormatter is not relevant then. Sounds like you have plans for it
so I guess I leave this work. Let me know if you still think I can
contribute. If you reviewed the code and saw serious shortages in general,
I appreciate if you are honest upon that! I am more or less self-taught and
know that I have much to learn.

Den tis 30 mars 2021 kl 20:14 skrev Larry White ***@***.***>:

> As Ben noted, this only addresses numeric output. It would be better to
> cover all the column types with a single approach I think. The date
> formatting code I recently added should be removed as well I think.
>

PR #889, opened 2021-03-17
Add rolling window result centering

XePeleato:
Pandas-like center parameter; allows centering the result values instead of having a bunch of NaNs at the beginning of the column.

Thanks for contributing.

## Description

It adds the option to center the result of a RollingColumn instead of having all the missing values at the beginning, like pandas does.

## Testing

Did you add a unit test?
Yes.

### DIFF
--- a/core/src/main/java/tech/tablesaw/columns/Column.java
+++ b/core/src/main/java/tech/tablesaw/columns/Column.java
@@ -164,6 +164,10 @@ public interface Column<T> extends Iterable<T>, Comparator<T> {
     return new RollingColumn(this, windowSize);
   }
 
+  default RollingColumn rolling(final int windowSize, final boolean center) {
+    return new RollingColumn(this, windowSize, center);
+  }
+
   String getUnformattedString(int r);
 
   boolean isMissing(int rowNumber);
--- a/core/src/main/java/tech/tablesaw/table/RollingColumn.java
+++ b/core/src/main/java/tech/tablesaw/table/RollingColumn.java
@@ -11,10 +11,16 @@ public class RollingColumn {
 
   protected final Column<?> column;
   protected final int window;
+  protected final boolean center;
 
-  public RollingColumn(Column<?> column, int window) {
+  public RollingColumn(Column<?> column, int window, boolean center) {
     this.column = column;
     this.window = window;
+    this.center = center;
+  }
+
+  public RollingColumn(Column<?> column, int window) {
+    this(column, window, false);
   }
 
   protected String generateNewColumnName(AggregateFunction<?, ?> function) {
@@ -31,7 +37,9 @@ public class RollingColumn {
   public <INCOL extends Column<?>, OUT> Column<?> calc(AggregateFunction<INCOL, OUT> function) {
     // TODO: the subset operation copies the array. creating a view would likely be more efficient
     Column<?> result = function.returnType().create(generateNewColumnName(function));
-    for (int i = 0; i < window - 1; i++) {
+    int bound = center ? (window - 1) / 2 : window - 1;
+
+    for (int i = 0; i < bound; i++) {
       result.appendMissing();
     }
     for (int origColIndex = 0; origColIndex < column.size() - window + 1; origColIndex++) {
@@ -44,6 +52,11 @@ public class RollingColumn {
         ((DoubleColumn) result).append(number.doubleValue());
       } else {
         result.appendObj(answer);
+      }
+    }
+    if (center) {
+      for (int i = 0; i < bound; i++) {
+        result.appendMissing();
       }
     }
     return result;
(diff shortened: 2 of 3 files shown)


core/src/main/java/tech/tablesaw/columns/Column.java
```diff
@@ -164,6 +164,10 @@ default RollingColumn rolling(final int windowSize) {
     return new RollingColumn(this, windowSize);
   }
 
+  default RollingColumn rolling(final int windowSize, final boolean center) {
```
benmccann, line 167: this could use a javadoc for people not familiar with pandas's `center`

benmccann:
so this puts n/a at front and back? I'm curious when you might use this

XePeleato:
Yes, that's exactly what it does. For my specific use case, it comes in handy when working with tables where the index is a date(time);
Say you compute the X-day rolling mean of some data, by using center=True you would label each window at its midpoint, which makes sense to me.

XePeleato:
(I deleted my old comment in order to write a more detailed one)

I use this centering when working with date(time)s indices so that the label is set to the midpoint of the rolling window, here's an example:

Let's say we have this Table
```
LocalDate base = LocalDate.of(2020, 01, 01);
DateColumn index = DateColumn.create("index", IntStream.range(0, 92).boxed().map(base::plusDays));
DoubleColumn data = DoubleColumn.create("data", IntStream.range(0, 92).asDoubleStream());

Table table = Table.create(index, data);
```

```
   index     |  data  |
-----------------------
 2020-01-01  |     0  |
 2020-01-02  |     1  |
 2020-01-03  |     2  |
 2020-01-04  |     3  |
 2020-01-05  |     4  |
        ...  |   ...  |
 2020-03-28  |    87  |
 2020-03-29  |    88  |
 2020-03-30  |    89  |
 2020-03-31  |    90  |
 2020-04-01  |    91  |
```

Now let's calculate a 7-Day rolling mean for this table as usual

`table.addColumns(data.rolling(7).calc(AggregateFunctions.mean));`

```
   index     |  data  |  data 7-period Mean  |
----------------------------------------------
 2020-01-01  |     0  |                      |
 2020-01-02  |     1  |                      |
 2020-01-03  |     2  |                      |
 2020-01-04  |     3  |                      |
 2020-01-05  |     4  |                      |
 2020-01-06  |     5  |                      |
 2020-01-07  |     6  |                   3  |
 2020-01-08  |     7  |                   4  |
 2020-01-09  |     8  |                   5  |
 2020-01-10  |     9  |                   6  |
        ...  |   ...  |                 ...  |
 2020-03-29  |    88  |                  85  |
 2020-03-30  |    89  |                  86  |
 2020-03-31  |    90  |                  87  |
 2020-04-01  |    91  |                  88  |
```

As you can see, the mean corresponding to the period 2020-01-01 to 2020-01-07 is set to the last value of the interval (2020-01-07). Now let's do the same centering the value:

```
   index     |  data  |  data 7-period Mean  |
----------------------------------------------
 2020-01-01  |     0  |                      |
 2020-01-02  |     1  |                      |
 2020-01-03  |     2  |                      |
 2020-01-04  |     3  |                   3  |
 2020-01-05  |     4  |                   4  |
 2020-01-06  |     5  |                   5  |
 2020-01-07  |     6  |                   6  |
        ...  |   ...  |                 ...  |
 2020-03-26  |    85  |                  85  |
 2020-03-27  |    86  |                  86  |
 2020-03-28  |    87  |                  87  |
 2020-03-29  |    88  |                  88  |
 2020-03-30  |    89  |                      |
 2020-03-31  |    90  |                      |
 2020-04-01  |    91  |                      |
```

Now the same value is set to the central point of the interval, this can be useful when working with dates, specially if this data is going to be subject to interpolation,where the results would be way more accurate this way.

I hope I made it clear enough, I'll be happy to answer any other question,
Thank you for your time!

benmccann:
Uhh. This PR looks messed up now. It says 79 files changed

XePeleato:
You're right, I mistakenly pushed other changes to master, it should be fine now.

core/src/main/java/tech/tablesaw/columns/Column.java
lwhite1: apologies if this is duplicating something i wrote elsewhere. I made a comment but don't see it now, so perhaps i didn't save it.

I think the javadoc for this method needs to incorporate a slightly shorter version of the explanation you wrote for the PR. The comment here isn't enough for anyone to really get what it does, but the explanation you wrote is very clear. If you can add that, I think we can resolve this and approve and merge the PR

benmccann:
Can you give a more concrete use case? It's not clear to me why you might want to use this. In the example dataset you shared, you're utilizing data from the future, which makes it useless for prediction, so I'm not sure what you'd use it for. Perhaps there are use cases I'm not aware of though

benmccann:
I'm personally not a big fan of adding more code complication to handle specific use cases that can already be handled rather simply. In this case you could use `lead` or `lag` with `windowSize / 2`

PR #890, opened 2021-03-23
Add support for nanoseconds on DateTimeColumn & InstantColumn

XePeleato:
Thanks for contributing.

## Description

This PR aims to add support for nanosecond precision on `DateTimeColum` & `InstantColumn`. This is achieved by storing them internally the same way Java's `Instant` does (long array for epoch-seconds and int array for nanosecond-of-second).

This is just a draft and does not fully eliminate the dependency on `PackedLocalDateTime`/`PackedInstant`. I'll continue to work on it if it's considered worthy.
`DateTimeColumn` & `InstantColumn` now rely on two arrays instead of one as described above, `DateTimeFilters` have also been modified to remove the dependency on `PackedLocalDateTime`/`PackedInstant`.

## Testing

Did you add a unit test?
I will if the idea gets accepted.


lwhite1:
Hi Eduardo,

I appreciate the effort, but i can't foresee merging this. I don't really
have an interest in more column types as it would just be more things to
maintain.

larry

Issue #822, opened 2020-07-14
Reading data with Excel drops columns with no values.


vortexkd:
When reading an excel with data like this:
```
some_col, my_empty_col, other_col
1,,"hi"
```
"my_empty_col" does not exist on the resulting Table.

For reference the Csv reader seems to work fine.

I think this has something to do with the way columns are added in XlsxReader, the createColumn is inside the cell != null check, so it never gets added?
```
if (cell != null) {
          if (column == null) {
            column = createColumn(headerNames.get(colNum), cell);  // looks like it runs only if cell != null
            columns.set(colNum, column);
            while (column.size() < rowNum - tableArea.startRow) {
              column.appendMissing();
            }
          }
```

**Steps to reproduce:**
```
Table csvTable = Table.read().usingOptions(
                XlsxReadOptions.builder(
                        "my_file.csv")
        );
csvTable.column("my_empty_column"); // works fine

Table excelTable = Table.read().usingOptions(
                XlsxReadOptions.builder(
                        "my_file.xlsx")
        );
excelTable.column("my_empty_column"); // throws error.
```

lwhite1:
Hi vortexkd,

Is a column with no values useful to you? 

If we read it, it would have to be as a StringColumn, which would likely lead to a bug report from another user who intended their empty column to be a column of numbers or dates.

See also #815, #812

vortexkd:
My particular use case:
I'm reading the file, processing the data (adding a bunch of things) and outputting the file again.
And it seems really weird for a column to just magically disappear, especially if it isn't always empty and as a result only disappears sometimes.

I understand that typing the column would then become an issue - but it seems that converting the excel file to Csv and then reading it solves the missing column issue - so if dropping empty columns is the stance you want to take on it as a project, then the CSV reader is working strangely (because it keeps empty columns).

-> Either way, it seems like one could reasonably expect the behaviour to be the same if the file is a CSV or an Excel file.

Maybe this sort of thing should have a setting that can be passed into the   XlsxReadOptions / CsvReadOptions?

lujop:
I think that should be useful to have that configurable in `readOptions`.
To be able to solve this and #815 and #812 I think that something like this should work:

    preserveColumnWithMissingData(String columnName, ColumnType, defaultValue)

radiocold:
I have they same problem, this will be fixed??? I need don't remove column with data empty.

lujop:
[...]

PR #899, opened 2021-04-18
Allow pre-defined empty column types

ArslanaWu:
Thanks for contributing.

## Description

Fix issue #822 by letting user specify the type of empty column in advance. 

## Testing

Add columns-with-empty-column.xlsx, which has one empty column.

Add two test in XlsxReaderTest.java
- testColumnsWithEmptyColumn()
Determine whether the empty columns have not been created when the user does not specify the type of empty columns
- testColumnsWithEmptyColumnAndPreserve()
Determine whether the empty columns are created and the type is correct when the user specifies the type of empty columns


lwhite1:
I'm not sure I agree with the design here for two reasons: 

- I suspect that the existing CsvReadOption columnTypes method could handle this use case if it were added to the Excel read options (or better, to the ReadOption base class). 
- If that's not the case, then _this_ method should be generalized so that it covers the other file types.  

Between the two , I have a strong preference for using the columnTypes method unless that's not feasible for some reason.

ArslanaWu:
I think you are right, sorry for disturbing.

PR #901, opened 2021-04-23
Allow reading of malformed CSV

ChangzhenZhang:
Thanks for contributing.

## Description

add ignoreInvalidRows option in `CsvReadOptions` and `ReadOptions`  to skip invalid csv rows

## Testing

add 2 unit tests

### DIFF
--- a/core/src/main/java/tech/tablesaw/io/ReadOptions.java
+++ b/core/src/main/java/tech/tablesaw/io/ReadOptions.java
@@ -42,6 +42,7 @@ import tech.tablesaw.api.ColumnType;
 public class ReadOptions {
 
   public static final boolean DEFAULT_IGNORE_ZERO_DECIMAL = true;
+  public static final boolean DEFAULT_IGNORE_INVALID_ROWS = false;
 
   private static final List<ColumnType> DEFAULT_TYPES =
       Lists.newArrayList(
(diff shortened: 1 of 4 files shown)


ChangzhenZhang:
@benmccann, I find that I didn't create branch for this issue. What should I do now?

core/src/main/java/tech/tablesaw/io/ReadOptions.java
```diff
@@ -154,6 +157,10 @@ public boolean ignoreZeroDecimal() {
     return ignoreZeroDecimal;
   }
 
+  public boolean ignoreInvalidRows() {
```
lwhite1, line 160: Needs a method comment. 

I'm not sure this is a good method name. It seems there are many ways that a row could be invalid, but the behavior is simply to skip (more descriptive than ignore) rows when the row length differs from the expected number of columns (skipRowsWithInvalidColumnCount()?)

The method comment should describe how the 'valid' number of columns is determined. Is it from the header?, What if there is no header (ie, the noHeader() option is used). Add a test that shows that it works as expected when there is no header.

benmccann:
> I find that I didn't create branch for this issue. What should I do now?

That's okay. You can push additional changes to your `master` and they will show up here. Then when this PR is done you can simply delete and recreate your fork as an easy way to reset your repo and next time you can create a branch

ChangzhenZhang:
@lwhite1 @benmccann, Thanks a lot! I changed the method name and add comments. Also added a test without Header. Could you please check if there is anything else that needs to be changed?

benmccann (approved):

lwhite1 (approved):

PR #903, opened 2021-04-24
Improve innerJoin performance

### LINKED ISSUE #902: Improve innerjoin performance when two tables have different size
As Issue #562 mentioned, currently the join algorithm is sensitive to the order of the tables. For example, if table A(named orders) has 10,000 rows and 2 columns, table B(named customers) has  10,000,000 rows and 2 columns, then  B join A is much lower than A join B.  On my computer, the B join A is 5~6 times slower than the A join B. If join number increases, the case may worsen, because the table on the left will continuously grow, but in the current implementation, the size of the left table bigger, the speed slower.

**I would like to add some simple logic in `private Table joinInternal()` method, and tablesaw will use the right table to do iteration if the right table is smaller than the left table.** In my test, while the time cost of A join B remains the same, the time cost of B join A  is reduced from minutes to 15 seconds, which is a significant performance improvement.

I would like to give a PR, **the logic olny varies when two tables do inner join and  table1 is larger than table2**. I think this part of code has much space to improve, and I would like to do further work to improve the code quality and program performance.

Could you give me some suggestions to this feature? Could I do further work later? @benmccann

DanielMao1:
Thanks for contributing.

## Description

Fix #902 
1. `tech.tablesaw.joining.DataFrameJoiner#joinInternal` method
I add the logic to determine which table will be used to do row iteration when table1 and table2 have different sizes. If table1 has a larger size than table B, table2 will be used to do row iteration. It significantly improves performance when table1 is much larger than table2.
2.`tech.tablesaw.joining.DataFrameJoiner#createMultiColSelection` method
This method is to choose the columns of a table used as the join key.  Originally, only the left table is chosen, **while I add the method parameters `List<Integer> joinColumnIndexes`, so both tables can be chosen to generate the col key index to do join.** 
## Testing
Did you add a unit test?
I added a test case called `tech.tablesaw.joining.DataFrameJoinerPerformanceTest#innerJoinCustomersFirstTimeMeasure`, and you can use this case to compare the performance between big table join the small table and small table join big table.

### DIFF
--- a/core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
+++ b/core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
@@ -94,12 +94,7 @@ public class DataFrameJoiner {
 
     for (Table currT : tables) {
       joined =
-          joinInternal(
-              joined,
-              currT,
-              tech.tablesaw.joining.DataFrameJoiner.JoinType.INNER,
-              allowDuplicateColumnNames,
-              joinColumnNames);
+          joinInternal(joined, currT, JoinType.INNER, allowDuplicateColumnNames, joinColumnNames);
     }
     return joined;
   }
@@ -160,13 +155,7 @@ public class DataFrameJoiner {
    */
   public Table inner(Table table2, boolean allowDuplicateColumnNames, String... col2Names) {
     Table joinedTable;
-    joinedTable =
-        joinInternal(
-            table,
-            table2,
-            tech.tablesaw.joining.DataFrameJoiner.JoinType.INNER,
-            allowDuplicateColumnNames,
-            col2Names);
+    joinedTable = joinInternal(table, table2, JoinType.INNER, allowDuplicateColumnNames, col2Names);
     return joinedTable;
   }
 
@@ -184,7 +173,7 @@ public class DataFrameJoiner {
   private Table joinInternal(
       Table table1,
       Table table2,
-      tech.tablesaw.joining.DataFrameJoiner.JoinType joinType,
+      JoinType joinType,
       boolean allowDuplicates,
       String... table2JoinColumnNames) {
 
@@ -200,9 +189,7 @@ public class DataFrameJoiner {
             result, table1, table2, joinType, allowDuplicates, table2JoinColumnIndexes);
 
     validateIndexes(table1Indexes, table2Indexes);
-    if (table1.rowCount() == 0
-        && (joinType == tech.tablesaw.joining.DataFrameJoiner.JoinType.LEFT_OUTER
-            || joinType == tech.tablesaw.joining.DataFrameJoiner.JoinType.INNER)) {
+    if (table1.rowCount() == 0 && (joinType == JoinType.LEFT_OUTER || joinType == JoinType.INNER)) {
       // Handle special case of empty table here so it doesn't fall through to the behavior
       // that adds rows for full outer and right outer joins
       result.removeColumns(Ints.toArray(resultIgnoreColIndexes));
@@ -211,8 +198,9 @@ public class DataFrameJoiner {
 
     Selection table1DoneRows = Selection.with();
     Selection table2DoneRows = Selection.with();
-    if (table1.rowCount() >= table2.rowCount()
-        && joinType == tech.tablesaw.joining.DataFrameJoiner.JoinType.INNER) {
+    // 2021/04/24 @Daniel Modified: If size of table1 is larger than table2,
+    // use table 2 for row iteration, which can significantly increase performance
+    if (table1.rowCount() > table2.rowCount() && joinType == JoinType.INNER) {
       for (Row row : table2) {
         int ri = row.getRowNumber();
         if (table2DoneRows.contains(ri)) {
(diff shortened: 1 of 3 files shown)


DanielMao1:
I sign up the agreement now @benmccann ,too. Thanks.

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -355,10 +385,21 @@ private Selection selectionForColumn(Column<?> valueColumn, int rowIndex, Index
...
     }
   }
-
-  /** Create a big multicolumn selection for all join columns in the given table. */
+  /**
+   * Create a big multicolumn selection for all join columns in the given table. Joins two tables.
+   *
+   * @param table1 the table which used to generate Selection.
+   * @param ri row number of row in table 1.
+   * @param selectionSize max size in table 1.
```
lwhite1, line 393: It seems to be missing the description of argument 
   
    List<Index> indexes

I don't understand what the description of joinColumnIndexes is telling me. It appears that it should read
   the column indexes (plural) of table1
but what column indexes, and what are they used for?

lwhite1:
Hi @DanielMao1. Thanks for the PR. Did you demonstrate
a) this works on multi-table joins? This may be proven by existing tests, but it would be nice to add a test that we know exercises the new code.
b) the resulting table's column order is the same as it would be without this change?  It seems the fix only changes the row iteration order, but i just want to confirm that the resulting table is identical - at least in terms of column order, to what you would get prior to the change (i.e., table 1's columns appear in order, followed by table two's columns)?

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
DanielMao1: Thanks for your review:) 

> but what column indexes, and what are they used for?


It is the index in tables' column list, representing "join key" for joining. and will be used for cross join and catcat table.

> I don't understand what the description of joinColumnIndexes is telling me. It appears that it should read
the column indexes (plural) of table1


Before I change, it would read the column index of table1; 
After I change, It can be the column index of table2 for better performance. **I will change the variable name and JAVA doc tomorrow for more clear understanding.**

DanielMao1:
> Hi @DanielMao1. Thanks for the PR. Did you demonstrate
> a) this works on multi-table joins? This may be proven by existing tests, but it would be nice to add a test that we know exercises the new code.
> b) the resulting table's column order is the same as it would be without this change? It seems the fix only changes the row iteration order, but i just want to confirm that the resulting table is identical - at least in terms of column order, to what you would get prior to the change (i.e., table 1's columns appear in order, followed by table two's columns)?

Thanks for your response! 
a) yeah. Thanks for your advice. I will add other tests later for multiple joins.
b) It won't change the order of columns. the join order is determined in `crossJoin` method. I did not change such method.
The result table is identical, I will add some tests to prove it later.

core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -198,31 +199,60 @@ private Table joinInternal(
...
-    for (Row row : table1) {
-      int ri = row.getRowNumber();
-      if (table1DoneRows.contains(ri)) {
-        // Already processed a selection of table1 that contained this row.
-        continue;
-      }
-
-      Selection table1Rows = createMultiColSelection(table1, ri, table1Indexes, table1.rowCount());
-      Selection table2Rows = createMultiColSelection(table1, ri, table2Indexes, table2.rowCount());
+    // 2021/04/24 @Daniel Modified: If size of table1 is larger than table2,
```
benmccann, line 202: I would remove "2021/04/24 @Daniel Modified:" from the comment. That's what Git history is for

DanielMao1: Thanks for your review: ) I deleted the comment and push another commit. Please check for the latest version.

DanielMao1:
@lwhite1 @benmccann  
Hello: )
For the latest commit, I: 
###1.added two tests for the larger left table join smaller right table.
[test1](https://github.com/jtablesaw/tablesaw/blob/3d36d4fdd09e964f8b8a97d5d44cc7250ba547f4/core/src/test/java/tech/tablesaw/joining/DataFrameJoinerTest.java#L826) is for single join key and [test2](https://github.com/jtablesaw/tablesaw/blob/3d36d4fdd09e964f8b8a97d5d44cc7250ba547f4/core/src/test/java/tech/tablesaw/joining/DataFrameJoinerTest.java#L892) is for multiple join keys. It tests the column order and final result row number. 
**It shows that the code I modified does not influence the order of columns or multiple join results.**
### 2. fix the comments and Javadoc
If there are other inappropriate codes in the commits, I am looking forward to your suggestions.
Thanks.

core/src/test/java/tech/tablesaw/joining/DataFrameJoinerPerformanceTest.java
```diff
@@ -62,6 +62,26 @@ public void innerJoinCustomersFirst() {
     assertTimeout(Duration.ofSeconds(1), () -> customers.joinOn("customerId").inner(orders));
   }
 
+  @Test
+  public void innerJoinCustomersFirstTimeMeasure() {
```
lwhite1, line 66: I'm not really sure what to do with this test, but I think you should disable it. We don't want to write (more) text strings to System.out. I spent a lot of time removing many that I put in myself. I would suggest using assertions rather than print statements, but I'm concerned that they will fail intermittently given the non-deterministic nature of performance tests. 

As long as you demonstrate correctness, and we "know" it improves performance, we don't need to run this every time. Also, any test that takes a long time to run is likely to break the CI runs.

lwhite1:
Thanks for the changes, @DanielMao1. I think if you disable the performance test, we're good to go.

core/src/test/java/tech/tablesaw/joining/DataFrameJoinerPerformanceTest.java
DanielMao1: Yeah. I agree. I do not want to use assertion to do non-deterministic tests either. It is unstable.
I will delete this test in my PR. Maybe In the future we can find better ways to do performance improvement and test.
Thank you for your careful review : )

DanielMao1:
> Thanks for the changes, @DanielMao1. I think if you disable the performance test, we're good to go.

@lwhite1 I disabled my tests in the performance test.  I will only keep them in my local repository for further performance improvement in the future.  Thanks for your review.

PR #904, opened 2021-04-24
Fix #441 allowing to build project with Windows

### LINKED ISSUE #441: Some unit tests fail on Windows

lujop:
Thanks for contributing.

## Description

Fixes issue #441 allowing to build project with Windows and with systems that don't has en-US as default locale

Changes made:
- Changing hardcoded line separator for `System.lineSeparator()'. Except in HTML writer because Jsoup seems to use always Unix line separator.
- Putting all surefire args in one line because two didn't work and system locale was used.
- Changed mocked URLs because Inet URLs were used as if they were valid local filesystem urls.
- Disabled BOM test. I think that test being failing is correct, at least in systems that don't use UTF-8. I haven't seen any management of BOM in the `tablesaw` code, and `univocity` seems to manage BOM but only when a `File` is passed and not a `Reader` as `tablesaw` uses.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
         java-version: [1.8, 11, 15]
         # Tests seem to be failing on Windows
         # os: [windows-latest, macOS-latest, ubuntu-latest]
-        os: [macOS-latest, ubuntu-latest]
+        os: [macOS-latest, ubuntu-latest, windows-latest]
     steps:
       - name: Checkout
         uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,8 +15,6 @@ jobs:
       matrix:
         # Test the Java LTS versions and latest version available
         java-version: [1.8, 11, 15]
-        # Tests seem to be failing on Windows
-        # os: [windows-latest, macOS-latest, ubuntu-latest]
         os: [macOS-latest, ubuntu-latest, windows-latest]
     steps:
       - name: Checkout
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         # Test the Java LTS versions and latest version available
-        java-version: [1.8, 11, 15]
+        java-version: [1.8, 11, 16]
         os: [macOS-latest, ubuntu-latest, windows-latest]
     steps:
       - name: Checkout
--- a/core/src/test/java/tech/tablesaw/analytic/AnalyticQueryTest.java
+++ b/core/src/test/java/tech/tablesaw/analytic/AnalyticQueryTest.java
@@ -13,6 +13,8 @@ import tech.tablesaw.api.Table;
 
 class AnalyticQueryTest {
 
+  private static final String LINE_END = System.lineSeparator();
+
   @Test
   public void testToSqlString() {
     Table table = Table.create("table1", IntColumn.create("sales"));
@@ -31,17 +33,17 @@ class AnalyticQueryTest {
 
     String expected =
         "SELECT"
-            + System.lineSeparator()
+            + LINE_END
             + "SUM(sales) OVER w1 AS sumSales"
-            + System.lineSeparator()
+            + LINE_END
             + "FROM table1"
-            + System.lineSeparator()
+            + LINE_END
             + "Window w1 AS ("
-            + System.lineSeparator()
+            + LINE_END
             + "PARTITION BY product, region"
-            + System.lineSeparator()
+            + LINE_END
             + "ORDER BY sales ASC"
-            + System.lineSeparator()
+            + LINE_END
             + "ROWS BETWEEN UNBOUNDED_PRECEDING AND UNBOUNDED_FOLLOWING);";
 
     assertEquals(expected, query.toSqlLikeString());
@@ -61,13 +63,13 @@ class AnalyticQueryTest {
 
     String expectd =
         "SELECT"
-            + System.lineSeparator()
+            + LINE_END
             + "MAX(sales) OVER w1 AS salesSum"
-            + System.lineSeparator()
+            + LINE_END
             + "FROM sales"
-            + System.lineSeparator()
+            + LINE_END
             + "Window w1 AS ("
-            + System.lineSeparator()
+            + LINE_END
             + "ROWS BETWEEN CURRENT_ROW AND 1 FOLLOWING);";
 
     assertEquals(expectd, query.toSqlLikeString());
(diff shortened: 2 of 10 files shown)


core/src/test/java/tech/tablesaw/analytic/AnalyticQueryTest.java
```diff
@@ -81,10 +85,14 @@ public void toSqlStringNumbering() {
             .build();
 
     String expectd =
-        "SELECT\n"
-            + "RANK() OVER w1 AS myRank\n"
-            + "FROM myTable\n"
-            + "Window w1 AS (\n"
+        "SELECT"
+            + System.lineSeparator()
```
benmccann, line 89: this seems a little hard to read to me. Can we put each `System.lineSeparator()` on the end of the line instead of putting it on a new line?

lujop: Yes, I don't like also. But it was done by the formatter plugin that is configurated in maven. 
In fact, it separated a lot of lines from the way I initially leave them.
Is it used?

lwhite1: it wouldn't hurt to use a static import so you could just write ```lineSeparator()```. Even that is too verbose, but you can't fight Java names.

benmccann: You could make it even shorter by assigning it to a constant at the top of the file like `NEW_LINE`

One day we'll have Java 13's triple quotes and this will all be cleaner, but it's probably too early for that since Java 11 is the last LTS unless maybe we can just run the tests on Java 13, but compile with Java 11

lujop: Yes I can use static import to reduce size, but about the new line I don't like it too.
But is what made the formatter plugin. I suppose that it sees a string concatenation that is two big and wants every concatenation in its new line.

What do you want me to do?
Only static import?
Reformat by hand trying to use fewer lines and ignoring formatter plugin? In that case are there any formatting rules that I must follow?

benmccann: You shouldn't format by hand. I'm not sure how the other options would look, but they're worth experimenting with

lujop: Ok, then I should do just use a static import on all line separators? Any other thing it's needed to be changed?

benmccann: I would say we should try it and see how it looks

core/src/test/java/tech/tablesaw/io/DataFrameReaderTest.java
```diff
@@ -50,8 +50,7 @@ public void csv() throws IOException {
   @Test
   public void readUrlWithExtension() throws Exception {
     URL url =
-        mockUrlHelper(
-            "http://something.other.com/file.csv", ImmutableList.of("region", "canada", "us"));
+        mockUrlHelper("something.other.com/file.csv", ImmutableList.of("region", "canada", "us"));
```
benmccann, line 53: It looks weird to me that `http://` is missing. I didn't understand the explanation for why it had to be changed. Is there a fix we can make to `mockUrlHelper` or something that will allow `http://` to be used?

core/src/test/java/tech/tablesaw/io/csv/UnicodeBomHandlingTest.java
```diff
@@ -22,6 +23,7 @@
   }
 
   @Test
+  @Disabled
```
benmccann, line 26: Maybe can we change it to read a File instead so that we at least ensure that case is working?

lujop: Maybe I'm wrong but I really think that the problem is not the test.
It's actually `tablesaw` code that fails to manage BOM in non Unicode environments.

Because if there isn't any management code in `tablesaw` code itself the management that does `univocity`, to my understanding,  only works if we call the parser using a File.
But actual code never uses that method and always preprocess the File itself and passes a reader.

To correct that I think that it will be better to create an issue and manage it in another PR because it will require some refactors on file processing and cache system used.

If my diagnostic is correct I think that disabling the test is better because it lies.

benmccann: Ah, ok. I forgot we didn't call the univocity file method even when the input is a file. That makes sense then

core/src/test/java/tech/tablesaw/io/DataFrameReaderTest.java
lujop: Changed the mockUrl to be able to use urls

core/src/test/java/tech/tablesaw/analytic/AnalyticQueryTest.java
lujop: At the end, I use a LINE_END constant that I've seen that was used in some places.

jsplot/src/test/java/tech/tablesaw/components/LayoutTest.java
```diff
@@ -55,58 +58,62 @@ public void testAutosize() {
...
       assertEquals(
-          "var layout = {\n"
-              + //
-              "    autosize: true,\n"
-              + //
-              "\n\n"
-              + //
-              "};\n",
+          "var layout = {"
+              + lineSeparator()
```
benmccann, line 62: this one should be `LINE_END` to match the others and then you no longer need the static `import`

benmccann:
can you rebase this PR? thanks!

benmccann:
Can you try adding Windows to the CI so that we avoid regressing? It should just be changing the line below:

https://github.com/jtablesaw/tablesaw/blob/56508ba0ba6e180b80e1809538a586689c1b963d/.github/workflows/ci.yml#L19

lujop:
> Can you try adding Windows to the CI so that we avoid regressing? It should just be changing the line below:
> 
> https://github.com/jtablesaw/tablesaw/blob/56508ba0ba6e180b80e1809538a586689c1b963d/.github/workflows/ci.yml#L19

Done

.github/workflows/ci.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         java-version: [1.8, 11, 15]
         # Tests seem to be failing on Windows
```
benmccann, line 18: can you remove this line and the next line?

lujop: OMG, I didn't see it.
Done

lujop:
I've seen that build on windows crashes on JAva 15, 
Can I change for 16 that is the lastest now and retry?

benmccann:
Hmm, that's weird. But sure, go ahead and try changing 15 to 16

PR #905, opened 2021-04-25
Fix leftOuterJoin on multiple tables

### LINKED ISSUE #807: leftOuter join will only join to the last table included in the list
If multiple tables are passed to a leftOuter join, only the last table in the list is actually joined in the return value. This is because of the line linked below. joined is reset for each iteration

https://github.com/jtablesaw/tablesaw/blob/ae52d999e0ee1152012ff6e18ac9bd031eacfea4/core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java#L453

Carl-Rabbit:
Thanks for contributing.

## Description

Fix bug of leftOuter join when using multi-tables. Fixes #807 

## Testing

Add 3 test tables to `DataFrameJoinerTest`.
Add 1 test `leftJoinStoreWithMultiTables` for left outer join with multi-tables.
Fix the incurrect test `leftOuterJoinWithDoubles4`.

### DIFF
--- a/core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
+++ b/core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
@@ -14,22 +14,11 @@ import tech.tablesaw.columns.booleans.BooleanColumnType;
 import tech.tablesaw.columns.dates.DateColumnType;
 import tech.tablesaw.columns.datetimes.DateTimeColumnType;
 import tech.tablesaw.columns.instant.InstantColumnType;
-import tech.tablesaw.columns.numbers.DoubleColumnType;
-import tech.tablesaw.columns.numbers.FloatColumnType;
-import tech.tablesaw.columns.numbers.IntColumnType;
-import tech.tablesaw.columns.numbers.LongColumnType;
-import tech.tablesaw.columns.numbers.ShortColumnType;
+import tech.tablesaw.columns.numbers.*;
 import tech.tablesaw.columns.strings.StringColumnType;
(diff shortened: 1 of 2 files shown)


core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java
```diff
@@ -442,7 +431,9 @@ public Table leftOuter(Table... tables) {
   public Table leftOuter(boolean allowDuplicateColumnNames, Table... tables) {
     Table joined = table;
     for (Table table2 : tables) {
-      joined = leftOuter(table2, allowDuplicateColumnNames, joinColumnNames);
+      joined =
+          joinInternal(
```
benmccann, line 435: Why do you have to use `joinInternal` here? Does that mean `leftOuter(table2, allowDuplicateColumnNames, joinColumnNames)` is still broken?

Carl-Rabbit: No. `leftOuter(table2, allowDuplicateColumnNames, joinColumnNames)` works well when joining field `table` with another table `table2`. 

https://github.com/jtablesaw/tablesaw/blob/627be4b660e6ff12eec3c06b7a2c9e353698f911/core/src/main/java/tech/tablesaw/joining/DataFrameJoiner.java#L488-L490

Problem is that field `table` is final and is assigned in the constructor. If we want to join multi-talbes, we must use the join result `joined` in the next iteration. Hence I choose to use `joinInternal`.

benmccann (approved):

Issue #204, opened 2017-12-01
Handle percentages and currency


benmccann:
It could be nice if percentages and dollar amounts were considered numbers instead of strings. I wonder if then we'd also want to store a display format with the column to keep them displaying as percentages and currency in that case.

hswick:
Should support Dogecoin

lwhite1:
Storing format info in the cols could help with numeric formats in general.

lwhite1:
We can now format numbers as percentages and currency. 

Still need to to handle them on input.

Kerwinooooo:
@lwhite1  Hi, developer. I pull a request #906 to support reading percentage from file.

I have modified `canParse()` and `parseDouble()` in *DoubleParser* to parse strings containing "%" to double.
However, I can't find suitable position to set columnformatter of the percent column. Do you have any suggestions about this or just consider percentages to be double by default?

I want to make sure whether it is a feasible way to solve this issue.
Thanks a lot.

lwhite1:
It's hard to know how to handle percents. I'm not sure there's a good way
to automatically modify the column format, and in any case, it currently
only affects println() and not writing to a file.

I guess the best available solution would be to store the values after
dividing by 100. Then (maybe) if the user sets the column format it will
work properly. I would test that, though. It's possible it assumes the
values are already percentages and just need the symbol added.

PR #906, opened 2021-04-25
Support percentage by parsing percentage to double

Kerwinooooo:
Thanks for contributing.

## Description

What was changed

`canParse()` and `parseDouble()` in *DoubleParser*
They are modified to accept the strings that contains number and end with %.
Thecorressponding issue is #204 .

## Testing

Did you add a unit test?

Yes. In *CsvReaderTest* and *DoubleParserTest*


core/src/main/java/tech/tablesaw/columns/numbers/DoubleParser.java
```diff
@@ -24,9 +26,15 @@ public boolean canParse(String s) {
       return true;
     }
     try {
-      Double.parseDouble(AbstractColumnParser.remove(s, ','));
+      if (isPercent(AbstractColumnParser.remove(s, ','))) {
+        s = AbstractColumnParser.remove(s, ',');
+        NumberFormat numberFormat = NumberFormat.getPercentInstance();
```
benmccann, line 31: I think you could combine this line and the next into a single line:
```
NumberFormat.getPercentInstance().parse(s);
```

As it is now, `double d` looks to be unused

Kerwinooooo: Yes, it simplify code. I have modified it. Thx for suggestion.

core/src/main/java/tech/tablesaw/columns/numbers/DoubleParser.java
```diff
@@ -24,9 +26,14 @@ public boolean canParse(String s) {
       return true;
     }
     try {
-      Double.parseDouble(AbstractColumnParser.remove(s, ','));
+      if (isPercent(AbstractColumnParser.remove(s, ','))) {
+        s = AbstractColumnParser.remove(s, ',');
+        Number number = NumberFormat.getPercentInstance().parse(s);
```
benmccann, line 31: remove unused variable
```suggestion
        NumberFormat.getPercentInstance().parse(s);
```

lwhite1:
@benmccann do you happen to know if the use of an appropriate locale in ReadOptions would permit reading numbers with "," as a decimal point? 
If not, that would also require number parsing changes.  If there are other common use-cases, it makes me wonder if this isn't too specialized to add as its own option. Thoughts?

benmccann:
I'm not sure what using locale for "," as decimal separator and then handling percentages have to do with one another. Maybe I'm not understanding?

lwhite1:
I think they're both cases typically handled in Java by DecimalFormat or NumberFormat, so any solution to the "," issue that used a DecimalFormat, would probably also solve the percentage issue.

benmccann:
Ah, okay. I'm not sure if that's supported yet, but it sounds like a good solution to this and would be interesting to test. The majority of European countries use a comma for decimal separator such as Spain, France, Norway, and Denmark.

lwhite1:
I checked the code and Locale is only accessed by Date/Time column parsers. I will try do a little research on number parsing with locales. 

Looking at the changes proposed here, the hard-coded reference to `AbstractColumnParser.remove(s, ',')` seems like a locale-related issue in the making. Unless, I'm missing something, I don't think it should be approved as-is.

lwhite1:
I'm still not ready to approve this. It continues to seem too limited and specialized to get it's own option.
I checked both Pandas and R, neither provide a special hook for percentages. 

- Pandas provides an option for supplying a map of column-names to conversion functions, which it calls 'converters'
- R relies on post-import processing: using a function to convert the String column to a numeric column after the data is loaded, which as @benmccann pointed out elsewhere, we already support using the mapInto() method. 

The Pandas solution is nice and general, it can be used by any columnType to convert the data to any other type. I'm not completely sure we can do that given the added complexity of strong types and generics, but it seems like we should try - if the problem is indeed worth solving.

lwhite1:
@benmccann I'm going to close this (and the other, older version) on the grounds of method proliferation (yes i know i'm not real consistent on that). 

What are your thoughts about making this the default behavior by modifying the column type detection for floats and doubles to consider percentages?

benmccann:
I'd be fine with that

lwhite1:
@Kerwinooooo If you resolve the conflicts I will merge this.  Thank you.

emilianbold:
As long as you made a decision about percentages do comment/close on #703 too please.

Kerwinooooo:
> @Kerwinooooo If you resolve the conflicts I will merge this. Thank you.

Conflict is resolved. Thank you.

Issue #662, opened 2019-09-11
[Question] Ability to assign column types by column name (header) when importing CSV


zsoki:
Hi All,

First of all, really great work on Tablesaw, it's a joy to use so far!

I have a use case for which I yet to find a convenient solution: I have a CSV I need to import, which could have columns I don't need to process. However, I'd like to set the column types of the needed columns explicitly by 
```
val csvReadOptions = CsvReadOptions.builder(inputStream)
                .header(true)
                .columnTypes(columnTypeArray)
Table.read().csv(csvReadOptions)
```
The problems are the following:
- There could be columns missing
- The order of the columns could be different each time
- To increase performance, it would be ideal to skip all unknown columns not defined

This means the only thing I could determine the column types dynamically is if I could map the types to column names (headers).

My question is: is there a built-in functinality in Tablesaw for something like this?
```
val columnTypeMapping = mapOf(
    "column_name_1" to ColumnType.DOUBLE,
    "column_name_2" to ColumnType.STRING
)

val csvReadOptions = CsvReadOptions.builder(inputStream)
    .header(true)
    .columnTypeMapping(columnTypeMapping)
```

The only workaround I've could think of so far is the following:
- Read the first line of the CSV manually to get the headers, create the mapped ColumnType array manually, then reset the InputStream to feed into Tablesaw's CsvReader

I could also read the CSV twice with Tablesaw and get the headers from the Table but it seems a "little bit" wasteful 🙂

Also If you have any other ideas how to achieve this, I'd be glad to hear. Thanks!

lwhite1:
@Suppoze Thanks for the kind words

This is kind of a tricky case because of the variability between file versions. 

I'm not sure I completely understand what you're trying to do, but i might attempt writing a method that:
- reads a "baseline" file. This has the columns you want 
- uses table.structure() to get a table containing the names and types.
- saves the structure table as a CSV file

Then write a second method that:
- reads the baseline structure table.
- get the headers [1] from your new file
- create a column type array to match the new file structure, using the SKIP columnType for any columns you want to skip, and using the data from your structure table to provide the other column types by matching on the headers
- load your file

[1] We have a method to just get the headers for a file, but it's private. It's in FileReader. If you want to submit a PR for making that method public that would be fine.

lwhite1:
@zsoki LMK if you have any additional concerns. Otherwise I will close this in a couple days.

zsoki:
@lwhite1 
Thank you for the detailed response!

[...]

PR #909, opened 2021-04-26
Fixes #822 and #815 providing more extensive columntype options for readers

### LINKED ISSUE #815: No column detected from CSV with no data (header row only)
Hi,

I read a CSV file with no data (header row only) and I noticed that no column was detected. 

The resulting table is empty.

```java
final ByteArrayInputStream inputStream = new ByteArrayInputStream("Column1;Column2;Column3\n".getBytes(StandardCharsets.UTF_8));

final CsvReadOptions readOptions = ((CsvReadOptions.Builder) CsvReadOptions.builder(inputStream)
        .columnTypesToDetect(Arrays.asList(ColumnType.STRING)))
        .header(true)
        .separator(';')
        .build();

final Table table = Table.read().csv(readOptions);

assertThat(table.columns(), hasSize(0)); // OK
```

Did I make something wrong?

Any help would be greatly appreciated.

### LINKED ISSUE #822: Reading data with Excel drops columns with no values.

lujop:
Thanks for contributing.

## Description

Fixes #822 and fix #815 by providing more configurable column type options in `ReaderOptions`.

This options are available globally for all readers and can work in three ways.

 - If no information is provided column types are autodetected as before
 - A complete list of columns can be provided like was previously possible for CSV 
 - Column types can be specified by column name for all columns or for only some. If it's only specified for some autodetect column logic is used for the rest

## Testing
Tests are added for each `Reader` that previously had tests.

That PR depends on #904 as I developed that code on Windows and needed test to be reliable.


lujop:
Made a forced push because in the commit I forgot to upload one new csv test file that I added at the end.

lwhite1:
Hi @lujop, 

Thanks for the PR. 

Can you explain why it is desirable to have both a map option and two function options? What do the function arguments provide that the map and columnType array options do not? 

Is it not possible to separate the Windows-compatibility related changes from the Reader-related changes. This is kind of a big PR. Not necessary if it's a big deal.

On the naming of the additional columnType options in the builder: Is it possible to use a single overloaded name for the (four?) variations with different arguments? This would prevent the user from specifying more than one approach in a single builder, which seems like it could produce unexpected behavior.

lujop:
Thanks @lwhite1, I comment your questions above:


> Can you explain why it is desirable to have both a map option and two function options? What do the function arguments provide that the map and columnType array options do not?

The use of the Map  it's only syntactic sugar to make it more comfortable for simple usage. It could be removed by introducing a map functionhelper or just let the user create one if it wants.

But the functions offer functionality in some use cases that I wanted to support, basically related to processing an input file where you don't know/control all the columns.
That is a real use case that I've encountered in my usage of `tablesaw`

The functions allow to potentially apply a default value for unknown columns, or apply some logic, like the columns that start with 'xxx' must be of type Y.

And about the two functions, one allows you to map partial information, and the other forces you to specify a `ColumnType` for each possible column.
The partial column type customization it's also a use case I encountered and wanted to cover:
When you process a file where you want to specify column types for columns you know, but also can contain other optional columns you don't know and want them to be autodetected.

And about the distinction between complete function and partial one I initially started only with the partial one, but when I made the CSV code I found that if the code knows if it has column type information for all columns it can be a lot more efficient by avoiding column autodetection logic, and then I introduced the other function.
With a bigger refactor it will be able to parse only the header to know if we have all information, but it's a little more complex with the current buffer strategy and to support streaming and all reader scenarios. But I understand can be a good inversion to simplify API and have only the partial function.


> Is it not possible to separate the Windows-compatibility related changes from the Reader-related changes. This is kind of a big PR. Not necessary if it's a big deal.
I try always to do separate PRs, but that I made it together because I needed to run the tests.
But I've the changes in two separate commits, and I can cherry-pick and create a new PR. Or if it's for review you can review the commit for the columntype logic.

 
> On the naming of the additional columnType options in the builder: Is it possible to use a single overloaded name for the (four?) variations with different arguments? This would prevent the user from specifying more than one approach in a single builder, which seems like it could produce unexpected behavior.
Yes I will change the name of the method to `columnTypes` for all. But it won't prevent the user to call all options, will be?
Do you want that I throw an Exception if the user call two variants in same builder?

lwhite1:
Thank you for the explanation. I wouldn't remove any of the functions.

With regard to the separate PRs, i think it's fine the way it is for this
one. The line separator changes are trivial to understand. Don't go to any
extra effort here.

You're right, that it won't prevent the user from calling all options, but
i think it signals that they're to do the same thing. If they have
different names it implies they're more different.

An exception is a good idea if it's not too hard.

benmccann:
I would say let's focus on getting the Windows PR in and then you can rebase this one so that we can review it, which will be a bit easier

lujop:
@lwhite1 Oks, I will add the exception and change the method names.
Only have doubts is about about the map, dou you want to remove it and leave only the functions or not?

lwhite1:
@lujop <https://github.com/lujop> I think the Map api is the simplest way
to set the types for one or two columns, which is a pretty common use-case
for me, especially using something like
```

   Map<String, ColumnType> map = ImmutableMap.of("Column A", DOUBLE,
"Column B", STRING);

```

lwhite1:
@lujop I think having partialColumnTypes() and completeColumnTypes() is fine given the type erasure problem, although I would name them columnTypesPartial() and columnTypesComplete(), as this will make them show up together in name-sorted method lists and in autocomplete popups. 

I'm unconvinced about deprecating the existing method. It works, it's widely used, and it will result in a lot of deprecation warnings. Also, it's the only method that doesn't require you to provide the names, so I'm not sure it would ever be removed. Finally, i sometimes have to force two tables to have the same structure and it's easy to do this like:

```
   Table t2 = Table.read().csv(CsvReadOptions.builder("x.csv")
        .columnTypes(table1.columnTypes())
        ...);
```

The map version would be my goto option when I have only a couple of types I want to control, but I could live without it. I'll just wrap my map inside a function.

core/src/main/java/tech/tablesaw/io/ReadOptions.java
```diff
@@ -327,8 +357,112 @@ public Builder minimizeColumnSizes() {
       return this;
     }
 
+    /**
+     * Provide column types for all columns preventing autodetect column type logic. It's expected
+     * that the array contains all columns
```
lwhite1, line 362: "The array must contain a ColumnType for each column in the table" would be more accurate. An error will be thrown if they don't match up.

core/src/main/java/tech/tablesaw/io/ReadOptions.java
```diff
@@ -327,8 +357,112 @@ public Builder minimizeColumnSizes() {
...
+        Function<String, Optional<ColumnType>> columnTypeFunction) {
+      this.columnTypeFunction = columnTypeFunction;
+      return this;
+    }
+
+    /**
+     * Provide a function that determines ColumnType for all column names. To provide only for some
+     * use {@link #columnTypeByNameFunction(Function)}
+     *
+     * <p>Providing that function prevents running autodetect column type logic
```
lwhite1, line 389: Change 
"<p>Providing that function prevents running autodetect column type logic"
to 
"<p>This method is generally more efficient because it skips column type detection."

core/src/main/java/tech/tablesaw/io/ReadOptions.java
```diff
@@ -327,8 +357,112 @@ public Builder minimizeColumnSizes() {
...
+      return this;
+    }
+
     public ReadOptions build() {
       return new ReadOptions(this);
     }
   }
+
+  /**
+   * Allow to customize read column types. It can work in three ways:
```
lwhite1, line 409: "Allows user to set column types."

core/src/main/java/tech/tablesaw/io/ReadOptions.java
```diff
@@ -327,8 +357,112 @@ public Builder minimizeColumnSizes() {
...
+   *
+   * <ul>
+   *   <li>If no information is provided column types are autodetected
+   *   <li>A complete list of columns can be provided using {@link
+   *       ReadOptions.Builder#columnTypes(ColumnType[])} or {@link
+   *       ReadOptions.Builder#completeColumnTypeFunction} and they are used preventing autodetect
```
lwhite1, line 415: I would say something like "skipping type detection"

"preventing autodetect" sounds almost like a bad thing.

lujop:
@lwhite1 You're right about the deprecations. I think that we can do then:

`columnTypesPartial` for the Map and the partial function
and `columnTypes` for the ColumnType array and the complete function.

It's quite coherent, solve the type erasure problem and no deprecation

core/src/main/java/tech/tablesaw/io/ReadOptions.java
```diff
@@ -327,8 +357,112 @@ public Builder minimizeColumnSizes() {
...
+      this.columnTypesByIdx = columnTypesByIdx;
+      this.columnTypesByNameMap = columnTypesByNameMap;
+      this.columnTypesByNameFunction = columnTypesByNameFunction;
+      this.completeColumnTypesByNameFunction = completeColumnTypesByNameFunction;
+    }
+
+    public Optional<ColumnType> columnType(int columnNumber, String columnName) {
+      Optional<ColumnType> columnType = Optional.empty();
+      if (columnTypesByIdx != null && columnNumber < columnTypesByIdx.length)
+        columnType = Optional.ofNullable(columnTypesByIdx[columnNumber]);
```
lwhite1, line 446: Is this allowing missing columnTypes?  That should be an error. If the user doesn't want a column to be loaded, they should specify ColumnType.SKIP.

core/src/main/java/tech/tablesaw/io/ReadOptions.java
```diff
@@ -327,8 +357,112 @@ public Builder minimizeColumnSizes() {
...
+
+    ColumnTypeReadOptions(
+        ColumnType[] columnTypesByIdx,
+        Map<String, ColumnType> columnTypesByNameMap,
+        Function<String, ColumnType> completeColumnTypesByNameFunction,
+        Function<String, Optional<ColumnType>> columnTypesByNameFunction) {
+      this.columnTypesByIdx = columnTypesByIdx;
+      this.columnTypesByNameMap = columnTypesByNameMap;
+      this.columnTypesByNameFunction = columnTypesByNameFunction;
+      this.completeColumnTypesByNameFunction = completeColumnTypesByNameFunction;
```
lwhite1, line 440: throw Exception if more than one is set? Or do you do that elsewhere?

core/src/main/java/tech/tablesaw/io/ReadOptions.java
```diff
@@ -327,8 +357,112 @@ public Builder minimizeColumnSizes() {
...
+    }
+
+    public Optional<ColumnType> columnType(int columnNumber, String columnName) {
+      Optional<ColumnType> columnType = Optional.empty();
+      if (columnTypesByIdx != null && columnNumber < columnTypesByIdx.length)
+        columnType = Optional.ofNullable(columnTypesByIdx[columnNumber]);
+      if (!columnType.isPresent() && columnTypesByNameMap != null)
```
lwhite1, line 447: Else if would be clearer here. Two of these branches should always provide a non-empty option. That's not clear from the way the code is structured.

core/src/main/java/tech/tablesaw/io/csv/CsvReadOptions.java
```diff
@@ -272,15 +273,7 @@ public Builder allowDuplicateColumnNames(Boolean allow) {
...
-      // Types need to be in certain order as more general types like string come last
-      // Otherwise everything will be parsed as a string
-      List<ColumnType> orderedTypes = new ArrayList<>();
-      for (ColumnType t : EXTENDED_TYPES) {
-        if (columnTypesToDetect.contains(t)) {
-          orderedTypes.add(t);
-        }
-      }
-      this.columnTypesToDetect = orderedTypes;
+      super.columnTypesToDetect(columnTypesToDetect);
```
lwhite1, line 276: why is the sorting skipped here?

core/src/main/java/tech/tablesaw/io/ReadOptions.java
```diff
@@ -327,8 +357,112 @@ public Builder minimizeColumnSizes() {
...
+    public ColumnType[] columnTypes() {
+      return columnTypesByIdx;
+    }
+
+    public boolean canCalculateColumnTypeForAllColumns() {
+      return hasColumnTypeForAllColumns() || completeColumnTypesByNameFunction != null;
+    }
+
+    public boolean hasColumnTypeForAllColumns() {
+      return columnTypesByIdx != null && columnTypesByIdx.length > 0;
```
lwhite1, line 465: this doesn't actually do what the name implies it does. If the array has one value, it would return true regardless of the number of columns.

core/src/test/java/tech/tablesaw/io/fixed/FixedWidthReaderTest.java
```diff
@@ -79,7 +82,7 @@ public void testWithColumnSKIP() throws Exception {
 
     assertEquals(4, table.columnCount());
 
-    String[] expected = new String[]{"Year", "Make", "Model", "Price"};
+    String[] expected = new String[] {"Year", "Make", "Model", "Price"};
```
lwhite1, line 85: if you're going to clean these up, why not just 
```
    String[] expected = {"Year", "Make", "Model", "Price"};
```

lwhite1 (changes requested):

core/src/main/java/tech/tablesaw/io/ReadOptions.java
lujop: In new commit applying suggestions, it will allow only one option throwing an exception if several are used.

core/src/test/java/tech/tablesaw/io/fixed/FixedWidthReaderTest.java
lujop: I don't understand what do you mean.
The change was from the formatted plugin. I ignored changes from not touched files but included the ones from changed files.

lwhite1: You introduced a space between new String[] and the array literal. 

The "new String[]" part is optional, so why not just remove it?

core/src/main/java/tech/tablesaw/io/csv/CsvReadOptions.java
lwhite1: awesome. thanks.

core/src/test/java/tech/tablesaw/io/fixed/FixedWidthReaderTest.java
lwhite1: Just a suggestion. Seems a little cleaner to me to leave it out entirely.

core/src/main/java/tech/tablesaw/io/ReadOptions.java
```diff
@@ -327,8 +358,133 @@ public Builder minimizeColumnSizes() {
...
+      this.completeColumnTypeFunction = columnTypeFunction;
+      return this;
+    }
+
+    /**
+     * Provide a function that determines ColumnType for some column names. To provide for all
+     * column names use {@link #columnTypes(Function)} that generally is more efficient because it
+     * skips column type detection
+     */
+    public Builder columnTypesPartial(Function<String, Optional<ColumnType>> columnTypeFunction) {
```
benmccann, line 393: Do we need both `columnTypes` and `columnTypesPartial`? Could we instead have the reader check if all column names were specified then it will skip detection?

lujop: @benmccann It's what I commented to @lwhite1 in comment 6 days ago.
Technically it's possible but it will require a refactor in the way we buffer and process files and we agreed to maintain that interface differentiating the complete cases and the partial ones.
I think that maintaining the partial is not a bad idea because it has important implications in performance for big files, and making the user aware can be a good thing.

In case we want to suppress the partials, I think that the best will be that in this PR always run type detection logic for all the new cases.  And open an issue to refactor the caching mechanism to allow a more intelligent approach for the new cases when all the column types can be determined.

lwhite1:
@lujop Just wanted to check. You're not waiting on me at this point, are you? Apologies if you are. It looks like you've addressed some of the comments I made but not all of them. I've been assuming another commit was coming.  TBH, it's hard to tell exactly what's changed after you rebased.  

Anyway, if you're waiting on me, I will probably just start fresh since the GitHub UI is kinda useless currently for looking at specific changes.

lwhite1:
FWIW, this PR will also fix #662

lujop:
Yes I was waiting @lwhite1 .
In theory I addressed all, but I suppose I miss something.

core/src/main/java/tech/tablesaw/io/ReadOptions.java
```diff
@@ -204,6 +223,10 @@ public DateTimeFormatter dateFormatter() {
     protected int maxCharsPerColumn = 4096;
     protected boolean ignoreZeroDecimal = DEFAULT_IGNORE_ZERO_DECIMAL;
     private boolean allowDuplicateColumnNames = false;
+    protected ColumnType[] columnTypes;
```
benmccann, line 226: I think we could make this cleaner than having four different ways of deciding. We could have a single interface that's responsible for making the decision and you could have a different implementation based on each of the user options. Then we would only need one variable here

core/src/main/java/tech/tablesaw/io/ReadOptions.java
```diff
@@ -327,8 +358,133 @@ public Builder minimizeColumnSizes() {
...
+        Function<String, Optional<ColumnType>> columnTypesByNameFunction) {
+      this.columnTypesByIdx = columnTypesByIdx;
+      this.columnTypesByNameMap = columnTypesByNameMap;
+      this.columnTypesByNameFunction = columnTypesByNameFunction;
+      this.completeColumnTypesByNameFunction = completeColumnTypesByNameFunction;
+    }
+
+    public Optional<ColumnType> columnType(int columnNumber, String columnName) {
+      if (columnTypesByIdx != null && columnNumber < columnTypesByIdx.length) {
+        return Optional.ofNullable(columnTypesByIdx[columnNumber]);
```
lwhite1, line 466: If columnTypesByIdx, i don't see why this returns ofNullable. Please explain if you think this is correct.

core/src/main/java/tech/tablesaw/io/csv/CsvReadOptions.java
```diff
@@ -115,8 +114,10 @@ public static Builder builder(InputStreamReader reader) {
     return new Builder(reader);
   }
 
+  /** @deprecated Use {@link #columnTypeReadOptions()} */
+  @Deprecated
   public ColumnType[] columnTypes() {
```
lwhite1, line 119: don't deprecate

lujop:
> Hi @lujop. Please make the requested few changes and consider @benmccann's comment on replacing the 4 option fields with an interface.

Sure, thank you for the review. I will also do the suggested change by @benmccann too, I was just waiting to do it all in a row.

core/src/main/java/tech/tablesaw/io/FileReader.java
```diff
@@ -25,42 +27,67 @@
...
   private static final int UNLIMITED_SAMPLE_SIZE = -1;
 
+  /**
+   * @deprecated Use {@link #getColumnTypes(Reader, ReadOptions, int, AbstractParser, String[])} }
+   */
+  @Deprecated
+  public ColumnType[] getColumnTypes(
+      Reader reader, ReadOptions options, int linesToSkip, AbstractParser<?> parser) {
+    return getColumnTypes(reader, options, linesToSkip, parser, null);
+  }
```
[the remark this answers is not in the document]
lujop: I can deprecate, but I think that this should be deprecated as it continues only to maintain binary compatibility as this is a public method.
In fact, I don't know why this method is public, but the correct one to call by now should be the variant with columnNames, because is the one that auto column detection logic will use.
If you call the old one you can get a different result than the ones you get by autodetecting column types.

The ones I remove deprecation were the columnType ones that you pointed in the first review.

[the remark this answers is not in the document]
lujop: Taking a second thought, if it's important to maintain that interface, I can remove the deprecated and change the current implementation making it call the `getColumnNames` method for the columnNames param that now is passed null.

lujop:
@lwhite1 All changes done except one of the deprecated methods, please look at the comments.

PD: I don't know why but it seems that in some places GH is showing old coding without the last commit. If in doubt, please look directly at last commit changes.

lwhite1 (approved):

PR #914, opened 2021-04-28
Printformatting

lwhite1:

## Description

Deprecated support in CsvWriter for writing Date and DateTime columns using Java's DateTimeFormatter directly.

Replaced them with a new option usePrintFormatters(Boolean) that tells CsvWriter to use any printFormatters defined on all types of columns, unifying the output created by printAll() and other toString() variations with the writing of delimited text files. The new approach also handles missing values flexibly.

If this looks ok, I will add support for the other file output types using WriteOptions: html, json, fixed-width. I'm not really sure how the excel writing works, and i'm happy that way.

The sonar failure is sonar being stupid. It says I should remove the deprecated code some day. Which is true, but not a failure or code-smell today. 

## Testing
I added 10 tests

### DIFF
--- a/core/src/main/java/tech/tablesaw/io/csv/CsvWriteOptions.java
+++ b/core/src/main/java/tech/tablesaw/io/csv/CsvWriteOptions.java
@@ -166,12 +166,22 @@ public class CsvWriteOptions extends WriteOptions {
       return this;
     }
 
+    /**
+     * Writes date column output using the given DateFormatter
+     *
+     * @deprecated
+     */
     @Deprecated
     public CsvWriteOptions.Builder dateFormatter(DateTimeFormatter dateFormatter) {
       this.dateFormatter = dateFormatter;
       return this;
     }
 
+    /**
+     * Writes DateTime column output using the given DateFormatter
+     *
+     * @deprecated
+     */
     @Deprecated
     public CsvWriteOptions.Builder dateTimeFormatter(DateTimeFormatter dateTimeFormatter) {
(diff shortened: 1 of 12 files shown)


core/src/main/java/tech/tablesaw/io/csv/CsvWriteOptions.java
```diff
@@ -159,16 +166,37 @@ protected Builder(OutputStream stream) {
...
+     * @deprecated
+     */
+    @Deprecated
     public CsvWriteOptions.Builder dateTimeFormatter(DateTimeFormatter dateTimeFormatter) {
       this.dateTimeFormatter = dateTimeFormatter;
       return this;
     }
 
+    /**
+     * Sets the usePrintFormatters option @see{ColumnPrintFormatter} When true, printFormatters will
```
benmccann, line 192: I've never seen `@see` used like this. I think the standard way to do this might be `{@link ColumnPrintFormatter}`

core/src/test/java/tech/tablesaw/io/csv/CsvWriterTest.java
```diff
@@ -54,7 +58,135 @@ public void dateFormatter() throws IOException {
   }
 
   @Test
-  public void dateTimeFormatter() throws IOException {
+  void printFormatter1() throws IOException {
```
benmccann, line 62: could we give these more descriptive names like `printFormatter_number`, `printFormatter_boolean`, etc.

benmccann (approved):

PR #916, opened 2021-04-29
methods that parse stringColumns into the various numeric cols

lwhite1:

## Description

Added map functions that parse StringColumn and TextColumn values, returning one of the 5 numeric column types (LONG, SHORT, INT, DOUBLE, FLOAT).  These functions can be used to convert anything of a numeric nature (like currencies, percents, etc) or more generally, to extract numbers from strings, into an appropriate numeric column type using a user-provided parse function. 

## Testing

Tests were added for each of the new methods.

### DIFF
--- a/core/src/main/java/tech/tablesaw/columns/strings/StringMapFunctions.java
+++ b/core/src/main/java/tech/tablesaw/columns/strings/StringMapFunctions.java
@@ -21,6 +21,9 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.function.Function;
+import java.util.function.ToDoubleFunction;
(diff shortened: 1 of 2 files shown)


benmccann:
Don't we already have `map`/`mapInto` methods that allow for this?

lwhite1:
@benmccann AFAIK, we have methods like `col.parseInt()` which apply a general approach (e.g. calling Integer.parseInt(string)), but not methods that allow you to provide your own parsing logic. 

If those methods exist, LMK and I will withdraw the PR. I don't want to add useless code.

lwhite1:
nevermind, I see you are correct. closing.
so many methods...

PR #926, opened 2021-05-05
Fix #887

### LINKED ISSUE #887: XlsxReader doesn't respect calculated tableArea for header column names
There is a buggy condition on some excels when there are blank columns at the start of the sheet.

When `es.decide.toolbox.util.XlsxReader#createTable` calculates header names it uses a `for(Cell cell : row) ` instead of using the supplied `tableArea` parameters.

And when there are some blank columns it doesn't always work. I don't know exactly why, but sometimes the first blank column is discarded in the iterator, but sometimes although all values are blank poi maintains the column, and the code doesn't work correctly detecting the header column names.

To calculate the header column names the code should use the `tableArea.startColumn` and `tableArea.endColumn`

lujop:
Thanks for contributing.

## Description
Fixes #887 doing a refactor on header column parsing

## Testing
The `columns.xlsx` file was modified by adding a numeric cell format in one cell in the first column.
That change makes prior code fail the `testColumns` test and demonstrates that it can manage that case with current code.

Note: there is some reformatted code in the PR that is done by automatic formatted plugin.


benmccann (approved):
I didn't look very closely, so maybe Larry can take more of a look but this seems like a fairly safe change to me

PR #928, opened 2021-05-06
Fixes #927 - Allow default parsing to be overridden per column.

### LINKED ISSUE #927: Enhance to customize parsing per column.
Tablesaw currently allows column data parsing to be customized via two strategies:

- Each column type defines a DEFAULT_PARSER that is used by the Column appendCell(value) method.  The list of strings to be interpreted as missing values can be changed on a default parser so that the changes apply to all columns of that type.  Many Tablesaw methods that add new data to columns call the appendCell(value) method, thereby "baking in" default parsing behavior.
- A custom parser can be provided in the appendCell(value, parser) and set(value, parser) methods.

What is proposed is for each Column instance to define its own parser, and this parser shall be used by the appendCell(value) method.  This change shall allow users to define custom parsers for arbitrarily complex text input, and limit the scope of such parsing to individual columns without having to resort to specifying the custom parser per append/set call.

The existing column constructors shall initialize the parser value to the DEFAULT_PARSER, so that the enhancement does not impact existing functionality.

Analogous to the name() and setName() methods, column instances shall also define a parser() method that returns the parser value, and a setParser(parser) method that updates the parser value.

jglogan:
## Description

This PR fixes issue #927 with the following changes:
- This enhancement defines parser() and setParser() methods on the Column interface, and adds a parser argument to the AbstractColumn constructor.
- All column types have been updated to call super with the type-appropriate DEFAULT_PARSER, to preserve functional compatibility with prior releases.
- All column appendCell() methods now parse input using the parser field value, instead of a hardcoded DEFAULT_PARSER.

## Testing

Added unit tests for simple override of parsing strategy for each column type.

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/BooleanColumn.java
+++ b/core/src/main/java/tech/tablesaw/api/BooleanColumn.java
@@ -70,7 +70,7 @@ public class BooleanColumn extends AbstractColumn<BooleanColumn, Boolean>
   private BooleanFormatter formatter = new BooleanFormatter("true", "false", "");
 
   private BooleanColumn(String name, ByteArrayList values) {
-    super(BooleanColumnType.instance(), name);
+    super(BooleanColumnType.instance(), name, BooleanColumnType.DEFAULT_PARSER);
     data = values;
   }
 
@@ -304,7 +304,7 @@ public class BooleanColumn extends AbstractColumn<BooleanColumn, Boolean>
 
   @Override
   public BooleanColumn appendCell(String object) {
-    return append(BooleanColumnType.DEFAULT_PARSER.parseByte(object));
+    return append(parser().parseByte(object));
   }
 
   @Override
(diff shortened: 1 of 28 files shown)


jglogan:
Notes:
- My code formatter got a little busy adding `@Override`, and it removed a couple casts it deemed unnecessary.  I have no problem backing out that diff clutter if you prefer!
- In the discussion on the previous PR, @benmccann mentioned adding custom parser arguments to create() calls.  I have not yet done this, for two reasons.
  - First, I wasn't sure which create calls - Column, ColumnType, both, and on which calls?  Looking at all the variants made me worry about method proliferation, so I wanted to hold off on any of that work until after review and discussion.
  - Second, it seemed to me that method chaining after construction might be sufficient to support all use cases without adding create() methods.  We could do this PR that allows custom parsing via method chaining, and if we need parser in a create() method, we can follow on with another PR.

lwhite1:
Looks pretty good to me. In hindsight, we should have provided a builder interface rather than all these create methods.  My biggest concern with this code is (a) did you miss any create methods, so that the parser may be null at some point, and (b) will someone do that in the future when adding or modifying a create method.  I agree that adding custom parser options to each create is overkill.

jglogan:
Rebased to current main and force-pushed.

PR #931, opened 2021-05-09
Fixes #751 and #882

### LINKED ISSUE #751: Using XlsxReader Ignore column types to detect option and causing undesirable behavior.
Hi, 
First of all, thanks for this awesome package.

I'm using the XlsxReader to import excel file, in one of my tests I came across the following issue:
When the reader auto-detect the column types and the first column value is empty,  it automatically detects it as a String Column and creates a String typed column, till here everything is fine.
When any other value in that column is a Numeric type(in my case all of them) it gets ignored and instead, it returns an empty value.
So, as a result, I get a table with a column that all of the values are empty.
I would except that I get all the values at least as a string and not empty value.

The logic is in the   ``` XlsxReader.private Column<?> appendValue(Column<?> column, Cell cell)```.

I try to solve it by using the  "specifying the datatypes for each column" approach 
I did the following : 
```Java 
Source source1 = new Source(new ByteArrayInputStream(byteArrayOutputStream.toByteArray()));
List<ColumnType> typesTotalData = Arrays.asList(new ColumnType[] {ColumnType.STRING, ColumnType.STRING, ColumnType.STRING, ColumnType.STRING, ColumnType.INTEGER, ColumnType.INTEGER});
Table table = Table.read().usingOptions(XlsxReadOptions.builder(source1).sheetIndex(0).columnTypesToDetect(typesTotalData));
```
Then I realize that there is no use in the columnTypedToDetect in the XlsxReader.

*Another suggestion, it will be good if I'm using multiple sheets and using ```public List<Table> readMultiple(XlsxReadOptions options) throws IOException ```
That I can pass a list of options that each one of them is bounded to a specific sheet, Because in the current situation if I want for example different name ( or different column type list) for each sheet/table I need to perform a multiple reads ( as the number of the sheets ) and set the options for each one even though under the hood for each call it goes over all the sheets and return the sheet that correspond the index the user pass in the options.

I 'm using the latest version 0.37.3.

Thanks, 
Nadav

### LINKED ISSUE #882: XlsxReader casting integers cells to blanks
Hello devs,

Thanks for making/maintaining this library. I am currently in a microservice that validates the data in excel spreadsheets.

I'm reading an excel workbook with multiple sheets in tablesaw-excel. 
```
public static List<Table> readXlsx(InputStream inputStream) throws IOException {
    Source source = new Source(inputStream);
    XlsxReadOptions options = XlsxReadOptions.builder(source).build();
    XlsxReader reader = new XlsxReader();
    return reader.readMultiple(options);
  }
```
I have cell types in Excel of both numeric and text in a column. When the table is read in, the numeric cell types become blank cells. Therefore, when I try to pull up missing values I get the rows where the numeric type cells are at
```
Table missing = table.where(table.column("myColumn").isMissing());
```
Is there some way I can change the behavior of this using XlsxReadOptions? Or is there another method I can use to avoid this behavior?

Thanks!

lujop:
Thanks for contributing.

## Description

Column type detection uses the same logic as before but all column cell types are used instead of only the first one.
If all are the same type it's used, and if not string is used to be able to accommodate all values.

This fixes #751 and with the combination of 74a54aa573a0a881add7a637a4d5f80941520d2e fixes #882 also

## Testing
Existing tests were extended to cover new behavior and use cases from #751 and #882


lujop:
The CI fails seems to be temporary ones due to network problems.
Can some one relaunch the suite?

lwhite1:
i don't know anything about the Excel reader code, but why not reuse the value-based approach used for text files?

lwhite1:
i will bow to your expertise.  I will merge after the conflicts are resolved.

Issue #509, opened 2019-04-03
asd support for violin plots


lwhite1:

LUUUAN:
I would like to provide a PR for this feature.

lwhite1:
That would be great

lwhite1:
Implemented by @LUUUAN

PR #936, opened 2021-05-17
Add support for violin plot

LUUUAN:
Thanks for contributing.

## Description
This is the fix for issue #509  

Two files are added: `ViolinPlot.java` and `ViolinTrace.java`. Also I modified the file `trace_template.html` so that violin plot can draw the inner box plot at the same time. Also, considering that the mean line of violin plot is also frequently used, it is also added into the `trace_template.html` but I comment the code (`context.put("meanLine", "{visible: true}")`) that draw the mean line in `ViolinTrace.java` because it doesn't perform well together with inner box plot. You can changed the code and style according to your preference.

## Testing

Test is developed following the style of other plot tests.

### DIFF
--- a/jsplot/src/main/java/tech/tablesaw/plotly/api/ViolinPlot.java
+++ b/jsplot/src/main/java/tech/tablesaw/plotly/api/ViolinPlot.java
@@ -1,0 +1,23 @@
+package tech.tablesaw.plotly.api;
+
+import tech.tablesaw.api.Table;
+import tech.tablesaw.plotly.components.Figure;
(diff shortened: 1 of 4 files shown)


lwhite1:
the box and meanline options should be added to the builder rather than always including one and excluding the other

LUUUAN:
Ok

LUUUAN:
I have added two method to `ViolinTrace.java` to configure whether box plot and mean line will show up in the figure.

lwhite1:
@LUUUAN This looks good. Thank you

Issue #812, opened 2020-06-11
is it possible to auto fill the missing column when import csv with columns of different sizes?


starli-sl:
Issue:
When import a csv file with column of diffrent sizes, it will return an AddCellToColumnException.
Do the same opreation in pandas, it will fill the missing column with 'NaN' automatically.

Error message:
tech.tablesaw.io.AddCellToColumnException: Error while adding cell from row 551 and column time(position:0): Row number 551 contains 2 columns. 5 expected.
Caused by: java.lang.IndexOutOfBoundsException: Row number 551 contains 2 columns. 5 expected.
	at tech.tablesaw.io.FileReader.addRows(FileReader.java:161)

Test data:
time,c1,c2, c3,c4
2018/7/25 14:30,6,2.57,8.5,10.41
2018/7/25 16:30,9,3.58,6.5,11.41
2018/7/25 18:30,122.5
2018/8/4 8:30,155

Suggestion:
Add a option to auto fill the missing column when import csv.

Thanks

ChangzhenZhang:
@lwhite1 Hi, I pull a [request](https://github.com/jtablesaw/tablesaw/pull/937) for this issue.
I add an option `autoFillMissingColumn` to fill the missing column with the Missing Value when import file with columns of different sizes automatically. And I want to make sure that if this change fixes the issue.
Thanks!

lwhite1:
Closing as "won't fix"

PR #937, opened 2021-05-18
Fix issue 812

ChangzhenZhang:
Thanks for contributing.

## Description

Fix [issue #812](https://github.com/jtablesaw/tablesaw/issues/812)
Add an Read Option `autoFillMissingColumn`. Allow the reader to fill the missing column with the Missing Value when import file with columns of different sizes automatically.   
This option only allow to fill the last missing column of each row with `null`.

## Testing

Add two tests for testing this option.

### DIFF
--- a/core/src/main/java/tech/tablesaw/io/FileReader.java
+++ b/core/src/main/java/tech/tablesaw/io/FileReader.java
@@ -7,14 +7,7 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Streams;
 import com.univocity.parsers.common.AbstractParser;
 import java.io.Reader;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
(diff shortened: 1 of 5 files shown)


benmccann:
You can already do this with `set` and `isMissing`. I'm not sure it's a good idea to add a second way to do it that works only with CSVs

lwhite1:
To add to what @benmccann said, I don't think we should fix this.  A row should either be correct or it should be skipped (as added by your last PR #901). We shouldn't try to handle arbitrarily broken CSV. CSV files that are broken can be fixed in Excel or by hand-coding solutions (e.g., by ensuring the row has enough commas in it). 

in any case, a fix that only works on the last column of the table being imported is not appropriate. If the user is capable of moving the broken column to the end of the row, they're capable of fixing the issue. Solutions should be general or they're not worth doing. 

I appreciate the effort by I'm going to close this, and also close the issue.

PR #962, opened 2021-07-21
fix(table): method 'where' apply 2 times selection function

zhiwen95:
Thanks for contributing.

## Description

fix(table): method 'where' apply 2 times selection function

## Testing

Did you add a unit test?

No


lwhite1:
Hi, Could you update the description with the number for the issue that you're addressing here?

Also, i will ask you to provide a test that shows the fix works, unless that's very difficult to do for some reason. 

thanks.

lwhite1:
yeah, never mind the prior comments. The change is self-explanatory

PR #970, opened 2021-08-16
Shallow cloning on travis

YunLemon:
According to [https://docs.travis-ci.com/user/customizing-the-build#git-clone-depth](url), Travis CI provide a way to shallow clone a repository. This has the obvious benefit of speed, since you only need to download a small number of commits.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: java
 
 # for sonarcloud issue tagging
 git:
-  depth: false
+  depth: 3
 
 addons:
   sonarcloud:


lwhite1:
Hi Yun, Can you explain why this change is a good idea for this project? I'm not overly keen on making changes to something that seems to work reasonably well.

benmccann:
We should get rid of Travis. They've had some major security issues. Most projects I've used have moved to Github Actions

lwhite1:
That certainly sounds reasonable to me.

Issue #972, opened 2021-08-17
Trying to add BigDecimal column


daugasauron:
Need the precision so tried implementing it. 

In TableSliceGroup, aggregate(ListMultimap<String, AggregateFunction<?, ?>> functions) it seems to me that there is some assumption about auto boxing "if (result instanceof Number)" fails because BigDecimal extends Number.

lwhite1:
I think you would find that it is a non-trivial task to add a new column type, unfortunately, but let me know how it goes.

daugasauron:
I have a quite small project with a lot of tables and simple aggregations, was quite straightforward to add the column type for what I need (with a lot of UnsupportedOperationExceptions..) but that if statement is quite annoying. I created my own version of tablesaw with the following and everything works like a charm. 

Would appreciate being able to use the official version.

`          
Object result = function.summarize(subTable.column(columnName));
if (firstFunction) {
  groupColumn.append(subTable.name());
}
if (function.returnType().equals(ColumnType.DOUBLE)) {
  Number number = (Number) result;
  resultColumn.append(number.doubleValue());
} else {
  resultColumn.append(result);
}
`
In my understanding it seems more reasonable since the return type will be determined by the AggregateFunction, and the standard numerical aggregation functions will return double column.

lwhite1:
ok. I merged your PR.  Thanks

PR #973, opened 2021-08-18
Aggregate by non-primitive column type that extends Number

daugasauron:
#972

Thanks for contributing.

## Description

What was changed

## Testing

Did you add a unit test?
No


lwhite1:
This looks harmless at first glance, but I will sit on it until there is an implementation of a non-primitive type.

Issue #1134, opened 2022-08-09
ICU4J vulnerability


kzmv:
Hello,
I want to use tablesaw but the underlying dependencies are flagged for [vulnerability](https://nvd.nist.gov/vuln/search/results?form_type=Advanced&results_type=overview&isCpeNameSearch=true&seach_type=all&query=cpe:2.3:a:icu-project:international_components_for_unicode:65.1:*:*:*:*:c%5C/c%5C%2B%5C%2B:*:*): 

Upgrade to 68.2 should fix the issue. Happy to raise a PR.

However, I don't see the library being used anywhere in the project or is it a transitive dependency?

lwhite1:
it's used, IIRC, in the io package for dealing with unusual encodings.
If that's the case, a PR would be appreciated..

kzmv:
@benmccann When can a new version be released? Can I help with that?

PR #1135, opened 2022-08-10
fix: (#1134) Update icu4j package since it is flagged for vulnerabili…

kzmv:
…ties by package validators

Thanks for contributing.

## Description

A version update to a later version of icu4j as the previous one was being flagged for vulnerabilities

## Testing

No test was needed


kzmv:
Didn't see specific pom config for the bumping version. Could you advise?

PR #1145, opened 2022-09-05
Support for multiple grouped / aggregated columns in pivoting

mattjubb:
Thanks for contributing.

## Description

 - Added support for multiple grouping / aggregation columns via the pivot method. kept the old method contract as is but now delegates to shared logic
 - renamed column1/column2/column3 to groupedColumn/pivotColumn/aggregatedColumn to make it more
   obvious what the meaning of each argument is
- not entirely sure the approach of concatenating the pivot value with aggregated column name ie "2012.Score" is the best, open to suggestions here.

Note: first time contributing to a project on github so let me know if i've done something wrong

## Testing

yes added tests for 3 use cases 1) both multiple grouping and aggregation columns 2) multiple grouping and single aggregation 3) single grouping and

### DIFF
--- a/core/src/main/java/tech/tablesaw/api/Table.java
+++ b/core/src/main/java/tech/tablesaw/api/Table.java
@@ -864,19 +864,41 @@ public class Table extends Relation implements Iterable<Row> {
     return newTable;
   }
 
-  /**
+
+  public Table pivot(
+    List<CategoricalColumn<?>> groupingColumn,
+    CategoricalColumn<?> pivotColumn,
+    List<NumericColumn<?>> aggregatedColumn,
+    AggregateFunction<?, ?> aggregateFunction) {
+    return PivotTable.pivot(this, groupingColumn, pivotColumn, aggregatedColumn, aggregateFunction);
+  }
+
+  public Table pivot(
+      List<String> groupingColumnNames,
+      String pivotColumnName,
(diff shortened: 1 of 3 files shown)


core/src/main/java/tech/tablesaw/api/Table.java
```diff
@@ -864,35 +864,87 @@ public Table dropWhere(Selection selection) {
...
+   * @param aggregateFunction
+   * @return
+   */
+  public Table pivot(
+      List<String> groupingColumnNames,
+      String pivotColumnName,
+      List<String> aggregatedColumnNames,
+      AggregateFunction<?, ?> aggregateFunction) {
+        return pivot(
+            groupingColumnNames.stream().map(this::categoricalColumn).collect(Collectors.toList()),
```
benmccann, line 912: Codacy is complaining that you're doing `Collectors.toList` here when `toList` is already statically imported. I'm fine either way to want it as long as it's consistent

frankwondon:

core/src/main/java/tech/tablesaw/api/Table.java
mattjubb: fixed, thanks

mattjubb:
Could i get another review on this please? addressed previous feedback

PR #1149, opened 2022-09-21
Fix not reading when the entire row is 0 (jtablesaw#1147)

mo20053444:
Thanks for contributing.

## Description

The underlying Apache library reads a blank cell as a 0. Therefore, when the value in a cell is 0, the previous logic cannot distinguish between a blank cell and a 0, which leads to ruling out the entire row where all cells are 0.

## Testing

No new tests added.

### DIFF
--- a/excel/src/main/java/tech/tablesaw/io/xlsx/XlsxReader.java
+++ b/excel/src/main/java/tech/tablesaw/io/xlsx/XlsxReader.java
@@ -147,7 +147,7 @@ public class XlsxReader implements DataReader<XlsxReadOptions> {
       case NUMERIC:
         if (DateUtil.isCellDateFormatted(cell)
             ? cell.getDateCellValue() != null
-            : cell.getNumericCellValue() != 0) {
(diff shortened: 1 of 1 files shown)


excel/src/main/java/tech/tablesaw/io/xlsx/XlsxReader.java
```diff
@@ -147,7 +147,7 @@ private Boolean isBlank(Cell cell) {
       case NUMERIC:
         if (DateUtil.isCellDateFormatted(cell)
             ? cell.getDateCellValue() != null
-            : cell.getNumericCellValue() != 0) {
+            : cell.getNumericCellValue() != 0 || cell.getCellType() == NUMERIC) {
```
benmccann, line 150: `cell.getCellType() == NUMERIC` will always be `true` here, so this is the same as rewriting the entire expression to:
```
if (!DateUtil.isCellDateFormatted(cell) || cell.getDateCellValue() != null)
```

frankwondon:

PR #1197, opened 2023-03-27
Name to lowercase in renameDuplicateColumnHeaders

s1ny1998:

renameDuplicateColumnHeaders when allowDuplicateColumnNames is true is case sensitive which causes validateColumn(final Column<?> newColumn) to fail and throw IllegalArgumentException. 

What was changed

Used toLowerCase while using HashMap.
Also thought of using treeMap, but get() in treeMap has timecomplexity of O(log(n))

Did you add a unit test?
No


ccleva:
Hi @s1ny1998. A fix based on your code is available in the [new maintenance release](https://github.com/tlabs-data/tablesaw/discussions/33) from the maintenance fork. See [this discussion](https://github.com/jtablesaw/tablesaw/discussions/1261) for details.

PR #1203, opened 2023-04-18
1202 fixed, by using parseCaseInsensitive() function

SahibYar:
Thanks for contributing.

## Description
Through out the project DateTimeFormatter.ofPattern(pattern) is being used, This doesn't caters DateTime formats with alphabets like am/AM, pm/PM or Jan/JAN, Feb/FEB etc. And because of that some test cases were failing in `tablesaw-core`

So I used `parseCaseInsensitive` for taking care of such cases. 

## Testing

I didn't added the test case, but ran `tablesaw-core` and all of them are working fine now. 
image


SahibYar:
CC: @benmccann @lwhite1 Please review this PR when you have time.

benmccann:
lgtm. what do you think about renaming `caseInsensitiveDTFormatter` to `caseInsensitiveFormatter`? I think the "DT" part is already covered by it being part of the `DateTimeParser` class. or really I think it could probably even be just `newFormatter` and add a JavaDoc explaining that it's case-insensitive

SahibYar:
Thank you for your guidance, Updated PR as per your comments.

PR #1217, opened 2023-06-11
Added ability to use a custom mapper for serialization

howard-3:
Thanks for contributing.

## Description

Moved the object mapper instance to the json writer options class and allow it to be configurable.

## Testing

Did you add a unit test? yes.

### DIFF
--- a/json/src/main/java/tech/tablesaw/io/json/JsonWriteOptions.java
+++ b/json/src/main/java/tech/tablesaw/io/json/JsonWriteOptions.java
@@ -1,6 +1,9 @@
 package tech.tablesaw.io.json;
 
 import java.io.Writer;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import tech.tablesaw.io.Destination;
 import tech.tablesaw.io.WriteOptions;
 
@@ -8,11 +11,13 @@ public class JsonWriteOptions extends WriteOptions {
 
   private final boolean asObjects;
   private final boolean header;
+  private final ObjectMapper mapper;
 
   private JsonWriteOptions(Builder builder) {
     super(builder);
     this.asObjects = builder.asObjects;
     this.header = builder.header;
+    this.mapper = builder.mapper;
   }
 
   public boolean asObjects() {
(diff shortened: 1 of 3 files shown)


benmccann:
It's not clear to me that this really make anything easier? You can already convert a table to JSON without anything being built into TableSaw. E.g. here's some code I use to convert a table to JSON to be consumed by Chart.js:
```
  public static JsonNode toChartJsJson(int startIdx, String ticker, InstantColumn dates, DoubleColumn... columns) {
    ObjectNode data = jackson.createObjectNode();

    long[] timestamps = dates.asEpochMillisArray();

    ArrayNode datasets = jackson.createArrayNode();
    for (int i = 0; i < columns.length; i++) {
      ObjectNode dataset = jackson.createObjectNode();
      dataset.put("label", ticker != null && i == 0 ? ticker : columns[i].name());
      ArrayNode dataArray = jackson.createArrayNode();
      dataset.set("data", dataArray);
      for (int j = startIdx; j < dates.size(); j++) {
        ObjectNode dataPoint = jackson.createObjectNode();
        dataPoint.put("x", timestamps[j]);
        dataPoint.put("y", columns[i].get(j));
        dataArray.add(dataPoint);
      }
      dataset.put("backgroundColor", COLORS[i]);
      dataset.put("borderColor", COLORS[i]);
      dataset.put("pointRadius", 0);
      dataset.put("fill", false);
      dataset.put("lineTension", 0);
      dataset.put("borderWidth", 1);
      datasets.add(dataset);
    }

    data.set("datasets", datasets);
    return data;
  }
```

howard-3:
So we use `JsonWriter` class to convert table into Json format. The code is simple 1-liner:
```
new JsonWriter().write(table, new Destination(stringWriter));
```
However JsonWriter contains a statically created object mapper with default date formats. Date formats tend to be different around the world. 
US prefers `MM/dd/yyyy`
EU prefers `dd/MM/yyyy`
some Asian countries prefer `yyyy/MM/dd`

To facilitate this we created our own custom `JsonWriter`, the code is similar what you have, however the only real change is the date format.

If we allow the use of a custom object mapper in the `JsonWriteOptions`, we don't need to implement our own writer class and reuse the `JsonWriter` class like
```
var asian = new JsonWriteOptions.Builder(new Destination(baos)).mapper(YYYY_MM_DD).build();
var eu = new JsonWriteOptions.Builder(new Destination(baos)).mapper(DD_MM_YYYY).build();
var us = ...
new JsonWriter().write(table, eu).write(...);
```

LordKa0S:
@benmccann is this abandoned? I'm facing the same issues for my use-case, and this would be a useful inversion.

PR #1249, opened 2024-02-24
Minor Documentation upgrades

jbsooter:
Some changes to the docs based on our students' experience with the user guide. 

- Add Export to Import/Export
- Remove SummarizeIf (not present to my knowledge in current version)

Thanks for contributing.

## Description

What was changed

## Testing

Did you add a unit test?

### DIFF
--- a/docs/userguide/importing_data.md
+++ b/docs/userguide/importing_data.md
@@ -237,6 +237,11 @@ Tablesaw supports importing data from HTML, JSON, and Excel. See the Javadoc for
 
 ## Exporting data
 
+### Exporting to CSV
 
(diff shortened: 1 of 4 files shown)


benmccann (changes requested):
Thanks. There are some good ideas here. I haven't touched this project or the docs in awhile, but I see there's a `docs-src` directory and I assume based on its name that the changes would need to be made there instead.

benmccann:
I found the docs. It looks like it is intended to have two copies of the docs, but you do indeed need to make changes to `docs-src` rather than `docs`. See https://github.com/jtablesaw/tablesaw/blob/master/docs-src/README.md

PR #1265, opened 2025-01-02
Fix Table.isDuplicate - check all matching rows before returning

### LINKED ISSUE #1248: Duplicate Rows May Remain After dropDuplicateRows Due to Early Return in isDuplicate
**Description:**
When using `dropDuplicateRows` to eliminate duplicate entries from a table, I observed that duplicates were still present in the output. Upon investigation, the root cause was identified in the `isDuplicate` function. This function is designed to iterate over rows that share a hash with the row being evaluated to determine if it is a duplicate. However, it incorrectly returns `false` (indicating the row is unique) during the first iteration if the first checked row does not match, without examining the remaining rows.

**Expected Behavior:**
The `isDuplicate` function should only return `false` after all rows with the matching hash have been checked and none are found to be identical to the row being evaluated. This ensures that a row is only considered unique if it has been verified against all potential duplicates.

**Actual Behavior:**
The function returns `false` prematurely after comparing with the first row that shares a hash, potentially leaving unexamined duplicates in the table.

**Resolution:**
The issue was resolved by modifying `isDuplicate` to complete its iteration over all rows with a matching hash before deciding that the row is not a duplicate. This change ensured that `dropDuplicateRows` correctly removed all duplicates from the table.

ccleva:
Thanks for contributing.

## Description

Table.isDuplicate now checks all matching rows before returning. Fixes #1248.

## Testing

Yes


core/src/test/java/tech/tablesaw/api/FixDropDuplicatesTest.java
```diff
@@ -0,0 +1,38 @@
...
+import static org.junit.jupiter.api.Assertions.*;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import it.unimi.dsi.fastutil.ints.Int2ObjectMap;
+import it.unimi.dsi.fastutil.ints.Int2ObjectOpenHashMap;
+import it.unimi.dsi.fastutil.ints.IntArrayList;
+import tech.tablesaw.io.csv.CsvReadOptions;
+
+class FixDropDuplicatesTest {
```
benmccann, line 15: we've already got three tests for duplicates in TableTest.java. I think it'd be preferable to add any new test there alongside the existing tests

core/src/test/java/tech/tablesaw/api/FixDropDuplicatesTest.java
```diff
@@ -0,0 +1,38 @@
...
+    
+    @BeforeAll
+    static void setUpBeforeClass() throws Exception {
+        testTable = Table.read().usingOptions(CsvReadOptions
+            .builder(new File(SOURCE_FILE_NAME))
+            .missingValueIndicator("-"));    }
+
+    @Test
+    void test() throws Exception {
+        Method privateMethod = Table.class.getDeclaredMethod("isDuplicate", Row.class, Int2ObjectMap.class);
```
benmccann, line 28: I'd prefer to just make it package visible for testing

Or perhaps better would be to test `dropDuplicateRows` like the other duplicate tests are doing if possible. I haven't closely checked the code itself to understand the issue yet. Is it really hard to reproduce the issue from the public APIs? (a  bit of description of the issue in the PR might help as well since it has been awhile since I've looked at this codebase)

ccleva: The issue only happens when non duplicate row hashes collide, it's pretty hard to test even with mocking.

As you prefer for method visibility, I personally prefer to keep things private.

ccleva: The issue description is detailed in #1248, do you prefer I copy it there ?

benmccann: Ah, sorry, I hadn't taken a look at the issue. No need to copy that into the PR

benmccann: There are a couple methods we've annotated with `@VisibleForTesting`. You could do that here when changing the visibility

benmccann:
Thanks for this. I think this whole method actually needs to be rewritten. See https://github.com/jtablesaw/tablesaw/pull/1267 which does that and fixes the issue in what I think is probably a simpler way

PR #1266, opened 2025-01-02
Fix, document, and test ColumnType.compare and Table.compareRows

### LINKED ISSUE #1229: Table.compareRows not working as expected
The static method `Table.compareRows` always returns `false` except for the last row. It happens even when we compare rows of the same table. 
```
Table table = Table.read().csv(new File("data.csv"));
System.out.println(table.shape());

for(int i = 0; i < table.rowCount(); i++) {
    System.out.println("ROW " + i + ": "+ Table.compareRows(i, table, table));
}
```
Actual output:
```
data.csv: 5 rows X 2 cols
ROW 0: false
ROW 1: false
ROW 2: false
ROW 3: false
ROW 4: true
```
Expected output:
```
data.csv: 5 rows X 2 cols
ROW 0: true
ROW 1: true
ROW 2: true
ROW 3: true
ROW 4: true
```

ccleva:
Thanks for contributing.

## Description

- ColumnType.compare now compares rows at the same rownumber instead of comparing with the last row of the other column
- Table.compareRows now returns false if the number of columns is different in both tables
- Added missing javadoc on both methods

Fixes #1229

## Testing

Yes, let me know if you want them moved to an existing test class


ccleva:
All conversation should be fixed. I rebased the branch also in case

benmccann:
Can you rebase this PR again to make sure the formatter runs against it now that the formatter's been fixed?

ccleva:
I'm unable to make the formatter working for the moment

ccleva:
> @ccleva could you try rebasing and formatting now that you've gotten the formatter working?

Yes, I rebased then ran the hook manually

core/src/main/java/tech/tablesaw/api/ColumnType.java
```diff
@@ -82,10 +82,13 @@ static ColumnType valueOf(String name) {
...
 
-  /** TODO: Research this method to provide a good comment */
-  default boolean compare(int rowNumber, Column<?> temp, Column<?> original) {
-    Object o1 = original.get(rowNumber);
-    Object o2 = temp.get(temp.size() - 1);
+  /**
+   * Compares the row at {@code rowNumber} in {@code column1} and {@code column2} and returns whether they are equal.
+   * @throws {@code IndexOutOfBoundsException} if {@code rowNumber} exceeds either column size
+   */
+  default boolean compare(int rowNumber, Column<?> column1, Column<?> column2) {
```
benmccann, line 89: this feels like kind of a weird method honestly. I wonder if we even need it to be a standalone method or if we can just inline it into `compareRows`

ccleva: It might be used by people for other purposes. No harm in keeping it I think.

benmccann: We're changing the functioning of it pretty drastically. Anyone who is relying on it is going to be broken. Better for them to find that out at compile-time because we removed it than at runtime

benmccann (approved):

PR #1269, opened 2025-01-03
Add test for #1197

ccleva:
Thanks for contributing.

## Description

- Add test for #1197
- Code refactored to call toLowerCase() only once

## Testing

Yes

### DIFF
--- a/core/src/main/java/tech/tablesaw/io/FileReader.java
+++ b/core/src/main/java/tech/tablesaw/io/FileReader.java
@@ -129,10 +129,10 @@ public abstract class FileReader {
    * @param headerNames The header names to be potentially adjusted.
    */
   private void renameDuplicateColumnHeaders(String[] headerNames) {
-    final Map<String, Integer> nameCounter = new HashMap<>();
+    Map<String, Integer> nameCounter = new HashMap<>();
     for (int i = 0; i < headerNames.length; i++) {
-      final String name = headerNames[i];
-      final String lowerCase = name.toLowerCase();
+      String name = headerNames[i];
+      String lowerCase = name.toLowerCase();
       Integer count = nameCounter.get(lowerCase);
       if (count == null) {
         nameCounter.put(lowerCase, 1);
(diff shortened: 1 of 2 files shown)


core/src/main/java/tech/tablesaw/io/FileReader.java
```diff
@@ -129,15 +129,16 @@ public String[] getColumnNames(
    * @param headerNames The header names to be potentially adjusted.
    */
   private void renameDuplicateColumnHeaders(String[] headerNames) {
-    Map<String, Integer> nameCounter = new HashMap<>();
+    final Map<String, Integer> nameCounter = new HashMap<>();
```
benmccann, line 132: Let's remove `final` everywhere. It doesn't match the style of the codebase

Maybe we can consider whether to migrate to `var` after a release has been out for awhile if folks don't complain about us dropping Java 8. Perhaps `final var` would look less noisy than `final` looks with types. The other option might be lombok's `val`, but not sure if I'd want to introduce something that non-standard

core/src/test/java/tech/tablesaw/io/csv/CsvReaderTest.java
```diff
@@ -161,6 +161,17 @@ void allowDuplicateColumnNames() throws IOException {
     assertEquals("Col1-2", dupes.columnNames().get(1));
   }
 
+  @Test
+  void allowDuplicateColumnNamesInsensitive() throws IOException {
+    final Reader reader2 =
+        new StringReader(
+            "Col1" + COMMA + "col1" + LINE_END + "\"first\"" + COMMA + "second" + LINE_END);
```
benmccann, line 168: We could probably make this a tad simpler 
```suggestion
            "Col1" + COMMA + "col1" + LINE_END + "first" + COMMA + "second" + LINE_END);
```

ccleva: Sure, I took that from the test above. Done

core/src/test/java/tech/tablesaw/io/csv/CsvReaderTest.java
```diff
@@ -161,6 +161,17 @@ void allowDuplicateColumnNames() throws IOException {
     assertEquals("Col1-2", dupes.columnNames().get(1));
   }
 
+  @Test
+  void allowDuplicateColumnNamesInsensitive() {
+    final Reader reader2 =
```
benmccann, line 166:
```suggestion
    Reader reader2 =
```

PR #1273, opened 2025-01-03
Fix tests

ccleva:
Thanks for contributing.

## Description

- Correct `DataFrameJoinerTest.leftOuterEmptyLeftTable` to use `JoinType.LEFT_OUTER` instead of the default INNER - inner join is tested in `innerJoinEmptyLeftTable`
- Added assertions in `AggregateFunctionsTest.testMultipleColumnTypes` and `DoubleArraysTest.testTo2dArray`

## Testing

Did you add a unit test?

### DIFF
--- a/core/src/test/java/tech/tablesaw/aggregate/AggregateFunctionsTest.java
+++ b/core/src/test/java/tech/tablesaw/aggregate/AggregateFunctionsTest.java
@@ -323,7 +323,12 @@ class AggregateFunctionsTest {
     StringColumn stringColumn = StringColumn.create("s", strings);
 
     Table table = Table.create("test", booleanColumn, numberColumn);
-    table.summarize(booleanColumn, numberColumn, countTrue, stdDev).by(stringColumn);
+    Table summary = table.summarize(booleanColumn, numberColumn, countTrue, stdDev).by(stringColumn);
+    
+    assertEquals(2.0, summary.doubleColumn(1).get(0), "Wrong summary");
+    assertEquals(0.0, summary.doubleColumn(1).get(1), "Wrong summary");
+    assertEquals(Math.sqrt(2), summary.doubleColumn(2).get(0), "Wrong summary");
+    assertEquals(Math.sqrt(2), summary.doubleColumn(2).get(1), "Wrong summary");    
   }
 
   @Test
--- a/core/src/test/java/tech/tablesaw/aggregate/AggregateFunctionsTest.java
+++ b/core/src/test/java/tech/tablesaw/aggregate/AggregateFunctionsTest.java
@@ -325,10 +325,10 @@ class AggregateFunctionsTest {
     Table table = Table.create("test", booleanColumn, numberColumn);
     Table summary = table.summarize(booleanColumn, numberColumn, countTrue, stdDev).by(stringColumn);
     
-    assertEquals(2.0, summary.doubleColumn(1).get(0), "Wrong summary");
-    assertEquals(0.0, summary.doubleColumn(1).get(1), "Wrong summary");
-    assertEquals(Math.sqrt(2), summary.doubleColumn(2).get(0), "Wrong summary");
-    assertEquals(Math.sqrt(2), summary.doubleColumn(2).get(1), "Wrong summary");    
+    assertEquals(2.0, summary.doubleColumn(1).get(0), 1e-6, "Wrong summary");
+    assertEquals(0.0, summary.doubleColumn(1).get(1), 1e-6,  "Wrong summary");
+    assertEquals(Math.sqrt(2), summary.doubleColumn(2).get(0), 1e-6,  "Wrong summary");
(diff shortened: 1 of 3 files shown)


core/src/test/java/tech/tablesaw/aggregate/AggregateFunctionsTest.java
```diff
@@ -323,7 +323,12 @@ void testMultipleColumnTypes() {
     StringColumn stringColumn = StringColumn.create("s", strings);
 
     Table table = Table.create("test", booleanColumn, numberColumn);
-    table.summarize(booleanColumn, numberColumn, countTrue, stdDev).by(stringColumn);
+    Table summary = table.summarize(booleanColumn, numberColumn, countTrue, stdDev).by(stringColumn);
+    
+    assertEquals(2.0, summary.doubleColumn(1).get(0), 1e-6, "Wrong summary");
+    assertEquals(0.0, summary.doubleColumn(1).get(1), 1e-6,  "Wrong summary");
+    assertEquals(Math.sqrt(2), summary.doubleColumn(2).get(0), 1e-6,  "Wrong summary");
+    assertEquals(Math.sqrt(2), summary.doubleColumn(2).get(1), 1e-6,  "Wrong summary");
```
benmccann, line 331: "Wrong summary" doesn't add anything and my memory is a bit fuzzy, but I think it might also hide what the actual values were
```suggestion
    assertEquals(2.0, summary.doubleColumn(1).get(0), 1e-6);
    assertEquals(0.0, summary.doubleColumn(1).get(1), 1e-6);
    assertEquals(Math.sqrt(2), summary.doubleColumn(2).get(0), 1e-6);
    assertEquals(Math.sqrt(2), summary.doubleColumn(2).get(1), 1e-6);    
```

core/src/test/java/tech/tablesaw/util/DoubleArraysTest.java
```diff
@@ -27,8 +28,9 @@ public class DoubleArraysTest {
   public void testTo2dArray() throws Exception {
     Table table = Table.read().csv("../data/tornadoes_1950-2014.csv");
     TableSliceGroup tableSliceGroup = table.splitOn("Scale");
-    int columnNuumber = table.columnIndex("Injuries");
-    DoubleArrays.to2dArray(tableSliceGroup, columnNuumber);
+    int columnNumber = table.columnIndex("Injuries");
+    final double[][] a2dArray = DoubleArrays.to2dArray(tableSliceGroup, columnNumber);
```
benmccann, line 32:
```suggestion
    double[][] a2dArray = DoubleArrays.to2dArray(tableSliceGroup, columnNumber);
```

core/src/test/java/tech/tablesaw/util/DoubleArraysTest.java
```diff
@@ -27,8 +28,9 @@ public class DoubleArraysTest {
...
+    assertEquals(7, a2dArray.length, "Wrong Scale array length");
```
benmccann, line 33:
```suggestion
    assertEquals(7, a2dArray.length);
```

benmccann:
Can you rebase this PR to make sure the formatter runs against it now that the formatter's been fixed?

ccleva:
I rebased it without issue in local, but the new formatter gives me an error related to jvm version with both eclipse and command line. I need to see how to fix this before any other contribution.

benmccann:
Hmm. That's too bad. I have Java 11 installed for java and javac. What version do you have? I'd hate if it's causing issues for everyone who wants to contribute

ccleva:
> Hmm. That's too bad. I have Java 11 installed for java and javac. What version do you have? I'd hate if it's causing issues for everyone who wants to contribute

I'm using java 17 for command line, and latest eclipse requires (and provides) java 21. The issue happens for java 16+, see https://github.com/Cosium/git-code-format-maven-plugin?tab=readme-ov-file#jdk-16-peculiarities.

The recommended configuration solves the issue. This should be documented somewhere for contributors using java 16+.

ccleva:
Rebased, commit hooks ran, and force-pushed. Should be fine now.

ccleva:
> Agreed. Is this something we can update in the `pom.xml` or it should just go in the CONTRIBUTING.md? Either way, mind sending a PR for that? You're better able to test it since you're on a newer Java than I am

You have to add jvm options to maven, typically in the `.mvn/jvm.config` file in the repo. Would you prefer a commented-out option file in the repo and the documentation stating it needs to be uncommented for java 16+, or only the description of what needs to be done ? In any case `CONTRIBUTING.md` seems the right place.

benmccann:
I wonder if we can just add that file and leave it uncommented. Hopefully it won't have any ill effects for Java 11 users. I could test that if you share the file

ccleva:
Done in #1280. It did not make the builds with java 11 fail, but I would be more confident if you test it in local.

PR #1274, opened 2025-01-03
Fix empty csv test

ccleva:
Thanks for contributing.

## Description

- Fix `CSVReaderTest.testEmptyFileHeaderEnabled` by enabling header - disabled header tested in `testEmptyFileHeaderDisabled`
- Fix `FileReader.getColumnNames` to not throw NPE when header is enabled and file is empty

## Testing

Fixed one

### DIFF
--- a/core/src/main/java/tech/tablesaw/io/FileReader.java
+++ b/core/src/main/java/tech/tablesaw/io/FileReader.java
@@ -94,6 +94,12 @@ public abstract class FileReader {
     if (options.header()) {
 
       String[] headerNames = parser.parseNext();
+      
+      if (headerNames == null) {
+        // no header because file is empty
+        // return empty String[]
+        return new String[0];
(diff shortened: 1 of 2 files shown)


core/src/main/java/tech/tablesaw/io/FileReader.java
```diff
@@ -94,6 +94,12 @@ public String[] getColumnNames(
     if (options.header()) {
 
       String[] headerNames = parser.parseNext();
+      
+      if (headerNames == null) {
+        // no header because file is empty
+        // return empty String[]
+        return new String[0];
```
benmccann, line 101: the comment basically just duplicates the return line. if the return is a bit unclear it might be clearer written differently
```suggestion
        return new String[] {};
```

benmccann (approved):

PR #1309, opened 2026-04-07
Fix excel table area detection failures

### LINKED ISSUE #1307: XLSX can't read 0
When reading an XLSX file, if there is only one numeric column and a cell in the middle contains 0, the reading process stops, and the subsequent data is discarded.

ccleva:
Thanks for contributing.

## Description

Blank cells detection (used to identify the table area in excel sheets) did not take into account formulas and wrongly identified numerical 0 or false boolean as blank, leading to failed or incorrect table reading.

Fixes #1307 

## Testing

Added tests to confirm the issues. Tests are now passing with the fix.


ccleva:
Sonar complains about test methods being public, let me know if you prefer:
- to keep the new methods public for consistency with the other test methods
- to change the new test methods to package as recommended by Sonar
- to change all test methods to package for consistency

